PR #883, opened 2018-03-23
[BUILD] Bugfix Jenkins Pipeline

michal-raska:
This fixes the following:

- the modified files should be detected correctly
- adds separate checks per stage on GitHub

### DIFF
--- a/Jenkinsfile2
+++ b/Jenkinsfile2
@@ -1,3 +1,5 @@
+#!/usr/bin/groovy
+
 @Library('test-shared-library@1.2') _
 
 import ai.h2o.ci.Utils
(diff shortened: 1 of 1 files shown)


st-pasha:
This shows "12 successful checks" on GitHub -- that's good.
But on Jenkins it still display 3 steps only. Do you know why that's the case, and should we care about that or not?

michal-raska:
@st-pasha It's how the Jenkins UI works. We specify three "tasks" executed in parallel (x86_64_centos7, x86_64_linux, x86_64_macos). Each of these consists of multiple stages (build, coverage, test py35, test py36) however these stages are not displayed in BlueOcean UI, but if you check then "traditional" UI, you will see something like:
image

If you want, we can modify the pipeline, so that the BlueOcean UI will display the exact same number of stages, but AFAIK this will require all stages to run in parallel and that's something I would not recommend (the overhead for allocating node, pulling required files etc. is too big compared to the duration of stages).

st-pasha:
Seems like this is exactly what https://stackoverflow.com/a/36893393/958624 suggests: a limitation of Blue Ocean.

PR #908, opened 2018-04-02
Implement __str__ in CastExpr

### LINKED ISSUE #907: Class CastExpr should implement method __str__
Happens when applying a cast expression, for example:
```
import datatable as dt
f0 = dt.Frame(range(10))
f1 = f0[:, [dt.int64(dt.f[0])]]
```

st-pasha:
Closes #907


st-pasha:
It should speed up calculation of uniques for `int64`s and `float64`s.

st-pasha:
It's just a syntactic sugar...

PR #911, opened 2018-04-03
uncomment ppc build

CreatureDev:


abal5:
Just switch the job to use Jenkinsfile instead of Jenkinsfile2 but the build numbers will be reset.

PR #935, opened 2018-04-13
Travis ci cache

abal5:
Adding ccache and pip to cache


abal5:
Added ccache and pip to caching, was not able to add python 3.7 says not available.

mmalohlava:
Thanks a lot! It seems clang gets slower with CCACHE - http://peter.eisentraut.org/blog/2014/12/01/ccache-and-clang-part-3/

PR #1021, opened 2018-05-08
[BUILD] GitHub Release

### LINKED ISSUE #1016: Make promotion stage optional
The promotion stage which copies artifacts from artifacts.h2o.ai to h2o-release needs to be optional and ask user for confirmation.

Furthermore, the action needs to:
- [x] tag a commit which gets promoted by a version tag
- [x] create a GitHub release
  - [x] `title` is the release number
  - [x] `description` needs to be scraped from CHANGELOG.md
  - [x] `description` needs to contain links to files in public h2o-release bucket
  - [x] Attach artifacts to github release page.

michal-raska:
Closes #1016.

Flow of the pipeline is now following (nothing is changed for PRs):

1. Initialization
2. Build, Coverage and Test all platforms
3. If we are building `master` or `rel-*` branch then publish to `artifacts.h2o.ai`
    1. ~If building `master`, publish to `h2o-release` under `nightly`~. Nightly builds are uploaded to `artifacts.h2o.ai` only.
    2. If building other branch, ask for user approval, if granted, then publish to `h2o-release` under `stable` and create a GitHub release.

### DIFF
--- a/ci/default.groovy
+++ b/ci/default.groovy
@@ -6,6 +6,8 @@
 //  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 //------------------------------------------------------------------------------
 
+GENERIC_VERSION_REGEX = /^### \[v(\d+\.)+\d\].*/
+
 def buildAll(stageDir, platform, ciVersionSuffix, py36VenvCmd, py35VenvCmd, extraEnv) {
     dumpInfo()
 
@@ -135,6 +137,45 @@ def pullFilesFromArch(final String filter, final String targetDir) {
           target              : "${targetDir}/",
           flatten             : true
     ])
+}
+
+def getChangelogPartForVersion(final version, final changelogPath = 'CHANGELOG.md') {
+    if (version == null || version.trim().isEmpty()) {
+        error 'Version must be set'
+    }
+
+    echo "Reading changelog from ${changelogPath}"
+    def changelogLines = readFile(changelogPath).readLines()
+
+    def startIndex = changelogLines.findIndexOf {
+        it ==~ /^### \[v${version}\].*/
+    }
+    if (startIndex == -1) {
+        error 'Cannot find Changelog for this version'
+    }
+
+    def endIndex = startIndex
+    if ((startIndex + 1) < changelogLines.size() - 1) {
+        endIndex = startIndex + changelogLines[(startIndex + 1)..-1].findIndexOf {
+            it ==~ GENERIC_VERSION_REGEX
+        }
+    }
+
+    return changelogLines[startIndex..endIndex].join('\n').trim()
+}
+
+def getReleaseDownloadLinksText(final folder, final s3PathPrefix) {
+    def files = sh(script: "cd ${folder} && find . \\( -name '*.whl' -o -name '*.tar.gz' \\) -printf '%P\n'", returnStdout: true).trim().readLines()
+    def resultLines = ['## Download links ##', '']
+    files.each { file ->
+        resultLines += "- [${file}](${s3PathPrefix}/${file})"
+    }
+    return resultLines.join('\n').trim()
+
+}
+
+def getReleaseBucket(final targetFolder, final versionText) {
+    return "s3://h2o-release/datatable/${targetFolder}/datatable-${versionText}/"
 }
 
 return this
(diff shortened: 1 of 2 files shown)


st-pasha:
Could you please link in the issue(s) this implements?

michal-raska:
@st-pasha updated the description

Jenkinsfile2
```diff
@@ -106,6 +106,8 @@ def needsLargerTest
 SOURCE_DIR = "/home/0xdiag"
 TARGET_DIR = "/tmp/pydatatable_large_data"
 
+S3_FOLDER_STABLE = 'stable'
```
st-pasha: Can this be replaced with `"s3://h2o-release/datatable/stable"`, so that looking at the top of the file one could see immediately where the releases are being published?

ci/default.groovy
```diff
@@ -137,5 +139,44 @@ def pullFilesFromArch(final String filter, final String targetDir) {
...
+    def files = sh(script: "cd ${folder} && find . \\( -name '*.whl' -o -name '*.tar.gz' \\) -printf '%P\n'", returnStdout: true).trim().readLines()
+    def resultLines = ['## Download links ##', '']
+    files.each { file ->
+        resultLines += "- [${file}](${s3PathPrefix}/${file})"
+    }
+    return resultLines.join('\n').trim()
+
+}
+
+def getReleaseBucket(final targetFolder, final versionText) {
```
st-pasha: This method appears to be unused.

Jenkinsfile2
```diff
@@ -236,63 +248,96 @@ ansiColor('xterm') {
...
-                                sh """
-                                   ls
-                                   s3cmd put -P ${publishS3Dir}/dist/*.whl s3://h2o-release/datatable/${targetFolder}/datatable-${versionText}/
-                                   s3cmd put -P ${publishS3Dir}/dist/*.tar.gz s3://h2o-release/datatable/${targetFolder}/datatable-${versionText}/
-                               """
+                    stage('GitHub Release') {
+                        dumpInfo()
+                        docker.withRegistry("https://docker.h2o.ai", "docker.h2o.ai") {
+                            docker.image('docker.h2o.ai/opsh2oai/hub').inside("--init -v /home/jenkins/.ssh:/home/jenkins/.ssh:ro -v /home/jenkins/.gitconfig:/home/jenkins/.gitconfig:ro") {
+                                withCredentials([string(credentialsId: 'h2o-ops-personal-auth-token', variable: 'GITHUB_TOKEN')]) {
```
st-pasha: probably `H2O_OPS_TOKEN_CREDS_ID` (and even that variable name seems too specific, how about renaming it into just `CREDS_ID`?)

Jenkinsfile2
```diff
@@ -236,63 +248,96 @@ ansiColor('xterm') {
...
-                                def versionText = utilsLib.getCommandOutput("cat build/dist/VERSION.txt")
-                                sh """
-                                   ls
-                                   s3cmd put -P ${publishS3Dir}/dist/*.whl s3://h2o-release/datatable/${targetFolder}/datatable-${versionText}/
-                                   s3cmd put -P ${publishS3Dir}/dist/*.tar.gz s3://h2o-release/datatable/${targetFolder}/datatable-${versionText}/
-                               """
+                    stage('GitHub Release') {
+                        dumpInfo()
+                        docker.withRegistry("https://docker.h2o.ai", "docker.h2o.ai") {
+                            docker.image('docker.h2o.ai/opsh2oai/hub').inside("--init -v /home/jenkins/.ssh:/home/jenkins/.ssh:ro -v /home/jenkins/.gitconfig:/home/jenkins/.gitconfig:ro") {
```
st-pasha: Is it ok that `jenkins` user is hard-coded here?

ci/default.groovy
michal-raska: removed

mmalohlava:
@michal-raska may you please update the branch and resolve the conflict. And let's merge it after that.

Jenkinsfile2
mmalohlava: Good catch - i would not depend on host settings. 

@michal-raska can we change it to use Jenkins credentials?

michal-raska:
@mmalohlava rebase on top of master. .gitconfig and ssh key is now stored as a credential in Jenkins. Also the releases should create "regular" releases instead of drafts.

PR #1027, opened 2018-05-09
Added a readme file with the description of the release process

st-pasha:
This document should give a *detailed* description of the release process. It should be updated whenever the process changes. Ideally it should be sufficiently complete that it can be given to a new hire, and they would be able to make a new release without anyone's help.

The document is still incomplete. Please review and post any updates directly into this branch.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -41,9 +41,17 @@ update applies.
   additional commits into this branch to fix any emergent problems.
 - Merge the PR into master, but **DO NOT delete the branch**. This branch
   should be kept forever.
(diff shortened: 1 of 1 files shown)


mmalohlava:
One more thing is missing - clarify where we upload artifacts.

st-pasha:
Thanks for the update, @michal-raska.
I wonder though what would happen if I press "Promote" button and it fails (for example S3 is down). Would I be able to try and promote again?

michal-raska:
@st-pasha No, but if you want, it shouldn't be hard to add such functionality. However, I would not expect the S3 to fail when we take size of the wheels into consideration.

PR #1062, opened 2018-05-27
Moments

jlavileze:
Adding capabilities to compute the third and fourth central moments: skewness and kurtosis. Both are computed using Welford's method, sourced from John Cook's guide.


st-pasha:
Still need to add tests to make sure that it works as intended...

st-pasha:
I've added #1084 as a separate task, so that this can be merged.

st-pasha:
I'll merge this, and then we can fix all the kinks later.

Issue #1100, opened 2018-06-06
dockerfile to build datatable for Ubuntu 18.04


jangorecki:
Building datatable on latest LTS Ubuntu does not seems to be as straightforward as on 16.04.  It would help a lot if we would add dockerfile build on that version.

st-pasha:
With the recent changes to the build script, datatable still cannot build on Ubuntu 18.04 automatically. However, it now goes further than before:

<details><summary>`$ make` output</summary>

```
(py36) pasha@mr-dl1:~/datatable$ make
make clean
make[1]: Entering directory '/home/pasha/datatable'
rm -rf .asan
rm -rf .cache
rm -rf .eggs
rm -rf build
rm -rf dist
rm -rf datatable.egg-info
rm -f *.so
rm -f datatable/lib/_datatable*.so
rm -f ci/fast.mk
find . -type d -name "__pycache__" -exec rm -rf {} +
rm -f Dockerfile-centos7.x86_64_centos7
make[1]: Leaving directory '/home/pasha/datatable'
make build
make[1]: Entering directory '/home/pasha/datatable'
python setup.py build

Find an LLVM installation
  Environment variable LLVM is not set
  Environment variable LLVM7 is not set
  Environment variable LLVM6 is not set
  Environment variable LLVM5 is not set
  Environment variable LLVM4 is not set
  Candidate directory `/usr/local/opt/llvm` not found
  The build will proceed without Llvm support

Determine the compiler
  Environment variable `CXX` is not set
  Environment variable `CC` is not set
  Llvm installation not found, cannot search for the clang++ compiler
  distutils.ccompiler reports the default compiler to be `cc`
  Compiler `clang++` is not found
  Compiler `gcc` will be used

Prepare the environment
  CC = gcc
  CXX = gcc
  LDFLAGS = 
  ARCHFLAGS = -m64
  LLVM_CONFIG = 
  MACOSX_DEPLOYMENT_TARGET = 

Generate git version file
  Environment variable DTBL_GIT_HASH not present
  `git rev-parse HEAD` = 0fa80696ce4fbb79cd56795e36de1d8639214c3d
  Generating file /home/pasha/datatable/datatable/__git__.py

Find the required dynamic libraries
  Library `libomp.so` found at /home/arno/datatable/datatable/lib/libomp.so
  Library `libc++.so.1` found at /home/arno/datatable/datatable/lib/libc++.so.1
[...]

PR #1112, opened 2018-06-13
[BUILD] Rewrite Jenkinsfile. Remove Ubuntu build, keep Ubuntu tests for now.

michal-raska:


Makefile
```diff
@@ -220,6 +220,163 @@ ubuntu_docker_build: Dockerfile-ubuntu.$(PLATFORM).tag
 ubuntu_docker_publish: Dockerfile-ubuntu.$(PLATFORM).tag
 	docker push $(CONTAINER_NAME_TAG)
 
+ARCH_NAME ?= $(shell uname -m)
+DOCKER_IMAGE_TAG ?= 0.3.2-PR-996.3
+CENTOS_DOCKER_IMAGE_NAME ?= docker.h2o.ai/opsh2oai/datatable-build-$(ARCH_NAME)_centos7:$(DOCKER_IMAGE_TAG)
+UBUNTU_DOCKER_IMAGE_NAME ?= docker.h2o.ai/opsh2oai/datatable-build-$(ARCH_NAME)_ubuntu:$(DOCKER_IMAGE_TAG)
```
mmalohlava: Do we still need Ubuntu here?

Makefile
```diff
@@ -220,6 +220,163 @@ ubuntu_docker_build: Dockerfile-ubuntu.$(PLATFORM).tag
...
+		--init \
+		-u `id -u`:`id -g` \
+		-v `pwd`:/dot \
+		-w /dot \
+		--entrypoint /bin/bash \
+		-e "CI_VERSION_SUFFIX=$(CI_VERSION_SUFFIX)" \
+		$(CUSTOM_ARGS) \
+		$(CENTOS_DOCKER_IMAGE_NAME) \
+		-c ". activate $(BUILD_VENV) && \
+			export LD_LIBRARY_PATH=$$LD_LIBRARY_PATH:/opt/h2oai/dai/python/lib/ && \
```
mmalohlava: the LD_LIBRARY_PATH should be setup in Dockerfile

Makefile
```diff
@@ -220,6 +220,163 @@ ubuntu_docker_build: Dockerfile-ubuntu.$(PLATFORM).tag
...
+		--init \
+		-u `id -u`:`id -g` \
+		-v `pwd`:/dot \
+		-w /dot \
+		--entrypoint /bin/bash \
+		-e "DT_LARGE_TESTS_ROOT=$(DT_LARGE_TESTS_ROOT)" \
+		$(CUSTOM_ARGS) \
+		$(CENTOS_DOCKER_IMAGE_NAME) \
+		-c ". activate $(TEST_VENV) && \
+			export LD_LIBRARY_PATH=$$LD_LIBRARY_PATH:/opt/h2oai/dai/python/lib/ && \
```
mmalohlava: Setting `LD_LIBRARY_PATH` is little bit suspicious. Why do you need it?

Makefile
```diff
@@ -220,6 +220,163 @@ ubuntu_docker_build: Dockerfile-ubuntu.$(PLATFORM).tag
...
+		-v `pwd`:/dot \
+		-w /dot \
+		--entrypoint /bin/bash \
+		-e "DT_LARGE_TESTS_ROOT=$(DT_LARGE_TESTS_ROOT)" \
+		$(CUSTOM_ARGS) \
+		$(UBUNTU_DOCKER_IMAGE_NAME) \
+		-c ". /envs/$(TEST_VENV)/bin/activate && \
+			python --version && \
+			pip freeze && \
+			pip install --no-cache-dir --upgrade dist/*.whl && \
```
mmalohlava: Do we need to call --upgrade? In the container, they should not be any datatable.whl installed.
Furthermore, it is better to reference directly by name instead of wildcard (if it is possible).

ci/Jenkinsfile2
```diff
@@ -1,354 +1,636 @@
...
-        buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '180', numToKeepStr: ''))
-])
-
+///////////////
+// CONSTANTS //
+///////////////
+NODE_LABEL = 'docker && !mr-0xc8'
+OSX_NODE_LABEL = 'osx'
+PPC_NODE_LABEL = 'ibm-power'
+RELEASE_NODE_LABEL = 'master'
```
mmalohlava: Why we cannot keep in build matrix?

tomkraljevic:
@jangorecki there is no reason to build on ubuntu.  The centos7 build will run on ubuntu.  we just need to re-package the same bits in a .dpkg.

st-pasha:
@michal-raska @tomkraljevic 
Does this PR makes #1100 obsolete?

Makefile
michal-raska: We need ubuntu for coverage. Also I think we might test on Ubuntu, since it's widely used and the tests are quite fast.

ci/Jenkinsfile2
michal-raska: We should implement the build matrix in next iteration, WDYT?

michal-raska:
@st-pasha no, I think #1100 is still relevant. Maybe we can stop testing on 16.04 and start testing on 18.04, WDYT? @tomkraljevic @mmalohlava

Issue #952, opened 2018-04-18
Add download links for datatable


st-pasha:
- [x] Publish `datatable` wheels into proper bucket on S3: `s3://h2o-release/datatable/`. Currently we place datatable builds into `s3://artifacts.h2o.ai/releases/ai/h2o/pydatatable/` which is very hard to find, and gives wrong impression that datatable is not a standalone project.
- [x] Decide whether we want to publish after each commit to master, or on a nightly basis.
- [ ] Create redirect files to the "latest nightly" and "latest stable" wheels. There should be separate files for different flavors of wheels: linux/macosx/ppc/centos/linux-debug/macos-debug. Redirect should be "hard", so that following the link would immediately initiate download of the relevant wheel.
- [ ] Add links to those redirect files in the README.md front page.

mmalohlava:
@abal5 can you please help with creating the latest links? 

The best would be:
  1. have a promotion stage, which will copy datatable release artifacts to PUBLIC `s3://h2o-release/datatable bucket`
  2. generate a latest link (look into DAI Docker build - there is a helper script which helps you to generate http forward): The link generator is here: https://github.com/h2oai/h2oai/blob/dev/scripts/generate-stable-links.sh

mmalohlava:
@abal5 may you pls help with creation of the links for latest stable and latest nightly?

st-pasha:
This is also  needed for proper generation of documentation on readthedocs.io

PR #1155, opened 2018-07-10
Generate latest stable links for releases and nightly

abal5:
#952 [WIP] Recreating a clean PR, the script works but changes are needed on the s3 bucket end which is being tested right now.


jangorecki:
I am interested how to install nightly releases, but this branch is too big to find that, could you point me to some docs on that? currently I rebuilt from source to latest master each time (which might eventually not pass all tests if there will be commit directly to master).

Issue #1066, opened 2018-05-30
Reimplement Frame in pure C


st-pasha:
In order to achieve better micro performance, we should make `Frame` a pure-python class. This is especially important for handling the `__getitem__` / `__setitem__` / `__call__` methods.

Methods that remain to convert into C++:
- [x] `__repr__()` ... `__repr_pretty__()`
- [x] `view()` (widget viewer)
- [x] `rbind()`
- [x] `save()` (rename into `to_jay()`)
- [x] `to_csv()`
- [x] `sort()`
- [x] `min()`, `max()`,  ..., `nmodal()`
- [x] `min1()`, `max1()`, ..., `nmodal1()`
- [x] `to_pandas()`
- [x] `to_numpy()`
- [x] `materialize()`
- [x] `__sizeof__()`
- [x] `internal.check()`
- [x] `internal.isview()`

PR #1156, opened 2018-07-10
Aggregator

### LINKED ISSUE #1077: Aggregator in datatable
- Is there something datatable can't do just yet, but you think it'd be nice if it did?
Aggregate

- Is it related to some problem you're trying to solve?
Solve slow reading of NFF format files.

- What do you think the API for your feature should be?
 See API in the Java code. Methods required are in base class DataSource

See Java code in https://github.com/h2oai/vis-data-server/blob/master/library/src/main/java/com/h2o/data/Aggregator.java

Plus other classes in that package for support. All of this should be done in C++

oleksiyskononenko:
Initial implementation of `datatable.extras.aggregate` for 1D, 2D and ND table aggregations.

Includes:
- 1D continuous binning;
- 2D continuous binning;
- 1D categorical aggregation;
- 2D categorical aggregation;
- 2D mixed (continuous/categorical) aggregation;
- ND aggregation that also includes a projection method when ncols > max_dimensions;
- implementations of the `first()` and `count()` `datatable` reducers;
- other minor changes to `datatable`.

Closes #1077


c/extras/aggregator.cc
```diff
@@ -0,0 +1,337 @@
+// License, v. 2.0. If a copy of the MPL was not distributed with this
```
st-pasha: license header truncated

c/extras/aggregator.h
```diff
@@ -0,0 +1,41 @@
+// License, v. 2.0. If a copy of the MPL was not distributed with this
```
st-pasha: license header truncated

c/extras/aggregator.cc
```diff
@@ -0,0 +1,337 @@
...
+#include "rowindex.h"
+#include "types.h"
+#include "datatable_check.h"
+
+Aggregator::Aggregator(DataTable* dt)
+  : dt_in(dt)
+{
+  create_dt_out();
+}
+
```
st-pasha: where's `~Aggregator()` ?

c/expr/reduceop.cc
```diff
@@ -212,6 +232,22 @@ static gmapperfn resolve0(int opcode, SType stype) {
...
+    switch (stype) {
+      case ST_BOOLEAN_I1:
+      case ST_INTEGER_I1:  		return count_skipna<int8_t, int64_t>;
+      case ST_INTEGER_I2:  		return count_skipna<int16_t, int64_t>;
+      case ST_INTEGER_I4:  		return count_skipna<int32_t, int64_t>;
+      case ST_INTEGER_I8: 		return count_skipna<int64_t, int64_t>;
+      case ST_REAL_F4:     		return count_skipna<float, int64_t>;
+      case ST_REAL_F8:     		return count_skipna<double, int64_t>;
+      case ST_STRING_I4_VCHAR:  return count_skipna<int32_t, int64_t>;
+      case ST_STRING_I8_VCHAR:  return count_skipna<int64_t, int64_t>;
```
st-pasha: these should be `<uint32_t, ...>` and `<uint64_t, ...>`.

c/extras/aggregator.cc
```diff
@@ -0,0 +1,337 @@
...
+  for (int32_t i = 0; i < dt_out->nrows; ++i) {
+    d_c1[i] = (int32_t) (norm_factor * (d_c0[i] - c0->min()));
+  }
+}
+
+
+void Aggregator::aggregate_2d_continuous(double epsilon, int32_t nx_bins, int32_t ny_bins) {
+  RealColumn<double>* c0 = (RealColumn<double>*) dt_out->columns[0];
+  RealColumn<double>* c1 = (RealColumn<double>*) dt_out->columns[1];
+  double* d_c0 = static_cast<double*>(dt_out->columns[0]->data_w());
```
st-pasha: You could use `c0->elements_w()` here and below

c/py_datatable.cc
```diff
@@ -261,6 +262,23 @@ PyObject* delete_columns(obj* self, PyObject* args) {
 
 
 
+PyObject* aggregate(obj* self, PyObject* args) {
```
st-pasha: This method should not be in (py)DataTable class: the class will become user-facing once #1066 is implemented, so we want to keep the API clean.
The easiest approach is to declare this function global (e.g. listed in `DatatableModuleMethods` in  datatablemodule.c) and move its body to extras/aggregate.cc

datatable/extras/aggregate.py
```diff
@@ -0,0 +1,14 @@
+    #!/usr/bin/env python3
```
st-pasha: no spaces before the `#!`

datatable/extras/aggregate.py
```diff
@@ -0,0 +1,14 @@
...
+#   file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#-------------------------------------------------------------------------------
+from datatable import Frame
+    
+#@typed(n_bins=int,nx_bins=int,ny_bins=int)
+def aggregate(self, epsilon=1.0e-15, n_bins=500, nx_bins=50, ny_bins=50, max_dimensions=50, seed=0):
+    dt_agg = self._dt.aggregate(epsilon, n_bins, nx_bins, ny_bins, max_dimensions, seed)
+    return Frame(dt_agg)
+
+Frame.aggregate = aggregate
```
st-pasha: I wouldn't recommend adding this method to `Frame` class. If you just remove this line the code will work just fine:
```
from datatable.extras import aggregate
aggregate(frame, ...)
```

tests/extras/test_aggregate.py
```diff
@@ -0,0 +1,166 @@
...
+                             [2, 2, 1, 0, 0, 2, 0, 0, 1, 0]]
+    
+#-------------------------------------------------------------------------------
+# Aggregate 2D 
+#-------------------------------------------------------------------------------    
+    
+def test_aggregate_2d_continuous_integer_sorted():
+    d0 = dt.Frame([[0, 1, 2, 3, 4, 5, 6, 7, 8, 9], 
+                   [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]])
+    d1 = d0.aggregate(1e-15, 0, 3, 3)
```
st-pasha: perhaps it's better to use named parameters here... How do you know what those `0, 3, 3` mean?

c/extras/aggregator.cc
oleksiyskononenko: Yes, something happened to my editor when pasted. Fixed.

c/extras/aggregator.h
oleksiyskononenko: Yes, something happened to my editor when pasted. Fixed.

c/extras/aggregator.cc
oleksiyskononenko: This `dt_out` is the dataframe created and returned by the aggregator to the user. I'm not sure we should destroy it when the aggregator is destroyed. Do you think we should?

c/expr/reduceop.cc
oleksiyskononenko: Yes, you're right. Will fix it. Thanks.

c/py_datatable.cc
oleksiyskononenko: Ok, got it.

datatable/extras/aggregate.py
oleksiyskononenko: Yes, another bug of my editor. Fixed now. Thanks.

datatable/extras/aggregate.py
oleksiyskononenko: Been using it as `dt.aggregate(...)`. Sure, will remove it from the `Frame`.

tests/extras/test_aggregate.py
oleksiyskononenko: The definition of aggregate is as follows:

`def aggregate(self, epsilon=1.0e-15, n_bins=500, nx_bins=50, ny_bins=50, max_dimensions=50, seed=0)`

So these are just the number of bins for the test.

c/extras/aggregator.cc
oleksiyskononenko: Not sure where to comment on the review in general. So will do it here. Thanks a lot for your time, 

1) Yes, this is something I've also noticed when looked at the code on `github`. It looks ok in the editor, though. I will find the way to fix it.

2) I've already merged the `master` branch into the `aggregator` and found the problems you mention. Fixed the code so that it can be built.

3) Yes, getting rid out of the warnings now.

4) Sure, that's an easy thing to do.

Thanks!

c/expr/reduceop.cc
oleksiyskononenko: Fixed.

datatable/extras/aggregate.py
oleksiyskononenko: Fixed

st-pasha:
Regarding comments -- the easiest way is to leave them from the form at the bottom of the page in "Conversation" tab of the PR. Comments left as answers to in-line comments will eventually be collapsed out of view when the line they were assigned to changes.

Regarding my comment about `~Aggregator()`. I merely noticed that you declare it in the `aggregator.h` file, but then do not define in the `aggregator.cc` file. Normally this would compile but then fail at linking stage. However, your code doesn't fail at linking, which got me very surprised. I suspect the reason is that there is no place in the code where the destructor is invoked. Indeed, if you look at `pydatatable::aggregate()` function you'll see that you're creating Aggregate objects via `new Aggregate()`, but then never delete them. Thus, it's a resource leak.

As for `dt_in` and `dt_out` members of `Aggregate` -- there should be a clear understanding on who owns these pointers, i.e. who is required to free those resources once they are no longer needed. It looks like `Aggregate` is not owning them right now -- if that's intentional, then I suggest to leave a clarifying comment about the correct usage of the `Aggregate` class: in particular, what measures must the user take to ensure that the pointers do not become dangling.

c/py_datatable.cc
oleksiyskononenko: Fixed.

oleksiyskononenko: Fixed.

datatable/extras/aggregate.py
oleksiyskononenko: Fixed.

datatable/extras/__init__.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python3
+# © H2O.ai 2018; -*- encoding: utf-8 -*-
+#   This Source Code Form is subject to the terms of the Mozilla Public
+#   License, v. 2.0. If a copy of the MPL was not distributed with this
+#   file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#-------------------------------------------------------------------------------
+
+__all__ = ("aggregate", )
```
st-pasha: Generally, in order to export a symbol, you need to import or define it first. 
However in your case (allow Python to find `datatable.extras.aggregate`) having an empty `__init__.py` should work just fine.

oleksiyskononenko: Yes, I was about to remove this line. Thanks!

c/expr/reduceop.cc
```diff
@@ -77,6 +78,24 @@ static void sum_skipna(const int32_t* groups, int32_t grp, void** params) {
 
 
 
+
+//------------------------------------------------------------------------------
+// Count calculation
+//------------------------------------------------------------------------------
+
+template<typename IT, typename OT>
+static void count_skipna(const int32_t* groups, int32_t grp, void** params) {
```
st-pasha: Traditionally, `count(x)` function returns the number of non-NA values in `x`.
In your implementation, however, the function doesn't count anything, but merely returns the number of elements in each group. It's a valid function, just not a suitable name...

c/extras/aggregator.cc
```diff
@@ -0,0 +1,351 @@
...
+  }
+}
+
+
+void Aggregator::aggregate_1d_continuous(int32_t n_bins) {
+  RealColumn<double>* c0 = static_cast<RealColumn<double>*>(dt_out->columns[0]);
+  const double* d_c0 = static_cast<const double*>(dt_out->columns[0]->data());
+  int32_t* d_c1 = static_cast<int32_t*>(dt_out->columns[1]->data_w());
+
+//TODO: handle the case when the column is constant, i.e. min = max
```
st-pasha: Please indent comments (here and below) at the same level as the surrounding code, otherwise it messes up with code-folding in my editor.

c/extras/aggregator.h
```diff
@@ -0,0 +1,46 @@
...
+    double calculate_distance(double*, double*, int64_t, double);
+    void adjust_radius(int32_t, double&);
+    double* generate_pmatrix(int32_t, unsigned int);
+    void project_row(double*, int32_t, double*, int32_t);
+};
+
+
+DECLARE_FUNCTION(
+  aggregate,
+  "aggregate()\n\n",
```
st-pasha: This argument is the function's docstring. In your case the function has signature `OiiiiI`, so it would be good to list what parameters the function actually takes.

datatable/frame.py
```diff
@@ -563,8 +563,7 @@ def _delete_columns(self, cols):
             newnames += self.names[(cols[i - 1] + 1):cols[i]]
         newnames += self.names[cols[-1] + 1:]
         self._fill_from_dt(self._dt, names=newnames)
-
-
+
```
st-pasha: Looks like your editor doesn't have "Trim trailing whitespace" option turned on.

c/expr/reduceop.cc
oleksiyskononenko: Reimplemented this function to match the existing name.

c/extras/aggregator.h
oleksiyskononenko: Done.

datatable/frame.py
oleksiyskononenko: Done. Actually C++ editor had this option turned on, had to fix the Python one only.

c/expr/reduceop.cc
```diff
@@ -77,6 +78,31 @@ static void sum_skipna(const int32_t* groups, int32_t grp, void** params) {
...
+  const IT* inputs = static_cast<const IT*>(col0->data());
+  OT* outputs = static_cast<OT*>(col1->data_w());
+  OT count = 0;
+  int32_t row0 = groups[grp];
+  int32_t row1 = groups[grp + 1];
+  col0->rowindex().strided_loop(row0, row1, 1,
+    [&](int64_t i) {
+      IT x = inputs[i];
+      if (!ISNA<IT>(x))
+        ++count;
```
st-pasha: `count += !ISNA<IT>(x);` should be faster (in theory), as it avoids branching

oleksiyskononenko: Right, fixed.

st-pasha:
See also #1177 for further suggestions for improvement (after this PR is merged)

PR #1160, opened 2018-07-12
Stype/Ltype refactoring

### LINKED ISSUE #1152: Convert SType / LType into enum classes
Enum classes are more type-safe alternatives to C-style enums.

st-pasha:
* Convert `SType` / `LType` into enum classes (before they were plain-C enums)
* Replaced `stype_info[...]` with `info(stype)` class which provides same information, and even some more
* Names of SType constants were shortened and simplified. For example, we now have `SType::STR32` instead of `ST_STRING_I4_VCHAR`.
* The `.window()` method of a DataTable now returns a list of stype constants, previously it returned a list of stype "short names". This was a leftover from the era when stypes were strings.
* Eliminated some obsolete code

Closes #1152

### DIFF
--- a/c/column.cc
+++ b/c/column.cc
@@ -22,17 +22,17 @@ Column::Column(int64_t nrows_)
 
 Column* Column::new_column(SType stype) {
   switch (stype) {
-    case ST_VOID:            return new VoidColumn();
-    case ST_BOOLEAN_I1:      return new BoolColumn();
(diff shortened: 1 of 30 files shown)


oleksiyskononenko:
When this is going to be merged to master? I will probably have to make the corresponding changes to aggregator.

st-pasha:
Right now, if you don't have any objections.
The tests have all passed.

PR #1166, opened 2018-07-16
Code reorganization for GenericReader

st-pasha:
This is an attempt to make different pieces of code easier to locate.
This PR puts several of the classes that comprise core of GenericReader functionality each into its own file, and within the `dt::read::` namespace.
This code cleanup is not final: further reorganizations will be needed.


c/read/column.cc
```diff
@@ -0,0 +1,280 @@
...
+  fields[2].name = nullptr;
+  fields[0].doc = nullptr;
+  fields[1].doc = nullptr;
+  fields[2].doc = nullptr;
+  desc.name = const_cast<char*>(tuple_name);
+  desc.doc = nullptr;
+  desc.fields = fields;
+  desc.n_in_sequence = 2;
+  // Do not use PyStructSequence_NewType, because it is buggy
+  // (see https://lists.gt.net/python/bugs/1320383)
```
oleksiyskononenko: Seems to be fixed as of today: https://bugs.python.org/issue28709

PR #1184, opened 2018-07-19
Preserve the original column types in the aggregated frame

### LINKED ISSUE #1183: Preserve original column types in the aggregated table
Currently aggregator converts all the continuous columns to double, this should be fixed, so that it preserves the original column datatypes.

oleksiyskononenko:
The input frame is now aggregated in-place preserving the original
column types. Based on the input, the members frame is created by the
aggregator and returned to Python.

Closes #1183

### DIFF
--- a/c/extras/aggregator.cc
+++ b/c/extras/aggregator.cc
@@ -24,35 +24,13 @@ PyObject* aggregate(PyObject*, PyObject* args) {
                         &dt, &n_bins, &nx_bins, &ny_bins, &max_dimensions, &seed)) return nullptr;
 
   DataTable* dt_in = PyObj(dt).as_datatable();
-  DataTable* dt_exemplars = nullptr;
-  DataTable* dt_members = nullptr;
-  Column** cols_exemplars = dt::amalloc<Column*>(dt_in->ncols + 1);
-  Column** cols_members = dt::amalloc<Column*>(static_cast<int64_t>(2));
+  DataTablePtr dt_members = nullptr;
 
-  for (int32_t i = 0; i < dt_in->ncols; ++i) {
-    LType ltype = info(dt_in->columns[i]->stype()).ltype();
-    switch (ltype) {
-      case LType::BOOL:
-      case LType::INT:
-      case LType::REAL: cols_exemplars[i] = dt_in->columns[i]->cast(SType::FLOAT64); break;
-      default:          cols_exemplars[i] = dt_in->columns[i]->shallowcopy();
-    }
-  }
-
-  cols_exemplars[dt_in->ncols] = nullptr;
-  cols_members[0] = Column::new_data_column(SType::INT32, dt_in->nrows);
-  cols_members[1] = nullptr;
-  dt_exemplars = new DataTable(cols_exemplars);
-  dt_members = new DataTable(cols_members);
 
   Aggregator agg(n_bins, nx_bins, ny_bins, max_dimensions, seed);
-  agg.aggregate(dt_exemplars, dt_members);
+  dt_members = DataTablePtr(agg.aggregate(dt_in));
 
-  PyObject *res = PyTuple_New(2);
-  PyTuple_SET_ITEM(res, 0, pydatatable::wrap(dt_exemplars));
-  PyTuple_SET_ITEM(res, 1, pydatatable::wrap(dt_members));
-
-  return res;
+  return pydatatable::wrap(dt_members.release());
 }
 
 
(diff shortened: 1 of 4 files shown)


c/extras/aggregator.cc
```diff
@@ -24,35 +24,13 @@ PyObject* aggregate(PyObject*, PyObject* args) {
...
   Aggregator agg(n_bins, nx_bins, ny_bins, max_dimensions, seed);
-  agg.aggregate(dt_exemplars, dt_members);
+  dt_members = DataTablePtr(agg.aggregate(dt_in));
 
-  PyObject *res = PyTuple_New(2);
-  PyTuple_SET_ITEM(res, 0, pydatatable::wrap(dt_exemplars));
-  PyTuple_SET_ITEM(res, 1, pydatatable::wrap(dt_members));
-
-  return res;
+  return pydatatable::wrap(dt_members.release());
```
st-pasha: Not sure what's the point in creating an `std::unique_ptr<>` only to release in the very next line.

c/extras/aggregator.cc
```diff
@@ -66,13 +44,37 @@ Aggregator::Aggregator(int32_t n_bins_in, int32_t nx_bins_in, int32_t ny_bins_in
...
+      case LType::REAL: cols_double[i] = dt->columns[i]->cast(SType::FLOAT64); break;
+      default:          cols_double[i] = dt->columns[i]->shallowcopy();
+    }
+  }
+
+  cols_double[dt->ncols] = nullptr;
+  cols_members[0] = Column::new_data_column(SType::INT32, dt->nrows);
+  cols_members[1] = nullptr;
+  dt_double = new DataTable(cols_double);
+  dt_members = new DataTable(cols_members);
```
st-pasha: Here, on the other hand, it would make perfect sense to have `dt_double` and `dt_members` as `std::unique_ptr`s. Because currently if any of the `group_*` functions fails, the constructed `DataTable`s will leak.

The `dt_members` can then be returned as `std::unique_ptr` as well.

datatable/extras/aggregate.py
```diff
@@ -9,7 +9,8 @@
 from datatable.lib import core
 
 def aggregate(self, n_bins=500, nx_bins=50, ny_bins=50, max_dimensions=50, seed=0):
```
st-pasha: Perhaps you should add some documentation here what this function does and what it returns. Because that' not trivial

datatable/extras/aggregate.py
```diff
@@ -9,7 +9,8 @@
...
-  dt_exemplars, dt_members = core.aggregate(self._dt, n_bins, nx_bins, ny_bins, max_dimensions, seed)
+  dt_members = core.aggregate(self._dt, n_bins, nx_bins, ny_bins, max_dimensions, seed)
   names_exemplars = self.names + ("count",)
   names_members = ("exemplar_id")
-  return Frame(dt_exemplars, names_exemplars), Frame(dt_members, names_members)
\ No newline at end of file
+  self.__init__(self.internal, names_exemplars)
+  return Frame(dt_members, names_members)
```
st-pasha: By convention (PEP 8) all python files should be indented at 4 spaces.

At the same time we indent all C++ files at 2 spaces, so there's that...

oleksiyskononenko: Got it. Fixed.

c/extras/aggregator.cc
oleksiyskononenko: Yes, that's just an artifact. Fixed.

datatable/extras/aggregate.py
oleksiyskononenko: What is the best place to add it? 

I've been thinking this is a convenient way of defining this function. If one only needs the aggregated table it could be as simple as `aggregate(dt)`. If, in addition, one also needs the members it is `dt_members = aggregate(dt)`. If I do aggregation in-place, not sure if there are any more intuitive solutions. Any suggestions?

st-pasha: I'm not saying the API is bad, just that it needs some form of documentation. See for example here: https://github.com/h2oai/datatable/blob/master/datatable/frame.py#L640-L653

oleksiyskononenko: Sure, I got your point. Will add some documentation to this function. Thanks!

c/extras/aggregator.cc
oleksiyskononenko: Right, I guess this is a reason why I kept `dt_members` as `DataTablePtr` in `PyObject* aggregate(PyObject*, PyObject* args)`. Was not sure what is the best place to create the required tables and converged somewhere in the middle. Fixing it now.

c/extras/aggregator.cc
st-pasha: `dt_members` still remains a `std::unique_ptr`, and you're still calling `.release()` on it...

datatable/extras/aggregate.py
```diff
@@ -9,8 +9,30 @@
...
+        Number of columns at which start using the projection method.
+    seed: int
+        Seed to be used for the projection method.
+    
+    Returns
+    -------
+    The target datatable is aggregated in-place and gets an additional column `count` 
+    with the number of members for each exemplar. The function returns
+    a new one-column datatable that contains exemplar_ids for each of the original rows. 
+    """
```
st-pasha: :+1:

PR #1186, opened 2018-07-20
Properly skip categorical columns during the ND aggregation

### LINKED ISSUE #1185: Properly skip categorical columns during the ND aggregation
Until we find a good way to treat categorical columns in the ND aggregation algo, we should properly skip those in the both cases: `ncols` <= `max_dimensions` and `ncols` > `max_dimensions`. This was missing in the initial implementation of the `aggregator` and must be fixed.

oleksiyskononenko:
Minor change, but prevents the `aggregator` from crashing.
Until we find a good way to treat categorical columns in the ND
aggregation algo, we skip those in the both cases: `ncols` <=`max_dimensions` and 
`ncols` > `max_dimensions`. 

Closes #1185

### DIFF
--- a/c/extras/aggregator.cc
+++ b/c/extras/aggregator.cc
@@ -325,8 +325,10 @@ void Aggregator::group_nd(DataTablePtr& dt_exemplars, DataTablePtr& dt_members) 
 void Aggregator::adjust_radius(DataTablePtr& dt_exemplars, double& radius) {
   double diff = 0;
   for (int i = 0; i < dt_exemplars->ncols; ++i) {
-    RealColumn<double>* ci = static_cast<RealColumn<double>*>(dt_exemplars->columns[i]);
-    diff += (ci->max() - ci->min()) * (ci->max() - ci->min());
+    if (dt_exemplars->columns[i]->stype() == SType::FLOAT64) {
+      RealColumn<double>* ci = static_cast<RealColumn<double>*>(dt_exemplars->columns[i]);
+      diff += (ci->max() - ci->min()) * (ci->max() - ci->min());
+    }
   }
 
   diff /= dt_exemplars->ncols;
@@ -352,10 +354,13 @@ double Aggregator::calculate_distance(double* e1, double* e2, int64_t ndims, dou
 
 void Aggregator::normalize_row(DataTablePtr& dt, double* r, int32_t row_id) {
   for (int32_t i = 0; i < dt->ncols; ++i) {
-    RealColumn<double>* c = static_cast<RealColumn<double>*>(dt->columns[i]);
-    const double* d_c = static_cast<const double*>(dt->columns[i]->data());
-
-    r[i] =  (d_c[row_id] - c->min()) / (c->max() - c->min());
+    if (dt->columns[i]->stype() == SType::FLOAT64) {
+      RealColumn<double>* c = static_cast<RealColumn<double>*>(dt->columns[i]);
+      const double* d_c = static_cast<const double*>(dt->columns[i]->data());
+      r[i] =  (d_c[row_id] - c->min()) / (c->max() - c->min());
+    } else {
+      r[i] = 0;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


c/extras/aggregator.cc
```diff
@@ -385,12 +390,15 @@ void Aggregator::project_row(DataTablePtr& dt_exemplars, double* r, int32_t row_
   std::memset(r, 0, static_cast<size_t>(max_dimensions) * sizeof(double));
 
   for (int32_t i = 0; i < (dt_exemplars->ncols) * max_dimensions; ++i) {
```
st-pasha: Looks like what you really wanted to have here is a double-loop with `col_id` running from 0 to `dt_exemplars->ncols`, and `row_id` from 0 to `max_dimensions`. Then you won't have to recompute `c`, `d_c`, etc on each iteration.

oleksiyskononenko: At some point I had a double loop here for the `double**` matrix and then replaced it with the `double*` one. Yes, obviously need to take another look at its implementation. Thanks!

st-pasha: You can still have a single-dimensional matrix, indexed by `irow * nrows +  icol` or whatever

oleksiyskononenko: Yes, sure, that is what I missed at the time of re-implementation...

oleksiyskononenko: I guess it had something to do with the matrix by vector multiplication I do here. I will double check it. Thanks.

c/extras/aggregator.cc
```diff
@@ -389,15 +389,17 @@ double* Aggregator::generate_pmatrix(DataTablePtr& dt_exemplars) {
...
+    if (dt_exemplars->columns[i]->stype() == SType::FLOAT64) {
+      const double* d_c = static_cast<const double*>(dt_exemplars->columns[i]->data());
       if (!ISNA<double>(d_c[row_id])) {
-        // TODO: handle missing values and do r[j] /= n normalization at the end
-        r[i % max_dimensions] +=  pmatrix[i] * (d_c[row_id] - c->min()) / (c->max() - c->min());
+
+        RealColumn<double>* c = static_cast<RealColumn<double>*> (dt_exemplars->columns[i]);
+        for (int32_t j = 0; j < max_dimensions; ++j) {
+          //TODO: handle missing values and do r[j] /= n normalization at the end
+          r[j] +=  pmatrix[i * max_dimensions + j] * (d_c[row_id] - c->min()) / (c->max() - c->min());
```
st-pasha: `(d_c[row_id] - c->min()) / (c->max() - c->min())` remains constant during the loop, so why not precompute it up-front?

c/extras/aggregator.cc
```diff
@@ -389,15 +389,17 @@ double* Aggregator::generate_pmatrix(DataTablePtr& dt_exemplars) {
...
 
-  for (int32_t i = 0; i < (dt_exemplars->ncols) * max_dimensions; ++i) {
-    int col_id = i / max_dimensions;
-    if (dt_exemplars->columns[col_id]->stype() == SType::FLOAT64) {
-      RealColumn<double>* c = static_cast<RealColumn<double>*> (dt_exemplars->columns[col_id]);
-      const double* d_c = static_cast<const double*>(dt_exemplars->columns[col_id]->data());
-
+  for (int32_t i = 0; i < dt_exemplars->ncols; ++i) {
+    if (dt_exemplars->columns[i]->stype() == SType::FLOAT64) {
+      const double* d_c = static_cast<const double*>(dt_exemplars->columns[i]->data());
```
st-pasha: A minor note: whenever you have `TYPE var = static_cast<TYPE>(...)` it's better to declare the variable as `auto`. That way you repeat yourself less, and the line is shorter too:
```C++
auto d_c = static_cast<const double*>(dt_exemplars->columns[i]->data());
// vs
const double* d_c = static_cast<const double*>(dt_exemplars->columns[i]->data());
``` 

Also, in this particular case you could have declared `d_c` first, and then
```C++
const double* d_c = c->elements_r();
```

c/extras/aggregator.cc
oleksiyskononenko: This has already been re-implemented as a part of  #1189 as this formula won't work in the case of the constant columns. So will be fixed with the next PR. Is it ok?

c/extras/aggregator.cc
oleksiyskononenko: Sure, fixing it now.

oleksiyskononenko: As for your second comment, I guess you meant 
>you could have declared `c` first
?

If so, it could be the case that this particular value is missing and we don't really need to declare `c` at all. Probably in most of the cases the value is not missing, so what you say makes perfect sense.

PR #1196, opened 2018-07-24
Handle `NaN`s in aggregator

### LINKED ISSUE #1195: Handle NaN values in the aggregator
For columns having only `NaN` values, that are loaded from `.csv`, `datatable` assigns `bool` datatype. This should be properly handled by the aggregator as `min` and `max` are also `NaN` in this case. 

Some `NaN` related adjustments are also required for other column types.

oleksiyskononenko:
- Ignore or replace `NaN`s with zeros depending on a particular
aggregator function.

- Small adjustment to the radius formula based on the Leland's
suggestions.

Closes #1195.


oleksiyskononenko:
Pasha, can we merge this to `master`, so that I do another PR based on these changes? Thanks!

PR #1202, opened 2018-07-28
Check that __version__ and __git_revision__ fields are non-empty

st-pasha:
This should help with https://github.com/h2oai/h2oai/issues/4180

### DIFF
--- a/tests/test_dt.py
+++ b/tests/test_dt.py
@@ -103,6 +103,15 @@ def test_dt_dependencies():
     assert int(n) < 200
 
 
+@pytest.mark.run(order=0.91)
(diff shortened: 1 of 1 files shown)


michal-raska:
@st-pasha I pushed a fix, please review. The issue is, we have `.git` directory only available in the first stage, it's not being stashed (and I don't think it should be). The modification is, to generate the `__git__.py` file while we are still in workspace with `.git` directory.

arnocandel:
no __git__.py file packaged:
image

st-pasha:
@arnocandel Are you looking at the most recent build of datatable (one after this PR)?

PR #1205, opened 2018-07-30
Stats integrity checks

### LINKED ISSUE #1143: Add integrity checks for Stats class
Currently, the `verify_integrity()` method of this class is empty.

General approach should be: create new Stats class for the column, for each computed stat in the current class, re-compute this same stat in the new class, and then check that the two quantities are equal. Also check basic properties such as `nacount > 0`, `sum = n * mean`, `min <= mean <= max`, etc.

st-pasha:
Adding tests to ensure that various stats would not get corrupted if a column is modified.
Since Stats are cached, it is easy to forget to invalidate the cache when the column is modified. These tests will attempt to catch such an occurrence.

Currently, the checks for StDev, Kurtosis, and Skewness are turned off, due to algos for their calculation being numerically unstable. This needs to be addressed in the future.

Also, some aggregator tests were marked XFAILing, because they are now failing.

Closes #1143

### DIFF
--- a/c/extras/aggregator.cc
+++ b/c/extras/aggregator.cc
@@ -112,6 +112,7 @@ void Aggregator::aggregate_exemplars(DataTable* dt_exemplars, DataTablePtr& dt_m
       }
     }
   }
+  dt_members->columns[0]->get_stats()->reset();
 
   // Applying exemplars row index and binding exemplars with the counts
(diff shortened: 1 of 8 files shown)


st-pasha:
No, it's the next commit, e617fb6

oleksiyskononenko:
Fixed the aggregator code accordingly, let me know if we need to do anything else.

Issue #1188, opened 2018-07-21
Allow each Column in a DataTable have its own RowIndex


st-pasha:
This is needed to implement efficient joins: the columns from the Left frame and the columns from the Right frame will have different RowIndices applied to them. 
This feature may also be handy when selecting regular columns alongside with the computed ones.

This change will remove the `rowindex` property in `DataTable` object, possibly replacing it with `nrowindices` field counting the number of different rowindices for all columns in the frame.

st-pasha:
Prerequisite: #1477

PR #1218, opened 2018-08-02
Mix rowindices

st-pasha:
Issue #1188 appears to be much harder than it seemed at the first sight, so I'll attempt to implement it in several batches.

* Internal method `columns_from_array()` removed -- same functionality can be achieved via `columns_from_columns()`;
* `ColumnSet` objects can now contain columns with different `RowIndex`es (however `DataTable`s still can't);
* Fixed a bug in DataTable's constructor where the last column was not checked properly during construction (the bug never manifested itself though);
* Python-facing `_Column` object acquired new property `rowindex`, and 2 new methods: `replace_rowindex()` and `topython()` (the last one mostly useful for debug purposes);
* `DataTable.isview` semantics has changed: now it returns True if *any* of the columns has a RowIndex (this change is backward-compatible);
* Python-facing `_RowIndex` object acquired property `ptr`, which can be used in python to compare whether to`RowIndex`es are equal;
* Methods `make_*()` in `EvaluationEngine` were removed, since they served as a useless indirection level;
* `SliceCSNode` no longer uses `columns_from_slice()` method (preparing to remove the latter);
* Added methods to `EvaluationEngine` to distinguish between "source"/"final" RowIndexes -- these are still work in progress, and currently coexist with the old-style `.rowindex` property;
* Attempt to disentangle final/source RowIndex calculation in RowsNode and GroupbyNode -- still WIP;
* Several places where column's "meta" was mentioned were cleaned up -- there is no such concept anymore.

WIP for #1188

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -900,7 +900,7 @@ $(BUILDDIR)/py_buffers.o : c/py_buffers.cc $(BUILDDIR)/column.h $(BUILDDIR)/enco
 	@echo • Compiling $<
 	@$(CC) -c $< $(CCFLAGS) -o $@
 
-$(BUILDDIR)/py_column.o : c/py_column.cc $(BUILDDIR)/py_column.h $(BUILDDIR)/py_rowindex.h $(BUILDDIR)/py_types.h $(BUILDDIR)/utils/pyobj.h $(BUILDDIR)/writebuf.h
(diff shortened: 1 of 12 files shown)


oleksiyskononenko:
With this you don't remove the `rowindex` property from `DataTable` (at least for the moment)?

st-pasha:
DataTable's `.rowindex` remains, for now. But it'll be gone in the near future (see description for #1188).

Issue #1220, opened 2018-08-03
Create new mechanism for translating Python to C++ code


st-pasha:
Currently, in order to define Python-facing objects, we follow the standard Python methodology for defining an extension type (see eg https://docs.python.org/3/extending/newtypes.html). The problem with this approach are:
* This is idiomatic CPython, but not idiomatic C++. The code is hard to write and read.
* All methods have to be wrapped in macro-based wrappers that capture any C++ exceptions and translate them into Python code.
* Retrieving arguments/keyword arguments in methods is inefficient (it involves comparison of strings and format parameter scanning).
* Doesn't integrate well with `PyObj` objects.

The new system should aim to correct all these deficiencies. The expected API for defining an extension type is:
```C++
class Frame : public BasePyType<Frame> {
  DataTable* dt;

public:
  void m__init__(Args) override;
  void m__dealloc__() override;
  PyObj m__repr__() override;
  
  PyObj my_method1(Args);
  void my_method2(Args);
 
  static const char* classname() { return "datatable.Frame"; }
  static const char* classdoc() { return nullptr; }
  static void init_methods();
  static void init_getsetters();
};
```

This is a prerequisite for #1066

PR #1222, opened 2018-08-07
Py framework

st-pasha:
This is a new framework for defining Python objects from C++. It intends to supplement the existing methodology (see py_utils.h), which is largely based on an assortment of macros.

The framework is not yet complete. The following is currently possible:
* Define a new python extension type by deriving from `py::ExtType`;
* Specify constructor/destructor/repr functions by defining methods with predefined names, such as `void m__dealloc__()`;
* Define any number of getters/setters, and then declare them via `static void init_getsetters(GetSetters&)`;
* Define any number of methods, and then declare them via `static void init_methods(Methods&)`;
* All methods are automatically wrapped into exception-catching blocks;
* Methods can receive an arbitrary number of parameters and keyword arguments, as controlled by the `Args` parameter for each method. The API to work with these args is still WIP.

-----
This new framework, besides being more convenient to work with, offers several advantages compared to the current technology:
* Easier and faster parsing of function arguments;
* Better error messages for parameters that have incorrect types;
* Less boilerplate code.

WIP for #1220

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -689,7 +689,7 @@ $(BUILDDIR)/python/args.h: c/python/args.h $(BUILDDIR)/python/arg.h
 	@echo • Refreshing c/python/args.h
 	@cp c/python/args.h $@
 
-$(BUILDDIR)/python/ext_type.h: c/python/ext_type.h $(BUILDDIR)/python/args.h $(BUILDDIR)/utils/exceptions.h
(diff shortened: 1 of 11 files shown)


oleksiyskononenko:
Great work. When the framework is completed, what will be the best way to parse the function arguments?

st-pasha:
It could look something like this:
```
void my_method(const PosAndKwdArgs& args) {
  DataTable* dt = args[0];
  RowIndex rowindex = args[1];
  int column = args.has(2)? args[2] : -1;
  ...
}
```

Though I'm not sure about the exact details of the API yet.

PR #1224, opened 2018-08-07
Move auto-generated parts out of the main makefile

st-pasha:
Previously `Makefile` contained an explicit dependency tree based on the source files. There was a separate rule for each `.h` and `.cc` file, and whenever the dependencies changed, those rules had to be updated manually. The `fastcheck.py` was used to verify that the dependencies are correct.

Now, the dependency tree file is generated by `make_fast.py` (which is a modified `fastcheck.py`), and no manual updates are necessary. As an added bonus, the auto-generated file needs no longer be in the repository, and changes to it no longer show up in the PRs.


Makefile
```diff
@@ -14,8 +21,19 @@ PLATFORM := $(ARCH)-$(OS)
 DIST_DIR := dist/$(PLATFORM)
 
 .PHONY: all clean mrproper build install uninstall test_install test \
-		benchmark debug bi coverage fast dist
-.SECONDARY: main-fast
+		benchmark debug bi coverage dist fast ci/fast.mk
+
+ifeq ($(MAKECMDGOALS), fast)
+-include ci/fast.mk
```
mmalohlava: hmm, i do not understand this one - this will include fast.mk and regenerate it in the next step?

oleksiyskononenko: As I've checked, If you remove `ci/fast.mk` from .PHONY, you won't need to run `make fast` twice after doing `make clean`. Was there a particular reason why you put it there?

PR #1229, opened 2018-08-09
[BUILD] Add tests for Python 3.7

michal-raska:


Makefile
```diff
@@ -243,10 +243,13 @@ ubuntu_docker_publish: Dockerfile-ubuntu.$(PLATFORM).tag
 	docker push $(CONTAINER_NAME_TAG)
 
 ARCH_NAME ?= $(shell uname -m)
-DOCKER_IMAGE_TAG ?= 0.6.0-PR-1210.1
+DOCKER_IMAGE_TAG ?= 0.6.0-PR-1229.4
```
mmalohlava: @michal-raska what do you think about removing the cache and pinning docker builders to a specific host. It would simplify lot of work.

PR #1230, opened 2018-08-09
Python framework, continued

### LINKED ISSUE #1220: Create new mechanism for translating Python to C++ code

st-pasha:
This PR introduces further improvements towards implementing the new Python-C++ framework. Although it is far from complete, I anticipate that further changes will be done on case-by-case need, and there won't be a point in time when we could definitely say "the framework is ready, now the issue can be closed".

Closes #1220

### DIFF
--- a/c/python/args.cc
+++ b/c/python/args.cc
@@ -7,7 +7,7 @@
 //------------------------------------------------------------------------------
 #include "python/args.h"
 #include <cstring>             // std::strrchr
-#include "utils/exceptions.h"
+#include "utils/assert.h"
 
 namespace py {
 
@@ -85,7 +85,12 @@ PKArgs::PKArgs(
     has_varkwds(vkwds),
     arg_names(_names)
 {
+  xassert(n_args == arg_names.size());
+  if (has_varargs) xassert(n_pos_kwd_args == 0);
   bound_args.resize(n_args);
+  for (size_t i = 0; i < n_args; ++i) {
+    bound_args[i].init(i, this);
+  }
 }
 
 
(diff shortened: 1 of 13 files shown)


c/python/args.cc
```diff
@@ -39,35 +63,114 @@ Arg Args::get(const char*) const { throw AssertionError(); }
...
+    has_varkwds(vkwds),
+    arg_names(_names)
+{
+  xassert(n_args == arg_names.size());
+  if (has_varargs) xassert(n_pos_kwd_args == 0);
+  bound_args.resize(n_args);
+  for (size_t i = 0; i < n_args; ++i) {
+    bound_args[i].init(i, this);
+  }
+}
```
oleksiyskononenko: `n_kwdonly_args` and `has_varkwds` are not used for the moment, right?

st-pasha: That's right, as the warning message will keep pestering you about:
```
c/python/args.h:73:18: warning: private field 'n_kwdonly_args' is not used [-Wunused-private-field]
    const size_t n_kwdonly_args;
                 ^
c/python/args.h:76:18: warning: private field 'has_varkwds' is not used [-Wunused-private-field]
    const bool   has_varkwds;
                 ^
2 warnings generated.
```

PR #1237, opened 2018-08-14
py::obj API cleanup

### LINKED ISSUE #1231: Update API of PyObj
* Move into "python/obj.h|cc" file
* Rename into `py::_obj`
* Make 2 derived classes: `py::bobj` and `py::oobj` ("borrowed" and "owned" reference). If possible, these classes should not be virtual.
* Other classes (such as `py::string`, `py::long`, `py::list`, etc) should derive from `py::?obj`.

st-pasha:
Closes #1231


c/column_from_pylist.cc
```diff
@@ -255,21 +253,21 @@ static bool parse_as_str(PyyList& list, MemoryRange& offbuf,
...
     if (item.is_none()) {
       offsets[i] = curr_offset | GETNA<T>();
       continue;
     }
     if (item.is_string()) {
-      size_t len = 0;
-      const char* cstr = item.as_cstring(&len);
-      if (len) {
-        T tlen = static_cast<T>(len);
+      CString cstr = item.to_cstring();
```
oleksiyskononenko: First thought this is the MS `CString` class and was not sure how it is going to work :)

oleksiyskononenko:
So are we going to replace all the calls to `PyArg_ParseTuple` with the new API?

st-pasha:
Eventually, yes

c/extras/aggregator.cc
```diff
@@ -5,25 +5,24 @@
...
   unsigned int seed;
-  PyObject* dt;
+  PyObject* arg1;
 
   if (!PyArg_ParseTuple(args, "OiiiiI:aggregate",
-                        &dt, &n_bins, &nx_bins, &ny_bins, &max_dimensions, &seed)) return nullptr;
+                        &arg1, &n_bins, &nx_bins, &ny_bins, &max_dimensions, &seed)) return nullptr;
 
-  DataTable* dt_in = PyObj(dt).as_datatable();
+  DataTable* dt_in = py::obj(arg1).to_frame();
```
oleksiyskononenko: Thanks for adjusting this part

c/options.cc
```diff
@@ -99,36 +98,36 @@ PyObject* set_option(PyObject*, PyObject* args) {
...
 
   } else if (name == "core_logger") {
-    set_core_logger(value.as_pyobject());
+    set_core_logger(py::oobj(value).release());
 
   } else if (name == "fread.anonymize") {
-    set_fread_anonymize(value.as_bool());
+    set_fread_anonymize(value.to_bool_strict());
 
   } else {
```
oleksiyskononenko: This whole `if` looks a little bit scary to me...

st-pasha: There are no switch-over-strings in C++. Of course, we can use a map to convert strings into enum values and then switch over those -- but setting up the map just seems like more work. Maybe when it gets bigger...

PR #1242, opened 2018-08-15
Fix problems with stringification of f-expressions

### LINKED ISSUE #1241: f- column selectors may produce invalid ids or throw errors
```
>>> str(f["A"])
'f_A'
>>> f.bind_datatable(dt.Frame())
FrameProxy('f', <Frame [0 rows x 0 cols]>)
>>> str(f["A"])
  File "/Users/pasha/github/datatable/datatable/expr/column_expr.py", line 55, in __str__
    # careful to ensure that this method never throws an error, even if the
TypeError: tuple indices must be integers or slices, not str
```
or
```
>>> str(f[-1])
'f_-1'
```

st-pasha:
Closes #1241

### DIFF
--- a/tests/test_dt_expr.py
+++ b/tests/test_dt_expr.py
@@ -34,18 +34,91 @@ dt_str = {("foo", "bbar", "baz"),
 #-------------------------------------------------------------------------------
 
 def test_f():
-    # Check that unbounded f-expressions can be stringified (see #1024). The
-    # exact representation may be modified in the future; however f-expressions
+    assert str(f) == "f"
+    assert repr(f) == "FrameProxy('f', None)"
+    with f.bind_datatable(dt.Frame()):
+        assert str(f) == "f"
+        assert repr(f) == "FrameProxy('f', <Frame [0 rows x 0 cols]>)"
+    assert repr(f) == "FrameProxy('f', None)"
+
+
+def test_f_col_selector_unbound():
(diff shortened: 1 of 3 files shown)


tests/test_dt_expr.py
```diff
@@ -34,18 +34,91 @@
...
+    dt0 = dt.Frame([[]] * 3, names=["a", "abcdefg", "abcdefghijklm"])
+    with f.bind_datatable(dt0):
+        assert str(f.a) == "f_a"
+        assert str(f.abcdefg) == "f_abcdefg"
+        assert str(f.abcdefghijkl) == "f_abcdefghijkl"
+        assert str(f.abcdefghijklm) == "f_2"
+        assert str(f[0]) == "f_a"
+        assert str(f[1]) == "f_abcdefg"
+        assert str(f[2]) == "f_2"  # "f_abcdefgijklm" is too long
+        assert str(f[3]) == "f_3"
```
oleksiyskononenko: When there are only 3 columns, what f[3] is supposed to return?

st-pasha: If `f` is bound to a frame with only 3 columns, then `f[3]` refers to a non-existing column. This is valid from `f`'s perspective, however, once you try to use `f[3]` in any context where its value is needed -- an error will be thrown. One place where it is legal to use a non-existing column is when you're assigning to that column.

PR #1243, opened 2018-08-15
Allow to use join columns within update expressions

### LINKED ISSUE #1240: Column assign with join should work properly
Currently, it produces errors such as
```
>>> d0[:, "new", join(d1)] = g.V0
  File "/Users/pasha/github/datatable/datatable/frame.py", line 550, in __setitem__
    mode="update", replacement=value)
  File "/Users/pasha/github/datatable/datatable/graph/__init__.py", line 88, in make_datatable
    replacement = datatable.Frame(replacement)
  File "/Users/pasha/github/datatable/datatable/frame.py", line 58, in __init__
    self._fill_from_source(src, names=names, stypes=stypes)
  File "/Users/pasha/github/datatable/datatable/frame.py", line 219, in _fill_from_source
    raise TTypeError("Cannot create Frame from %r" % src)
  File "/Users/pasha/github/datatable/datatable/expr/base_expr.py", line 264, in __repr__
    return "%s(%s)" % (self.__class__.__name__, self)
  File "/Users/pasha/github/datatable/datatable/expr/column_expr.py", line 55, in __str__
    colname = self._dtexpr.names[self._colid]
TypeError: tuple indices must be integers or slices, not str
```

st-pasha:
Closes #1240

### DIFF
--- a/tests/test_join.py
+++ b/tests/test_join.py
@@ -8,7 +8,7 @@ import datatable as dt
 import pytest
 import random
 from tests import random_string
-from datatable import join, ltype, stype
+from datatable import join, ltype, stype, f, g, mean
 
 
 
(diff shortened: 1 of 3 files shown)


tests/test_join.py
```diff
@@ -105,3 +105,17 @@ def test_join_random(seed, lt):
...
     assert djoined.names == ("KEY", "VAL")
     assert djoined.topython() == [main, res]
+
+
+
+def test_join_update():
+    d0 = dt.Frame([[1, 2, 3, 2, 3, 1, 3, 2, 2, 1], range(10)], names=("A", "B"))
+    d1 = d0[:, mean(f.B), f.A]
+    d1.key = "A"
+    d0[:, "AA", join(d1)] = g.V0
```
jangorecki: update made is by reference, correct?

st-pasha: The LHS table (`d0`) is updated in-place. The RHS needs to be computed in order to match the shape of the LHS.

oleksiyskononenko: `"AA"` here means that for `join` we're going to use a column `A` from `d0` with the column `A` from `d1`?

oleksiyskononenko: Sorry, got it. This is just the name of the new column to be created...

PR #1252, opened 2018-08-20
Perform name deduplication in C++

st-pasha:
This should improve performance, especially for frames where the number of rows is small, but the number of columns large.

### DIFF
--- a/datatable/extras/aggregate.py
+++ b/datatable/extras/aggregate.py
@@ -24,15 +24,15 @@ def aggregate(self, n_bins=500, nx_bins=50, ny_bins=50, max_dimensions=50, seed=
         Number of columns at which start using the projection method.
     seed: int
         Seed to be used for the projection method.
-    
+
     Returns
     -------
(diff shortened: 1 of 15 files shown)


datatable/extras/aggregate.py
```diff
@@ -24,15 +24,15 @@ def aggregate(self, n_bins=500, nx_bins=50, ny_bins=50, max_dimensions=50, seed=
...
+    a new one-column datatable that contains exemplar_ids for each of the original rows.
     """
     names_exemplars = self.names + ("count",)
-    names_members = ("exemplar_id")
+    names_members = ("exemplar_id",)
     dt_members = core.aggregate(self._dt, n_bins, nx_bins, ny_bins, max_dimensions, seed)
     self.__init__(self.internal, names_exemplars)
-    return Frame(dt_members, names_members)
\ No newline at end of file
+    return Frame(dt_members, names=names_members)
```
oleksiyskononenko: Thanks for the update

PR #1255, opened 2018-08-22
Core Frame class improvements

st-pasha:
* Added `varargs` and `varkwds` iterators to `PKArgs` class;
* Implement getters/setters for core `Frame` class properties `.nrows`, `.ncols` and `.shape`;
* Python `Frame` class now inherits from core `Frame` class, allowing incremental transition for #1066;
* Removed from python `Frame` properties `.nrows`, `.ncols` and `.shape` since they are now inherited;
* Small improvements to `py::obj` class;
* The `Args` class now better integrates with `py::obj`;
* Added `py::otuple` class for easier tuple creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 #### Changed
 - `names` argument in `Frame()` constructor can no longer be a string --
   use a list or tuple of strings instead.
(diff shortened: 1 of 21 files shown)


oleksiyskononenko:
So the idea is to move as much of the Python code as possible to C++?

st-pasha:
@oleksiyskononenko yes

PR #1256, opened 2018-08-23
core.Frame name-related functionality improvements

st-pasha:
This PR continues work to bring some of the Python functionality over to C++ `core.Frame` class. In particular, in this PR:

* `Frame.rename()` is removed -- `Frame.name` setter can be used instead;
* Getters for `stypes` and `ltypes` moved to `core.Frame` class and converted to `py::obj` API;
* Getter/setter for `names` implemented in C++ in `core.Frame`;
* `.colindex()` method reimplemented in C++ in `core.Frame`
* Better documentation for `.names` setter;
* Created `py::odict` class;
* Added element setters for `py::list`;
* Minor API improvements to `py::_obj`, `py::tuple` and `py::string`;
* `info::py_stype()` and `info::py_ltype()` now return `py::oobj`s instead of `PyObject*`s;
* `_dt.materialize()` now an in-place operation;
* Added more tests related to column renaming.

### DIFF
--- a/datatable/frame.py
+++ b/datatable/frame.py
@@ -62,17 +62,8 @@ class Frame(core.Frame):
     def key(self):
         """Tuple of column names that comprise the Frame's key. If the Frame
         is not keyed, this will return an empty tuple."""
-        return self._dt.names[:self._dt.nkeys]
+        return self.names[:self._dt.nkeys]
 
-    @property
-    def names(self):
-        """Tuple of column names."""
-        return self._dt.names
-
-
-    #---------------------------------------------------------------------------
-    # Property setters
-    #---------------------------------------------------------------------------
 
     @key.setter
     def key(self, colnames):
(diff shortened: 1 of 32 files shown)


datatable/frame.py
```diff
@@ -165,23 +141,23 @@ def _fill_from_source(self, src, names, stypes):
...
-            src._set_names(names)
             self._dt = src
+            self.names = names
         elif isinstance(src, str):
             srcdt = datatable.fread(src)
             if names is None:
                 names = srcdt.names
             self._dt = srcdt.internal
-            self._dt._set_names(names)
+            self.names = names
```
oleksiyskononenko: So for the `aggregator` I should also use `self.names = names` rather than the `_set_names()` or any other options?

st-pasha: `_set_names()` doesn't really exist anymore, so yes, use the `names` setter

PR #1258, opened 2018-08-23
Cleanup of py:: API

### LINKED ISSUE #1257: Review API of py:: classes
The classes in question are:
- [x] `py::olist`
- [x] `py::otuple`
- [x] `py::odict`
- [x] `py::ostring`
- [x] `py::oint`
- [x] `py::ofloat`

We need to ensure they exhibit consistent API, all objects properly inherit from `py::oobj`, and that no API cruft remains.

st-pasha:
This PR enforces consistent API among all `py::` classes.
From this point onward we should try to use the new API only, and stay away from pure-C pythonic API.

Closes #1257

### DIFF
--- a/c/python/list.h
+++ b/c/python/list.h
@@ -14,19 +14,19 @@ namespace py {
 class Arg;
 
 
-class list : public oobj {
+class olist : public oobj {
   private:
     bool is_list;
     size_t : 56;
 
   public:
     using oobj::oobj;
-    list();
-    list(PyObject*);
-    list(size_t n);
+    olist();
+    olist(PyObject*);
+    olist(size_t n);
 
-    operator bool() const;
-    size_t size() const;
+    operator bool() const noexcept;
+    size_t size() const noexcept;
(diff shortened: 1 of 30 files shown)


oleksiyskononenko:
>From this point onward we should try to use the new API only, and stay away from pure-C pythonic >API.

Does it mean that I should also review my usage of all the types and functions that start with `Py`?

c/python/list.h
```diff
@@ -137,7 +53,12 @@ class list : public oobj {
     void set(int64_t i, oobj&& value);
     void set(int i,     oobj&& value);
 
-    friend class Arg;
+  private:
+    // Wrap an existing PyObject* into an `olist`. This constructor is private,
+    // use `py::org(src).to_pylist()` instead.
```
oleksiyskononenko: Is it a typo or what is `py::org`?

st-pasha: A typo. Should be `py::robj(src).to_pylist()`

PR #1265, opened 2018-08-31
Frame initialization improvements

st-pasha:
* Added method `frame.copy()`;
* Several modes of Frame initialization are now done in C++ instead of in Python;
* Creating a Frame from `core._DataTable` object is no longer supported (the use of these objects will be slowly phased out).

WIP for #1066

### DIFF
--- a/c/extras/aggregator.cc
+++ b/c/extras/aggregator.cc
@@ -8,6 +8,7 @@
 #define dt_EXTRAS_AGGREGATOR_cc
 #include "extras/aggregator.h"
 #include <cstdlib>
+#include "frame/py_frame.h"
 #include "py_utils.h"
 #include "python/obj.h"
 #include "rowindex.h"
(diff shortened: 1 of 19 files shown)


c/extras/aggregator.cc
```diff
@@ -22,20 +23,21 @@ PyObject* aggregate(PyObject*, PyObject* args) {
...
 
   Aggregator agg(min_rows, n_bins, nx_bins, ny_bins, nd_bins, max_dimensions,
                  seed, progress_fn);
-  dt_members = agg.aggregate(dt_in);
-  return pydatatable::wrap(dt_members.release());
+  DataTable* dt_out = agg.aggregate(dt_in).release();
+  py::Frame* frame = py::Frame::from_datatable(dt_out);
+  frame->set_names(py::obj(arg_names));
+
+  return frame;
```
oleksiyskononenko: I will just move all the column name assignments to C++. Thanks!

PR #1271, opened 2018-09-05
(py)names properties moved to core DataTable class

### LINKED ISSUE #1270: set_names() should work on core DataTable object
Currently, the core `DataTable` object contains a vector of column names, but in order to set them, you need to have access to the wrapper `Frame` class. This means that any manipulation to the column set of `DataTable` may potentially put the DataTable into an invalid state. Of course, we fix up any inconsistencies in the wrapper methods, but this is very fragile.

st-pasha:
* `Frame.names` moved to `DataTable.py_names`;
* `Frame.inames` moved to `DataTable.py_inames`;
* Cleaned up integrity checks for `.names`, `.py_names` and `.py_inames` in `DataTable`;
* `DataTable` has now multiple constructors, depending on how the names of the columns should be initialized;
* Added C++ tests to ensure proper coverage of integrity checks;
* `.open_jay()` no longer takes string vector reference as a parameter -- instead it directly assign names to the DataTable returned;
* `.cbind()` now produces DataTable with correct names already set;

Closes #1270

### DIFF
--- a/c/datatable.h
+++ b/c/datatable.h
@@ -9,6 +9,8 @@
 #define dt_DATATABLE_h
 #include <vector>
 #include <string>
+#include "python/dict.h"
+#include "python/tuple.h"
 #include "rowindex.h"
 #include "types.h"
 #include "column.h"
@@ -17,6 +19,7 @@
 class Column;
 class Stats;
 class DataTable;
+class NameProvider;
 
 typedef std::unique_ptr<DataTable> DataTablePtr;
 typedef Column* (Column::*colmakerfn)(void) const;
@@ -57,7 +60,11 @@ class DataTable {
     RowIndex rowindex;
     Groupby  groupby;
     Column** columns;
+
+  private:
     std::vector<std::string> names;
+    mutable py::otuple py_names;   // memoized tuple of column names
+    mutable py::odict  py_inames;  // memoized dict of {column name: index}
 
   public:
     DataTable(Column**);
@@ -70,6 +77,7 @@ class DataTable {
     void reify();
     void rbind(DataTable**, int**, int64_t, int64_t);
     DataTable* cbind(DataTable**, int64_t);
+    DataTable* copy() const;
     size_t memory_footprint();
 
     /**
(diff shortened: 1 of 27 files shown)


c/datatable.h
```diff
@@ -83,6 +95,15 @@ class DataTable {
...
     RowIndex sortby(const arr32_t& colindices, Groupby* out_grps) const;
 
+    const std::vector<std::string>& get_names() const;
+    py::otuple get_pynames() const;
+    int64_t colindex(const py::_obj& pyname) const;
+    void copy_names_from(const DataTable* other);
+    void set_names_to_default();
+    void set_names(const py::olist& names_list);
+    void set_names(const std::vector<std::string>& names_list);
+    void replace_names(py::odict replacements);
```
oleksiyskononenko: So to replace particular names the only option is to use `py::odict`?

c/datatable_cbind.cc
```diff
@@ -18,41 +18,46 @@
...
+    newnames.insert(newnames.end(), namesi.begin(), namesi.end());
+  }
+  xassert(j == t_ncols);
 
-    // Done.
-    ncols = t_ncols;
-    return this;
+  // Done.
+  ncols = t_ncols;
+  set_names(newnames);
```
oleksiyskononenko: `set_names()` will check for possible duplicates, right?

c/extras/aggregator.cc
```diff
@@ -127,7 +126,7 @@ void Aggregator::aggregate_exemplars(DataTable* dt_exemplars,
   cols_counts[0] = Column::new_data_column(SType::INT32,
                                            static_cast<int64_t>(gb_members.ngroups()));
   cols_counts[1] = nullptr;
-  dt_counts = new DataTable(cols_counts);
+  dt_counts = new DataTable(cols_counts, {"counts"});
   auto d_counts = static_cast<int32_t*>(dt_counts->columns[0]->data_w());
   std::memset(d_counts, 0, static_cast<size_t>(gb_members.ngroups()) * sizeof(int32_t));
```
oleksiyskononenko: Will probably need to change this part as already did some refactoring in the `aggregator-dai` branch. Thanks!

c/datatable.h
st-pasha: Yes, because we have no other use cases for now. If in the future a use case arises to replace, say, from an `std::unordered_map`, we can add that function then.

c/datatable_cbind.cc
st-pasha: Yes

c/extras/aggregator.cc
st-pasha: The old `new DataTable(cols_counts);` will no longer compile -- this constructor became private.

oleksiyskononenko: Sure, this is even better. Don't need to remember setting up the names. Will rebase my branch on top of `master` shortly.

PR #1275, opened 2018-09-06
Frame constructor from pandas/numpy is now in C++

### LINKED ISSUE #1273: Create Frame from pandas/numpy sources directly in C++
This will allow us to remove helper methods `_fill_from_source()` etc from the python Frame class, and the awkward "fall back to python" code from `core.Frame` class.

st-pasha:
With this change, the Frame construction can now happen entirely in C++, which allows us to remove lots of C-Python code which now became obsolete.

Closes #1273


c/frame/py_frame_init.cc
```diff
@@ -118,10 +120,17 @@ class FrameInitializationManager {
...
       }
+      if (src.is_pandas_frame() || src.is_pandas_series()) {
+        return init_from_pandas();
+      }
+      if (src.is_numpy_array()) {
+        return init_from_numpy();
+      }
       if (src.is_ellipsis() &&
                !defined_names && !defined_stypes && !defined_stype) {
         return init_mystery_frame();
```
oleksiyskononenko: What is the ellipsis and the mystery frame? :)

st-pasha: `dt.Frame(...)`  🤫

PR #1279, opened 2018-09-07
Fix crash when saving a Frame to CSV

### LINKED ISSUE #1278: .to_csv() crashes for a file with many boolean columns
```
>>> import datatable as dt
>>> f0 = dt.Frame([[True, False] * 10] * 1000)
>>> a = f0.to_csv()
Python(48199,0x70000206a000) malloc: *** error for object 0x102809800: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

st-pasha:
The crash came because of mis-estimating the line length (forgot to account for the separators).

Closes #1278

### DIFF
--- a/c/csv/writer.cc
+++ b/c/csv/writer.cc
@@ -311,6 +311,7 @@ void CsvWriter::write()
   create_column_writers(ncols);
   size_t nstrcols32 = strcolumns32.size();
   size_t nstrcols64 = strcolumns64.size();
+  if (nthreads > nchunks) nthreads = nchunks;
 
   OmpExceptionManager oem;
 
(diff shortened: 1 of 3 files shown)


c/csv/writer.cc
```diff
@@ -464,6 +464,7 @@ size_t CsvWriter::estimate_output_size()
   size_t ncols = static_cast<size_t>(dt->ncols);
   size_t total_string_size = 0;
   size_t total_columns_size = 0;
+  fixed_size_per_row = ncols;  // 1 byte per separator
```
oleksiyskononenko: Do we also need to add 1 byte for `\n`?

st-pasha: There are `ncols-1` separators + 1 newline, `ncols` bytes in total.

PR #1301, opened 2018-09-18
[WIP] Latest release links for S3 

### LINKED ISSUE #952: Add download links for datatable

abal5:
Renaming the builds artifacts and uploading to a separate folder labeled as lastest to generate the latest links for download and install.

Closes #952


jangorecki:
so do we have latest releases published anywhere? so we can just `pip install --upgrade` instead of rebuilding from source?

jangorecki:
could we update readme and add instructions for installation of development version then?

st-pasha:
@abal5 This needs to be rebased on top of latest master, and unrelated changes (the "multi sort" commit) removed. I guess the best strategy would be to create a new PR, cherry-picking the changes from here.

PR #1304, opened 2018-09-20
Split string and nhot-encode

### LINKED ISSUE #1298: Support string columns with set/list of strings and convert internally to one hot encoding
Feedback from Adobe, evaluating DAI against DR.

DR has capability to detect a string column with values of the format ["VALUE1", "VALUE2",....,"VALUEN"] and convert that into a onehot encoding with column for each such value . The list of values is a SET (so unordered). They expressed such a feature would be helpful for them to minimize changes to their existing pipeline.

Linked issue - h2oai/h2oai#4753

Specifications

- Ideal place to have this functional logic is where the column type is being determined numeric vs string vs date etc.
- If the column is determined to the of string type then validate if it of this special category (i.e. a list of strings)
- Need to support both quoted and unquoted values in the list, for example 
```
["Bus"|"Car"|"Train"]
[Bus|Car|Train]
```
- Delimiter used to separate values in the list should be different than the delimiter used to separate columns in the dataset.
- To prevent a system crash due to a large number of columns generated to be cbinded to the original dataset post onehot encoding, there has to be a top limit to the number of possible unique values that are obtained by combining all values in that column in a single set. For example, in the column below the max possible values is 3 for `Bus`, `Car` and `Train` and there will be 3  onehot encoded column cbinded to this dataset.
```
| Column                   |
|--------------------------|
|["Bus"|"Car"]             |
|["Bus"|"Car"]             |
|["Train"]                   |
```
- If the system determines that this configured limit (in config.toml) is exceeded, then data processing should stop with a clear message indicating about the problem and the potential workaround of bumping up the config parameter value in `config.toml` and retry.

st-pasha:
This PR adds a new function `dt.split_into_nhot(frame, sep)`, which acts upon an `n x 1` Frame (of str32 or str64 type) and produces an `n x k` Frame with boolean columns. Each value in the input column is split according to the provided separator, the whitespace is trimmed, and the resulting pieces are converted into individual columns in the output frame.

For example, frame
```
| cat,dog       |
| mouse         |
| cat,mouse     |
| dog,rooster   |
| mouse,dog,cat |
```
gets converted into
```
| cat  dog  mouse rooster |
|   1    1      0       0 |
|   0    0      1       0 |
|   1    0      1       0 |
|   0    1      0       1 |
|   1    1      1       0 |
```

Closes #1298

### DIFF
--- a/c/str/split_into_nhot.cc
+++ b/c/str/split_into_nhot.cc
@@ -1,0 +1,157 @@
+//------------------------------------------------------------------------------
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// © H2O.ai 2018
+//------------------------------------------------------------------------------
+#include <unordered_map>
+#include <vector>
+#include "datatable.h"
+#include "utils/exceptions.h"
+#include "utils/omp.h"
+#include "utils/shared_mutex.h"
+
+namespace dt {
+
+/**
+ * Split string into tokens
+ */
+static void tokenize_string(
+    std::vector<std::string>& tokens,
+    const char* strstart,
+    const char* strend,
+    char sep
+) {
+  tokens.clear();
+  const char* ch = strstart;
+  while (ch < strend) {
+    if (*ch == ' ' || *ch == '\t' || *ch == '\n' || *ch == sep) {
+      ch++;
+    }
+    else {
+      const char* ch0 = ch;
+      // Handle the case of quoted token
+      if (*ch == '\'' || *ch == '"') {
+        char quote = *ch++;
+        while (ch < strend && *ch != quote) {
+          ch += 1 + (*ch == '\\');
+        }
+        if (ch < strend) {
+          tokens.emplace_back(ch0 + 1, ch);
+          ch++;  // move over the closing quote
+          continue;
+        }
+        // fall-through
+      }
+      // Regular non-quoted token, parse until the next sep
+      while (ch < strend && *ch != sep) {
+        ch++;
+      }
+      const char* ch1 = ch;
+      while (ch1 > ch0) {  // also omit trailing spaces
+        char c = *(ch1 - 1);
+        if (!(c == ' ' || c == '\t' || c == '\n')) break;
+        ch1--;
+      }
+      tokens.emplace_back(ch0, ch1);
+      ch++;  // move over the sep
+    }
+  }
+}
+
+
+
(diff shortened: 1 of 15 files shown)


c/str/split_into_nhot.cc
```diff
@@ -0,0 +1,159 @@
...
+      while (ch1 > ch0) {  // also omit trailing spaces
+        char c = *(ch1 - 1);
+        if (!(c == ' ' || c == '\t' || c == '\n')) break;
+        ch1--;
+      }
+      tokens.emplace_back(ch0, ch1);
+      ch++;  // move over the sep
+    }
+  }
+}
```
oleksiyskononenko: Is there any reason not to use `strtok()` function from `<cstring>`? Performance?

st-pasha: It doesn't have the ability to handle quotes, it doesn't strip whitespace, and it modifies the string that is passed to it, so I'm not even sure whether it's safe to use with `std::string`.

oleksiyskononenko: I see, yes, trimming and quotes should be done separately. I guess one can't use it with `std::string` directly, so a duplicate created with something like `strdup()` may be needed.

c/str/split_into_nhot.cc
```diff
@@ -0,0 +1,159 @@
...
+  std::vector<int8_t*> outdata;
+  std::vector<std::string> outnames;
+  dt::shared_mutex shmutex;
+
+  OmpExceptionManager oem;
+  #pragma omp parallel
+  {
+    try {
+      size_t ith = static_cast<size_t>(omp_get_thread_num());
+      size_t nth = static_cast<size_t>(omp_get_num_threads());
```
oleksiyskononenko: What about setting `nth` as `config::nthreads` here? Then number of threads used by this function will be controllable, just like for all the others.

c/str/split_into_nhot.cc
```diff
@@ -0,0 +1,159 @@
...
+
+  size_t nrows = static_cast<size_t>(col->nrows);
+  std::unordered_map<std::string, size_t> colsmap;
+  std::vector<Column*> outcols;
+  std::vector<int8_t*> outdata;
+  std::vector<std::string> outnames;
+  dt::shared_mutex shmutex;
+
+  OmpExceptionManager oem;
+  #pragma omp parallel
```
oleksiyskononenko: What about doing `#pragma omp parallel num_threads(nth0)` here? `nth0` can be `config::nthreads` it its simplest form, or 
```
int nth0 = std::min(config::nthreads,
                        static_cast<int>(nrows / min_nrows_per_thread));
```
like we do in other places.

PR #1308, opened 2018-09-24
Single-element getter

### LINKED ISSUE #1306: Reduce overhead of a single-element access
Although vector-like operations are the preferred way of accessing a datatable Frame, occasionally there will be use cases where the elements need to be accessed individually. (E.g. random access). However, when such cases arise, there may as well be millions+ of such calls, and therefore it is important to keep the overhead low.

This issue calls for the following changes:
* Change the return value of `DT[i, j]` call in the case when `i` is an integer and `j` is either an integer or a string (explicit single-cell access). Currently, this returns a 1x1 frame (which causes lots of overhead for the frame construction); with this proposal, the return type will be a python scalar.
* The `__getitem__` / `__setitem__` methods should be implemented in C++. The special case of integer `i` and `j` should be checked first, and then executed via fast path.

Current benchmark of element extraction operation is the following:
```python
import datatable as dt
import time

dt_frame = dt.Frame([range(100)] * 30)
py_frame = dt_frame.topython()
pd_frame = dt_frame.topandas()

def test_dt(n):
    t0 = time.time()
    for i in range(n):
        _ = dt_frame[35, 10]
    t1 = time.time()
    print("  [dt] Time/iter = %5dns" % ((t1 - t0)/n * 10**9, ))

def test_py(n):
    t0 = time.time()
    for i in range(n):
        _ = py_frame[10][35]
    t1 = time.time()
    print("  [py] Time/iter = %5dns" % ((t1 - t0)/n * 10**9, ))

def test_pd(n):
    t0 = time.time()
    for i in range(n):
        _ = pd_frame.iloc[35, 10]
    t1 = time.time()
    print("  [pd] Time/iter = %5dns" % ((t1 - t0)/n * 10**9, ))


for n in [100, 1000, 10000, 10**5]:
    print("n = %d:" % n)
    test_dt(n)
    test_pd(n)
    test_py(n)
```
Output:
```
n = 100:
  [dt] Time/iter = 37298ns
  [pd] Time/iter = 11618ns
  [py] Time/iter =    88ns
n = 1000:
  [dt] Time/iter = 25192ns
  [pd] Time/iter =  8647ns
  [py] Time/iter =    78ns
n = 10000:
  [dt] Time/iter = 22851ns
  [pd] Time/iter =  8257ns
  [py] Time/iter =    71ns
n = 100000:
  [dt] Time/iter = 23091ns
  [pd] Time/iter =  8102ns
  [py] Time/iter =    70ns
```

Although it is unlikely that we can match the speed of "py" implementation (no conversions involved: the `PyObject*` stored inside the array can be retrieved as-is), but at least we should aim to match or surpass the performance of pandas' `iloc`.

st-pasha:
- `DT[i, j]` now returns a python scalar value if `i` is integer, and `j` is integer/string. This is referred to as "explicit element selection". In the unlikely scenario when a single element needs to be returned as a frame, one can always write `DT[i:i+1, j]` or `DT[[i], j]`.
- The performance of explicit element selection improved by a factor of 200x.

Closes #1306

### DIFF
--- a/c/datatablemodule.cc
+++ b/c/datatablemodule.cc
@@ -31,6 +31,7 @@ extern void init_jay();
 
 namespace py {
   PyObject* fread_fn = nullptr;
+  PyObject* fallback_makedatatable = nullptr;
 }
 
 
@@ -101,6 +102,7 @@ PyObject* register_function(PyObject*, PyObject *args) {
   else if (n == 6) replace_dtWarning(fnref);
   else if (n == 7) py::Frame_Type = fnref;
   else if (n == 8) py::fread_fn = fnref;
+  else if (n == 9) py::fallback_makedatatable = fnref;
   else {
     throw ValueError() << "Incorrect function index: " << n;
   }
--- a/c/frame/getset.cc
+++ b/c/frame/getset.cc
@@ -14,13 +14,29 @@
 // limitations under the License.
 //------------------------------------------------------------------------------
 #include "frame/py_frame.h"
+#include "python/dict.h"
 #include "python/tuple.h"
+#include "python/string.h"
 
 namespace py {
 
+// Sentinel value for __getitem__() mode
+static obj GETITEM(reinterpret_cast<PyObject*>(-1));
 
-oobj Frame::m__getitem__(const obj item) const {
 
+
+
+oobj Frame::m__getitem__(obj item) {
+  return _getset(item, GETITEM);
+}
+
+void Frame::m__setitem__(obj item, obj value) {
+  _getset(item, value);
+}
+
+
+oobj Frame::_getset(obj item, obj value) {
+  odict kwargs;
   otuple args(5);
   args.set(0, py::obj(this));
   if (item.is_tuple()) {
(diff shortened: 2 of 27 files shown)


c/datatablemodule.cc
```diff
@@ -101,6 +102,7 @@ PyObject* register_function(PyObject*, PyObject *args) {
   else if (n == 6) replace_dtWarning(fnref);
   else if (n == 7) py::Frame_Type = fnref;
   else if (n == 8) py::fread_fn = fnref;
+  else if (n == 9) py::fallback_makedatatable = fnref;
```
oleksiyskononenko: Any reason not to use `switch` here?

c/frame/getset.cc
```diff
@@ -0,0 +1,129 @@
...
+          throw ValueError() << "Row `" << irow << "` is invalid for a frame "
+              "with " << dt->nrows << " row" << (dt->nrows == 1? "" : "s");
+        }
+        int64_t icol;
+        if (a1int) {
+          icol = arg1.to_int64_strict();
+          if (icol < 0) icol += dt->ncols;
+          if (icol < 0 || icol >= dt->ncols) {
+            if (icol < 0) icol -= dt->ncols;
+            throw ValueError() << "Column index `" << icol << "` is invalud "
```
oleksiyskononenko: Should probably be spelled "invalid".

c/datatablemodule.cc
st-pasha: it grew over time :) But really, this code is executed only once, during import, so not a big deal

c/frame/getset.cc
st-pasha: Yep. Looks like I forgot to add test cases for the exceptions thrown in this method.

PR #1319, opened 2018-10-01
Frame.replace() method

### LINKED ISSUE #1302: Implement pandas-like .replace
https://github.com/h2oai/datatable/issues/1127#issuecomment-400778330

Otherwise, when an inf or other item exists in dt frame, we have to go to pandas, do the .replace, and then push the column back into the dt frame.  This ends-up adding an enormous amount of code (about 1000 lines of code in DAI) that would be eliminated.

st-pasha:
Implement `.replace()` method for a Frame. This currently works with int or float columns only; handling of bool and str columns will be done later (this PR is already too big).

Additional changes:
* `Args` object now knows both the function name and its doc-string from the moment of construction (this allows to declare a method's signature much easier);
* Error messages now say `"Frame() constructor"` instead of `"Frame.__init__()"`;
* Start working on "progress" functionality -- a more unified way to run code in parallel. The reason to have this unified is to be able to eventually add the progress-bar functionality for all long-running tasks;

Closes #1302


c/frame/py_frame.cc
```diff
@@ -28,8 +28,18 @@ PyObject* Frame_Type = nullptr;
...
+                                  "__init__", nullptr);
+NoArgs Frame::Type::args_copy("copy",
+"copy(self)\n"
+"--\n\n"
+"Make a copy of this Frame.\n"
+"\n"
+"This method creates a shallow copy of the current Frame: only references\n"
+"are copied, not the data itself. However, due to copy-on-write semantics\n"
+"any changes made to one of the Frames will not propagate to the other.\n"
+"Thus, for all inents and purposes the copied Frame will behave as if\n"
```
oleksiyskononenko: Should it be "intents"?

c/frame/replace.cc
```diff
@@ -0,0 +1,569 @@
...
+R"(replace(self, replace_what, replace_with)
+--
+
+Replace given value(s) `replace_what` with `replace_with` in the entire Frame.
+
+For each replace value, this method operates only on columns of types
+appropriate for that value. For example, if `replace_what` is a list
+`[-1, math.inf, None, "??"]`, then the value `-1` will be replaced in integer
+columns only, `math.inf` only in real columns, `None` in columns of all types,
+and finally `"??"` only in string columns.
```
oleksiyskononenko: Is it because you don't want to change the column type during replacement?

c/frame/py_frame.cc
st-pasha: well spotted

c/frame/replace.cc
st-pasha: If the user wanted to change a column's type, he can do it explicitly beforehand. Otherwise, the semantics is weird: suppose you're changing integer `99` into string `"--"`. What happens if the column doesn't have any 99s? Should it remain integer, or get converted into string anyways? Having an unpredictable result type makes it hard to reason about the code.

oleksiyskononenko: I agree, also don't like an idea to change the column type during replacement as this seems to be completely separate functionality.

PR #1322, opened 2018-10-01
Implement replace function for boolean columns

### LINKED ISSUE #1320: Implement replace function for bool columns
This is a follow-up for #1319

st-pasha:
* Allow `replace()` to work with boolean columns;
* Improve test coverage

Closes #1320

### DIFF
--- a/tests/munging/test_replace.py
+++ b/tests/munging/test_replace.py
@@ -22,6 +22,63 @@ from tests import list_equals
 
 
 #-------------------------------------------------------------------------------
+# Frame.replace() signature
+#-------------------------------------------------------------------------------
+
+def test_replace_scalar_scalar():
+    df = dt.Frame([1, 2, 3])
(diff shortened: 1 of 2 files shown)


tests/munging/test_replace.py
```diff
@@ -21,6 +21,63 @@
...
+
+
+
+#-------------------------------------------------------------------------------
+# Replacing in bool columns
+#-------------------------------------------------------------------------------
+
+def test_replace_bool_simple():
+    df = dt.Frame([[True, False, None], [True] * 3, [False] * 3])
+    df.replace({True: False, False: True})
```
oleksiyskononenko: So it does replacement once for each value, i.e. there is no way `True` will be replaced with `False` and this `False` will be again replaced with `True`?

st-pasha: that's correct

PR #1328, opened 2018-10-06
Improve handling of Excel files

st-pasha:
* sheet name can now be used as a path component in the file name, causing only that particular sheet to be parsed;
* further, a cell range can be specified as a path component after the sheet name, forcing fread to consider only the provided cell range. For example, the following request is now valid:
   ``` 
   dt.fread("../h2oai/tests/data/diabetes_tiny_two_sheets.xlsx/Sheet1/A10:BC30")   
   ```
* fread can now handle the situation when a spreadsheet has multiple separate tables in the same sheet. They will now be detected automatically and returned to the user as separate Frame objects (the name of each frame will contain the sheet name and cell range from where the data was extracted).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - ability to convert object columns into strings.
 - implemented `Frame.replace()` function.
 - function `abs()` to find the absolute value of elements in the frame.
+- improved handling of Excel files by fread:
+  * sheet name can now be used as a path component in the file name,
+    causing only that particular sheet to be parsed;
+  * further, a cell range can be specified as a path component after the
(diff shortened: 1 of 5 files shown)


st-pasha:
* sheet name can now be used as a path component in the file name, causing only that particular sheet to be parsed;
* further, a cell range can be specified as a path component after the sheet name, forcing fread to consider only the provided cell range. For example, the following request is now valid:
   ``` 
   dt.fread("../h2oai/tests/data/diabetes_tiny_two_sheets.xlsx/Sheet1/A10:BC30")   
   ```
* fread can now handle the situation when a spreadsheet has multiple separate tables in the same sheet. They will now be detected automatically and returned to the user as separate Frame objects (the name of each frame will contain the sheet name and cell range from where the data was extracted).

PR #1338, opened 2018-10-11
Do proper N/A handling in 1D and 2D

### LINKED ISSUE #1337: Proper handling of N/A's in 1D and 2D
Right now N/A's are handled differently for 1 and 2D, and ND aggregations. In 1 and 2D they're replaced with zeros, and in ND they're ignored for the distance calculations. We should also ignore them for 1 and 2D, i.e. at least they should not end up in any bin with the actual numbers.

oleksiyskononenko:
For all 1D and 2D aggregators, if N/A is present in one of the columns, the row goes to a special N/A bin, that at the end becomes zero bin. An exemplar for this bin is just the first row that contains even one N/A value.

Tests are updated according to the above changes.

Closes #1337

### DIFF
--- a/c/extras/aggregator.cc
+++ b/c/extras/aggregator.cc
@@ -245,7 +245,7 @@ void Aggregator::group_2d(const DataTablePtr& dt, DataTablePtr& dt_members) {
 */
 void Aggregator::group_1d_continuous(const DataTablePtr& dt,
                                      DataTablePtr& dt_members) {
-  const auto c0 = static_cast<RealColumn<double>*>(dt->columns[0]);
+  auto c0 = static_cast<const RealColumn<double>*>(dt->columns[0]);
   const double* d_c0 = c0->elements_r();
   auto d_members = static_cast<int32_t*>(dt_members->columns[0]->data_w());
 
@@ -268,8 +268,8 @@ void Aggregator::group_1d_continuous(const DataTablePtr& dt,
 */
 void Aggregator::group_2d_continuous(const DataTablePtr& dt,
                                      DataTablePtr& dt_members) {
-  const auto c0 = static_cast<RealColumn<double>*>(dt->columns[0]);
-  const auto c1 = static_cast<RealColumn<double>*>(dt->columns[1]);
+  auto c0 = static_cast<const RealColumn<double>*>(dt->columns[0]);
+  auto c1 = static_cast<const RealColumn<double>*>(dt->columns[1]);
   const double* d_c0 = c0->elements_r();
   const double* d_c1 = c1->elements_r();
   auto d_members = static_cast<int32_t*>(dt_members->columns[0]->data_w());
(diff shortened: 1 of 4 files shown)


c/extras/aggregator.cc
```diff
@@ -273,7 +276,7 @@ void Aggregator::group_2d_continuous(DataTablePtr& dt,
   auto c1 = static_cast<RealColumn<double>*>(dt->columns[1]);
   double* d_c0 = c0->elements_w();
   double* d_c1 = c1->elements_w();
```
st-pasha: Hmm, if this method is not supposed to modify `dt`, then it should be declared `const`. Then `c0`, `c1`, `d_c0`, `d_c1` will be const too, and you wouldn't have to ask for a writable (`_w`) version of the data. Generally, you should request writable data only when you intend to write into it -- otherwise, you may be creating unnecessary copies.

c/extras/aggregator.cc
```diff
@@ -282,10 +285,12 @@ void Aggregator::group_2d_continuous(DataTablePtr& dt,
 
   #pragma omp parallel for schedule(static)
   for (int64_t i = 0; i < dt->nrows; ++i) {
-    double v0 = ISNA<double>(d_c0[i])? 0 : d_c0[i];
-    double v1 = ISNA<double>(d_c1[i])? 0 : d_c1[i];
-    d_groups[i] = static_cast<int32_t>(normy_factor * v1 + normy_shift) * nx_bins +
-                  static_cast<int32_t>(normx_factor * v0 + normx_shift);
+    if (ISNA<double>(d_c0[i]) || ISNA<double>(d_c1[i])) {
+      d_members[i] = NA_I4;
```
st-pasha: `GETNA<int32_t>()` is a more C++-ish way. Although more verbose, at least these constants are strictly typed.

c/extras/aggregator.cc
oleksiyskononenko: Oh, that's just something that is left from the old times. Thanks for catching it. I will also review the rest of the code if there is anything left like that.

oleksiyskononenko: Fixed.

c/extras/aggregator.cc
oleksiyskononenko: Fixed.

c/extras/aggregator.cc
```diff
@@ -258,22 +254,25 @@ void Aggregator::group_1d_continuous(DataTablePtr& dt,
...
 */
-void Aggregator::group_2d_continuous(DataTablePtr& dt,
+void Aggregator::group_2d_continuous(const DataTablePtr& dt,
                                      DataTablePtr& dt_members) {
-  auto c0 = static_cast<RealColumn<double>*>(dt->columns[0]);
-  auto c1 = static_cast<RealColumn<double>*>(dt->columns[1]);
-  double* d_c0 = c0->elements_w();
-  double* d_c1 = c1->elements_w();
-  auto d_groups = static_cast<int32_t*>(dt_members->columns[0]->data_w());
+  const auto c0 = static_cast<RealColumn<double>*>(dt->columns[0]);
```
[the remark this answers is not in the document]
oleksiyskononenko: No, I am not getting any warnings, but what you says makes sense. Let me fix it.

PR #1339, opened 2018-10-11
Better rendering of a Frame in Jupyter notebook

### LINKED ISSUE #68: Implement datatable view widget for Jupyter Notebook
For inspiration, see
* https://github.com/6pac/SlickGrid/wiki  
* https://github.com/quantopian/qgrid

st-pasha:
This PR adds method `Frame._repr_html_()`, which is invoked by JupyterLab to generate nice HTML view of the object.

Most styling is inherited from the Jupyter stylesheet, and only a few own styles are added. These styles are scoped by `.datatable` CSS selector: the idea is to render all HTML widgets inside `<div class=datatable />`. The styles are also emitted only once per session.

Generally, the rendered view is similar to how pandas DataFrame looks like; in particular, when there are too many rows/columns the middle ones are omitted.

Closes #68

### DIFF
--- a/c/frame/_repr_html_.cc
+++ b/c/frame/_repr_html_.cc
@@ -1,0 +1,270 @@
+//------------------------------------------------------------------------------
+// Copyright 2018 H2O.ai
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//------------------------------------------------------------------------------
+#include "frame/py_frame.h"
+#include <sstream>
+#include "python/string.h"
+#include "types.h"
+
+
+
+class HtmlWidget {
+  private:
+    std::ostringstream html;
+    DataTable* dt;
+    size_t ncols, cols0, cols1;
+    size_t nrows, rows0, rows1;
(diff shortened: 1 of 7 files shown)


c/frame/_repr_html_.cc
```diff
@@ -0,0 +1,270 @@
...
+      render_all();
+      const std::string htmlstr = html.str();
+      return py::ostring(htmlstr);
+    }
+
+
+  private:
+    void render_all() {
+      render_styles();
+      html << "<div class=datatable>";
```
oleksiyskononenko: I would quote attributes even though this is not required. Several reasons: 
- looks cleaner as you can quickly distinguish where the value starts/ends;
- later you can always do `<div class='datatable something_else'>` without the need to add quotes, so no way to forget to do that.

It's a matter of taste, of course.

c/frame/_repr_html_.cc
```diff
@@ -0,0 +1,270 @@
...
+
+
+    void render_escaped_string(const char* ch, size_t len) {
+      size_t maxi = std::min(len, size_t(50));
+      for (size_t i = 0; i < maxi; ++i) {
+        char c = ch[i];
+        if (c == '&') html << "&amp;";
+        else if (c == '<') html << "&lt;";
+        else if (c == '>') html << "&gt;";
+        else html << c;
```
oleksiyskononenko: I would also replace single quotes `\'` with `&apos;`, and double quotes `\"` with `&quot;`. Btw, what about using `switch`, as `c` is `char`? Should look cleaner.

st-pasha: You're right, those 2 characters are recommended to be escaped too. Also maybe any characters with ASCII value < 0x20 (such as `\n`, `\r`, `\t`)? They aren't dangerous per se, but perhaps the user would want to know they're present.

oleksiyskononenko: Yes, these could also be done just for the user convenience. Quotes are probably more important as they could mess up output significantly.

PR #1350, opened 2018-10-19
Implementation of exponential expression

nkalonia1:
`f[:, dt.exp(dt.f.x)]` is now a valid expression for Frame `f` with column `x`

### DIFF
--- a/tests/test_dt_expr.py
+++ b/tests/test_dt_expr.py
@@ -269,17 +269,17 @@ def test_abs_all_stypes():
 
 def test_exp():
     from datatable import exp
-    import cmath
-    assert exp(0) == cmath.exp(0)
-    assert exp(1) == cmath.exp(1)
-    assert exp(-2.5e12) == cmath.exp(-2.5e12)
+    import math
+    assert exp(0) == math.exp(0)
+    assert exp(1) == math.exp(1)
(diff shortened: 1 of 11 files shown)


tests/test_dt_expr.py
```diff
@@ -263,6 +263,65 @@ def test_abs_all_stypes():
 
 
 
+#-------------------------------------------------------------------------------
+# exp()
+#-------------------------------------------------------------------------------
+
+def test_exp():
+    from datatable import exp
+    import cmath
```
st-pasha:
```suggestion
    import math
```
because `cmath` is for complex math

tests/test_dt_expr.py
```diff
@@ -263,6 +263,65 @@ def test_abs_all_stypes():
...
+    assert exp(0) == cmath.exp(0)
+    assert exp(1) == cmath.exp(1)
+    assert exp(-2.5e12) == cmath.exp(-2.5e12)
+    assert exp(12345678) == float("inf")
```
st-pasha:
```suggestion
    assert exp(12345678) == math.inf
```

st-pasha:
Jenkins error seems to be in the environment, not related to datatable

PR #1388, opened 2018-11-06
Fix writing large columns to file

### LINKED ISSUE #1387: Cannot save big Frame
If a Frame contains any column over 2GB in size, the frame will fail to save. This appears to be caused by the limitations of standard library's `write()`, which is unable to write more than 2GB of data.
```
import datatable as dt
frame0 = dt.Frame(A=[1.0, 2.0, 10.0])
frame0.nrows = 2**28  # 8 bytes per row
frame0.save("test.jay", format="jay")
```
produces
```
  File "/Users/pasha/py36/lib/python3.6/site-packages/typesentry/config.py", line 155, in fdecorated
    ret = f(*args, **kws)
  File "/Users/pasha/github/datatable/datatable/nff.py", line 48, in save
    self.internal.save_jay(dest, self.names, _strategy)
RuntimeError: Cannot write to file: [errno 22] Invalid argument (bytes already written: 8)
```

st-pasha:
When a large column is written to file (> 2 GB), it will no longer fail with errno 22. Instead, we will write the data in chunks of 1GB until all data is written.

Closes #1387

### DIFF
--- a/c/writebuf.cc
+++ b/c/writebuf.cc
@@ -76,25 +76,43 @@ FileWritableBuffer::~FileWritableBuffer() {
 }
 
 
-size_t FileWritableBuffer::prep_write(size_t size, const void* src) {
+size_t FileWritableBuffer::prep_write(size_t src_size, const void* src)
+{
+  constexpr size_t CHUNK_SIZE = 1 << 30;
   size_t pos = bytes_written;
+  if (!src_size) return pos;
 
-  // See https://linux.die.net/man/2/write
-  ssize_t r = ::write(file->descriptor(), src, size);
-
-  if (r == -1) {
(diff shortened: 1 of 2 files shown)


c/writebuf.cc
```diff
@@ -76,25 +76,43 @@ FileWritableBuffer::~FileWritableBuffer() {
 }
 
 
-size_t FileWritableBuffer::prep_write(size_t size, const void* src) {
+size_t FileWritableBuffer::prep_write(size_t src_size, const void* src)
```
oleksiyskononenko: Any way to parallelize this function? Or it is fast enough?

st-pasha: Yes, it's implemented in class `MmapWritableBuffer`. However, it does not always work inside docker instances.

Issue #1369, opened 2018-10-24
Make datatable compile with MSVC


st-pasha:
This is a prerequisite for #1114 

Currently, there seems to be multiple compilation issues, including MSVC [not working properly with macros](https://blogs.msdn.microsoft.com/vcblog/2018/07/06/msvc-preprocessor-progress-towards-conformance/), and also lack of `sys/mman.h` library on windows (for memory-mapping)

thirdwing:
In case of `sys/mman.h`, there is a Windows version used in feather: https://github.com/wesm/feather/blob/master/cpp/src/feather/mman.h

dmillar:
What is the status of this issue? Is this being actively worked on? If not, I'd like to take a crack at getting datatable to compile on Windows. MSVC if possible, but also open to mingw and clang/llvm Windows implementations. VS2019 seems to have introduced better adherence to the standard, and might make this task a bit easier. Though, at first thought I'd probably want to use CMake to build. Thoughts on approach, known issues?

PR #1390, opened 2018-11-07
 Implement FTRL-Proximal online learning algorithm

### LINKED ISSUE #1389: Implement FTRL-Proximal online learning algorithm
Implement FTRL-Proximal online learning algorithm as outlined in "[Ad Click Prediction: a View from the Trenches](https://www.eecs.tufts.edu/~dsculley/papers/ad-click-prediction.pdf)" paper.

oleksiyskononenko:
Initial implementation of FTRL algo.

Closes #1389.


c/datatablemodule.cc
```diff
@@ -199,6 +200,7 @@ void DatatableModule::init_methods() {
   add(METHOD0(is_debug_mode));
   add(METHOD0(has_omp_support));
   add(METHODv(aggregate));
+  add(METHODv(ftrl));
```
st-pasha: Note that macros such as `METHOD0`, `METHODv` etc will soon be removed due to issue #1369. The newer mechanism for exposing API into python can be seen for example in `py_str.cc`.

c/extras/ftrl.cc
```diff
@@ -0,0 +1,527 @@
...
+  d(d_in),
+  n_epochs(nepochs_in),
+  hash_type(hash_type_in),
+  seed(seed_in),
+  inter(inter_in)
+{
+  n = DoublePtr(new double[d]);
+  w = DoublePtr(new double[d]);
+  std::memset(n.get(), 0, d * sizeof(double));
+  std::memset(w.get(), 0, d * sizeof(double));
```
st-pasha: instead of memset, you could have called `new double[d]()`, note the parentheses.
See https://stackoverflow.com/questions/620137

c/extras/ftrl.cc
```diff
@@ -0,0 +1,527 @@
...
+        double p = predict(x, n_features + n_inter_features);
+        double loss = logloss(p, target);
+
+        #pragma omp atomic update
+        total_loss += loss;
+
+        if ((j+1) % REPORT_FREQUENCY == 0) {
+          printf("Training epoch: %zu\tRow: %zu\tPrediction: %f\t"
+                 "Current loss: %f\tAverage loss: %f\n",
+                 i, j+1, p, loss, total_loss / (j+1));
```
st-pasha: I presume this is temporary debug code?

c/extras/ftrl.cc
```diff
@@ -0,0 +1,527 @@
...
+*/
+inline uint64_t Ftrl::hash_double(double x) {
+  uint64_t* h = reinterpret_cast<uint64_t*>(&x);
+  return *h;
+}
+
+
+/*
+*  Murmur2 hash function by Austin Appleby.
+*  More details at https://sites.google.com/site/murmurhash/
```
st-pasha: If some code is taken from external sources, the best practice is to:
1) Put that code in a separate file
2) Clearly note where it was taken from, and what was the license
3) Make sure the conditions of the license are satisfied.
For example, the link above says the code is licensed MIT. However, MIT license requires that "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

datatable/extras/ftrl.py
```diff
@@ -0,0 +1,72 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#-------------------------------------------------------------------------------
+import datatable as dt
+from datatable.lib import core
+
+
+def ftrl(df_train, df_test, a=0.01, b=1.0, l1=0.0, l2=1.0,
```
st-pasha: This function is probably not needed at all.
You can simply say `ftrl = core.ftrl`.
This is probably true for the `aggregate` function above too.

c/extras/ftrl.cc
oleksiyskononenko: Yes, sure. This part will probably have to be changed anyway as we may want to load these coefficients from a pre-trained model.

c/extras/ftrl.cc
oleksiyskononenko: Yes, we will probably need some kind of progress function here. Don't know for the moment.

c/extras/ftrl.cc
oleksiyskononenko: I'm not sure about the MIT license. This is what you get when go to the site where this code moved at some point

//-----------------------------------------------------------------------------
// MurmurHash3 was written by Austin Appleby, and is placed in the public
// domain. The author hereby disclaims copyright to this source code.

https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp
https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp

What do you think we should do?

oleksiyskononenko: Actually I just noticed that the link says "Update March 1, 2011 - The 32-bit variant of MurmurHash3 is finalized, and this page is deprecated", so it's probably better not to use it and replace with https://github.com/aappleby/smhasher On the latter It says "This is the home for the MurmurHash family of hash functions along with the SMHasher test suite used to verify them. SMHasher is released under the MIT license. All MurmurHash versions are public domain software, and the author disclaims all copyright to their code." Do you think it means `MurmurHash` is licensed under MIT or not?

datatable/extras/ftrl.py
oleksiyskononenko: I would prefer to keep the both for the moment, because 
- while debugging it is easier for me to change default parameters without re-compiling the whole thing;
- one can do `help(ftrl)` or `help(aggregate)` to get some help, not sure if this could be done with `core`.

c/extras/ftrl.cc
st-pasha: It's not.
In the copyright law, the owner (author) of the code has an exclusive right to decide what other people can or cannot do to their code. That decision is embodied in the form of a LICENSE. For example, MIT license says "do pretty much anything, but don't remove this notice".

However, once the author puts his code into the public domain, he is no longer the owner and cannot grant licenses. The code can be used by anyone, without any pre-conditions. 

Thus, the best approach would be to: (1) put it into a separate file, (2) add a notice at the top of the file that this code is written by Austin Appleby and placed into public domain, (3) add a URL where it was borrowed from.
Do not add our standard MIT license boilerplate, because that would require us to claim the copyright in the work, which is technically possible but not very fair.

datatable/extras/ftrl.py
st-pasha: sure, we can always remove them later

PR #1394, opened 2018-11-08
Added support for multi-column keys

### LINKED ISSUE #1376: Support multi-column keys
This is a pre-requisite for #1353

```
>>> import datatable as dt
>>> F = dt.Frame(A=[3, 1, 9, 0], B=list('cbba'), D = range(4))
>>> F.key = ("A", "B")
  File "/Users/pasha/github/datatable2/datatable/frame.py", line 68, in key
    self._dt.nkeys = nk
NotImplementedError: More than 1 key column is not supported yet
```

st-pasha:
A Frame can now have a multi-column key, for example:
```
frame.key = ("A", "B")
```

Closes #1376

### DIFF
--- a/c/frame/key.cc
+++ b/c/frame/key.cc
@@ -132,4 +132,3 @@ void DataTable::set_key(std::vector<size_t>& col_indices) {
 
   nkeys = K;
 }
-
--- a/c/frame/key.cc
+++ b/c/frame/key.cc
@@ -19,15 +19,20 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
+#include <numeric>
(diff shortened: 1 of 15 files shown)


c/frame/key.cc
```diff
@@ -0,0 +1,138 @@
...
+        throw ValueError() << "Column `" << names[col_indices[i]] << "` is "
+          "specified multiple times within the key";
+      }
+    }
+  }
+
+  // Sort the table by the keys
+  Groupby gb;
+  RowIndex ri = sortby(col_indices, &gb);
+  xassert(ri.length() == nrows);
```
oleksiyskononenko: What are the situations when `ri.length() == nrows` is false?

st-pasha: Well, this is an assertion, which means there shouldn't be any.

oleksiyskononenko: Right, but I was like this is already checked by the `sortby` function itself?

Issue #1391, opened 2018-11-07
Optimize performance of hash functions in FTRL


oleksiyskononenko:
More benchmarks are needed to optimize performance of FTRL implementation. In particular, issues related to memory alignment must be explored on different systems.

PR #1417, opened 2018-11-16
Convert FTRL code to a core datatable class

### LINKED ISSUE #1392: Add an ability to save/open an FTRL model
Also need to make sure hashes are reproducible between the systems, so that the trained model will result in the same predictions when re-opened somewhere else.

### LINKED ISSUE #1416: Convert FTRL algo into a datatable core class
This should be more convenient for future uses, and also helps in completing #1392.

oleksiyskononenko:
FTRL code has been converted into a core datatable class:
- added `fit` and `predict` methods for training and making predictions;
- added `reset` method to reset the model;
- added getters and setters for all the model parameters;
- the `model` itself can now be returned and loaded as a separate frame;
- some other refactoring and modifications.

The algo itself was also slightly optimized, however, it still gets the same score on Kaggle. 
So we're ok from the accuracy point of view.

Closes #1416 
Closes #1392


c/extras/ftrl.h
```diff
@@ -25,39 +25,73 @@ typedef std::unique_ptr<double[]> DoublePtr;
...
     double b;
     double l1;
     double l2;
-    size_t n_features;
-    size_t n_inter_features;
     uint64_t d;
-    size_t n_epochs;
+    bool inter;
+    size_t : 24;
     unsigned int hash_type;
```
st-pasha: `hash_type` should really be an enum with 3 values; and exposed into python as strings "std", "murmur2", "murmur3".

c/extras/ftrl.h
```diff
@@ -25,39 +25,73 @@ typedef std::unique_ptr<double[]> DoublePtr;
...
+
+    // Calculated during the learning process.
+    size_t n_features;
+    size_t n_inter_features;
+    double* z;
+    double* n;
     DoublePtr w;
-    bool inter;
-    uint64_t : 56;
+    dtptr dt_model;
```
st-pasha: I prefer class members to be listed at the beginning of the class -- this makes the class's structure more obvious.

c/extras/ftrl.h
```diff
@@ -25,39 +25,73 @@ typedef std::unique_ptr<double[]> DoublePtr;
...
+    // Input to the model.
     double a;
     double b;
     double l1;
     double l2;
-    size_t n_features;
-    size_t n_inter_features;
     uint64_t d;
-    size_t n_epochs;
+    bool inter;
```
st-pasha: if you move this `bool` after the variable `seed`, you'll have only 1 gap in your data structure.

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#include "extras/py_ftrl.h"
+#include "frame/py_frame.h"
+#include "python/float.h"
+#include "python/int.h"
+
+namespace py {
+
+PKArgs Ftrl::Type::args___init__(9, 0, 0, false, false,
```
st-pasha: It's better to have all (or most) of the arguments keyword-only

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+    A new Frame of shape `(d, 2)` containing `z` and `n` model coefficients.
+)");
+
+
+void Ftrl::fit(const PKArgs& args) {
+  if (!args[0].is_frame()) {
+    throw TypeError() << "argument must be a frame, not "
+        << args[0].typeobj();
+  }
+  DataTable* dt_train = args[0].to_frame();
```
st-pasha: this call will already perform the check `.is_frame()` and throw an exception with similar message as the above

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+    "`1` - Murmur2;\n"
+    "`2` - Murmur3.\n");
+  gs.add<&Ftrl::get_seed, &Ftrl::set_seed>("seed", "Seed to be used for Murmur hash functions.\n");
+
+  mm.add<&Ftrl::fit, args_fit>();
+  mm.add<&Ftrl::predict, args_predict>();
+}
+
+
+PKArgs Ftrl::Type::args_fit(1, 0, 0, false, false, {"df_train"}, "fit",
```
st-pasha: the argument name should probably be `frame` instead of `df_train`, same as in the docs

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+Train an FTRL model on a dataset.
+
+Parameters
+----------
+frame: Frame
+    Frame to be trained on, last column is treated as `target`.
+
+Returns
+----------
+    A new Frame of shape `(d, 2)` containing `z` and `n` model coefficients.
```
st-pasha: the function `Ftrl::fit` is declared as returning `void`. So either the signature is incorrect, or the documentation.

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+  if (!args[0].is_frame()) {
+    throw TypeError() << "argument must be a frame, not "
+        << args[0].typeobj();
+  }
+
+  DataTable* dt_test = args[0].to_frame();
+  DataTable* dt_target = fm->predict(dt_test).release();
+  py::Frame* df_target = py::Frame::from_datatable(dt_target);
+
+  return df_target;
```
st-pasha: I suspect this creates a memory leak when `df_target` is converted to `py::oobj`. Use `py::oobj::from_new_reference(df_target)` instead.

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+
+
+void Ftrl::set_model(robj model) {
+  if (!model.is_frame()) {
+    throw TypeError() << "`model` must be a frame, not "
+        << model.typeobj();
+  }
+
+  DataTable* dt_model_in = model.to_frame();
+  const std::vector<std::string> model_cols_in = dt_model_in->get_names();
```
st-pasha: declare as reference, unless you want to copy the entire vector

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+  if (dt_model_in->nrows == fm->get_d() && dt_model_in->ncols == 2 &&
+    dt_model_in->columns[0]->stype() == SType::FLOAT64 &&
+    dt_model_in->columns[1]->stype() == SType::FLOAT64 &&
+    model_cols_in == FtrlModel::model_cols) {
+    fm->set_model(dt_model_in);
+  } else {
+    throw ValueError() << "FTRL model frame must have " << fm->get_d() <<" rows, and" <<
+                          "2 columns named `z` and `n`, " <<
+                          "both columns must be of `FLOAT64` type.";
+  }
```
st-pasha: it would be better to split this into multiple checks, each with a more specific error message. For example, `"FTRL model frame must have " << fm->get_d() << " rows and 2 columns, whereas your frame has shape " << nrows << " rows and " << ncols << " columns"`. And so on.
This is of course more to write, but it will help the user to diagnose their problems.

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+}
+
+
+oobj Ftrl::get_model(void) const {
+  DataTable* dt_model = fm->get_model();
+  if (dt_model == nullptr) {
+    throw ValueError() << "There is no trained model available, train it first or set.";
```
st-pasha: Why not return `None` instead?

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,338 @@
...
+  }
+  py::Frame* df_model = py::Frame::from_datatable(dt_model);
+  return df_model;
```
st-pasha: again, a likely leak

datatable/extras/ftrl.py
```diff
@@ -20,29 +20,22 @@
...
 # IN THE SOFTWARE.
 #-------------------------------------------------------------------------------
-import datatable as dt
 from datatable.lib import core
 
 
-def ftrl(df_train, df_test, a=0.01, b=1.0, l1=0.0, l2=1.0, 
-         d=10**7, n_epochs=1, inter=False, hash_type=1,
-         seed=0):
+class Ftrl(core.Ftrl):
```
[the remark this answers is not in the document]
oleksiyskononenko: I guess this is just an artifact the the old times, will be removing it.

c/extras/ftrl.h
oleksiyskononenko: This feature is in a debug stage for the moment, I will finalize it as a part of #1391. Need to figure out which hashes will really work well on different platforms, and if we want to have several implementations at all.

c/extras/ftrl.h
oleksiyskononenko: Yes, It was like that before, need to see if I'm going to add anything else to this class, then finalize the order to minimize gaps.

c/extras/py_ftrl.cc
oleksiyskononenko: Yep, once I stabilize the list of arguments will probably convert it to keywords. Also need to work on the default values, that are not obvious to me for the moment.

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, should be `frame`. But where this name is actually used?

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, too extensive refactoring :) Will change documentation.

c/extras/py_ftrl.cc
oleksiyskononenko: Oops. I'm also using the same thing in `Aggregator`. Why this will create a memory leak?

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, can also do that. For the moment I'm not sure which way is better.

c/extras/py_ftrl.cc
st-pasha: The name will be used when generating error messages. Although that part is still incomplete. See `arg.cc` and `args.cc`.

c/extras/py_ftrl.cc
st-pasha: Whenever you're dealing with `PyObject*`s, you have to be really careful that its ref-count is incremented/decremented in equal amounts. This is why we have "python/_all.h" mini-library to begin with. And of course, it is my fault that `py::Frame` is visible externally as `PyObject*` instead of `py::oobj`. 

Anyways, the reason for the leak is the following: `py::Frame::from_datatable` creates a new `py::Frame*` object, which derives from `PyObject*`. Naturally, this object is created with ref-count of 1. When you return that object, the compiler silently converts `df_target` into `py::oobj((PyObject*)df_target)` to match the return type. And the constructor of `py::oobj()` increments the ref-count too (unless you construct via `py::oobj::from_new_reference()`). The end result is that the returned frame will have ref-count of 2. 

Later when the frame is deleted, its ref-count will drop to 1, leaving it as an orphaned pointer. I don't know whether the garbage collector will be able to clean this up, so I wouldn't trust that.

c/extras/py_ftrl.cc
st-pasha: Well, it seems to me that returning `None` is exactly the same as telling the user that "there is no trained model available". But it saves you from writing an extra method `is_model_available()`. However, exceptions are best to use for genuine error conditions. In your case having no trained model is a perfectly valid state, so why should it raise an error?

c/extras/py_ftrl.cc
oleksiyskononenko: Thanks for the explanation. From what I see `py::oobj::from_new_reference` takes `PyObject*` as an argument, and we have `DataTable*`. I see that in other parts of the code you're doing `py::oobj::from_new_reference(py::Frame::from_datatable(dt))`, is it what you suggest and won't this create the same kind of memory leak?

c/extras/py_ftrl.cc
oleksiyskononenko: I imagine the following use-case: someone created an `Ftrl` object, forgot to train it, and saved the model with something like `fm.model.to_csv("my_model.csv")`. As there is no error, one may think that everything went well, transfers this file somewhere to make predictions and get an error on the other side.

Sure, one can make a check weather or not the returned `model` is `None`, but this check is not always done. 

Anyways, I'm now thinking that may be the `model` should always be present. If it was not trained, it could be just zero-initialized. Then it will be always a valid state.

st-pasha: Well, if `fm.model` returns `None`, then calling `.to_csv()` on it is no good :)
On the other hand, if you return a zero-initialized model, then you'll get exactly the situation that you described: someone forgets to train the model, saves it, then makes garbage predictions on the other side.

I guess the question is: could there be a scenario were returning a zero-initialized model is useful? I can't think of any. And it won't let you answer a simple question: was the model trained already or not?

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,350 @@
...
+
+void Ftrl::set_model(robj model) {
+  DataTable* dt_model_in = model.to_frame();
+  const std::vector<std::string>& model_cols_in = dt_model_in->get_names();
+
+  if (dt_model_in->nrows != fm->get_d() || dt_model_in->ncols != 2) {
+    throw ValueError() << "FTRL model frame must have " << fm->get_d() << " rows,"
+                       << "and 2 columns, whereas your frame has " << dt_model_in->nrows
+                       << " rows and " << dt_model_in->ncols << " columns";
+  } else if (model_cols_in != FtrlModel::model_cols) {
```
st-pasha: No need to say `else if` if the previous `if()` condition throws an error or returns.
The code will look cleaner if you write
```
if (dt_model_in->nrows != fm->get_d() || dt_model_in->ncols != 2) { ... }
if (model_cols_in != FtrlModel::model_cols) { ... }
if (dt_model_in->columns[0]->stype() != SType::FLOAT64 ||
    dt_model_in->columns[1]->stype() != SType::FLOAT64) { ... }

fm->set_model(dt_model_in);
```

c/extras/py_ftrl.h
```diff
@@ -0,0 +1,69 @@
...
+          static bool is_subclassable() { return true; }
+          static void init_methods_and_getsets(Methods&, GetSetters&);
+      };
+      void m__init__(PKArgs&);
+      void m__dealloc__();
+      void fit(const PKArgs&);
+      oobj predict(const PKArgs&);
+
+      // Getters and setters.
+      oobj get_model(void) const;
```
st-pasha: This is C-like syntax. In C++ there is no reason to say `void` in the parameter list: https://softwareengineering.stackexchange.com/questions/286490

c/extras/py_ftrl.cc
oleksiyskononenko: Sure, this is something I've been thinking about too. I meant It will always be a valid state internally, once trained we can set up some `was_trained` flag, and then we control what to return as an `fm.model` to the user. May be `None` is not a bad idea, but I would also like some kind of warning.

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,350 @@
...
+  return df_target;
+}
+
+
+/*
+*  Getter and setter for the model datatable.
+*/
+oobj Ftrl::get_model(void) const {
+  DataTable* dt_model = fm->get_model();
+  py::oobj df_model = py::oobj::from_new_reference(py::Frame::from_datatable(dt_model));
```
st-pasha: This may actually throw an exception if `dt_model` is null.

c/extras/py_ftrl.cc
oleksiyskononenko: Sure, I'm still looking through the code and making some adjustments.

c/extras/ftrl.cc
```diff
@@ -165,28 +142,28 @@ dtptr Ftrl::test(const DataTable* dt) {
...
-
   return dt_target;
 }
 
 
 /*
 *  Make a prediction for an array of hashed features.
 */
-double Ftrl::predict(const Uint64Ptr& x, size_t x_size) {
+double FtrlModel::predict_row(const Uint64Ptr& x, size_t x_size) {
```
st-pasha: I think your `x` is not the same as defined in the paper (McMahan&al. 2013). In the paper, `x` is the vector of features for the t-th observation. In your case, it's more like a sparse representation of `x`. I.e., your `x` contains indices of elements within the paper's `x` that are equal to 1, while all others are considered 0. Confusing...

c/extras/ftrl.cc
```diff
@@ -218,12 +195,12 @@ inline double Ftrl::bsigmoid(double x, double b) {
 /*
 *  Update weights based on prediction and the actual target.
 */
-void Ftrl::update(const Uint64Ptr& x, size_t x_size, double p, bool target) {
+void FtrlModel::update(const Uint64Ptr& x, size_t x_size, double p, bool target) {
```
st-pasha: `target` is `y` in the paper, why not call it that?

c/extras/ftrl.cc
```diff
@@ -366,12 +344,148 @@ inline double Ftrl::signum(double x) {
...
+    return nullptr;
+  }
+}
+
+
+/*
+*  Set an FTRL model, assuming all the validation is done in `py_ftrl.cc`
+*/
+void FtrlModel::set_model(DataTable* dt_model_in) {
+  dt_model = dtptr(dt_model_in);
```
st-pasha: I suspect this may crash... You're taking `dt_model_in`, which is a pointer to a `DataTable*` owned from some python object, and put it inside a `unique_ptr`. At some point this `DataTable` will be freed twice.

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,413 @@
...
+}
+
+
+oobj Ftrl::get_seed() const {
+  return py::oint(static_cast<size_t>(fm->get_seed()));
+}
+
+
+void Ftrl::set_a(robj a) {
+  if (!a.is_numeric()) {
```
st-pasha: `a.to_double()` will do the numeric check, no need to do it manually

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,413 @@
...
+
+  if (!(args[7].is_undefined() || args[7].is_none())) {
+    fmp.hash_type = static_cast<unsigned int>(args[7].to_size_t());
+  }
+
+  if (!(args[8].is_undefined() || args[8].is_none())) {
+    fmp.seed = static_cast<unsigned int>(args[8].to_size_t());
+  }
+
+  if (args.num_varkwd_args() > 0) {
```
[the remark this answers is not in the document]
oleksiyskononenko: Hm, then it seems that I don't really understand the purpose of this `has_varkwds`argument. What I need here is simply to identify that there is no unknown keyword argument passed. What is the best way to do it by using `PKArgs`?

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,413 @@
...
+    Number of epochs to train for.
+inter : boolean
+    If feature interactions to be used or not.
+hash_type : int
+    Hashing method to use for strings:
+    `0` - std::hash;
+    `1` - Murmur2;
+    `2` - Murmur3.
+seed: unsigned int
+    Seed to be used for Murmur hash functions.
```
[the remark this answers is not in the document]
oleksiyskononenko: Ok for boolean. But how do you indicate only positive integers are allowed here if you just say `int`?

c/extras/py_ftrl.cc
```diff
@@ -0,0 +1,413 @@
...
+  DataTable* dt_model_in = model.to_frame();
+  const std::vector<std::string>& model_cols_in = dt_model_in->get_names();
+
+  if (dt_model_in->nrows != fm->get_d() || dt_model_in->ncols != 2) {
+    throw ValueError() << "FTRL model frame must have " << fm->get_d() << " rows,"
+                       << "and 2 columns, whereas your frame has " << dt_model_in->nrows
+                       << " rows and " << dt_model_in->ncols << " columns";
+  }
+
+  if (model_cols_in != FtrlModel::model_cols) {
```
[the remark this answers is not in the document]
oleksiyskononenko: Right now the model frame that one can get with `model_frame = fm.model` has this kind of layout: two columns, names `z` and `n`, both `float64`. So there is no way to confuse these two sets of coefficients by looking at the frame content. Then there is a way to re-use the model with `fm.model = model_frame`, so I wanted to be as strict as possible and not load anything that just has two `float64` columns. Because the probability is higher that this frame is not a model frame, but something else. 

In principle, this could be replaced by some convention like `z` is the first column, and `n` is the second one, but then the probability is higher that some mess could be accepted as a model. Right now I just have the same requirements as to how the `model` frame should look like for both getter output and setter input.

c/extras/ftrl.cc
```diff
@@ -25,95 +25,72 @@
...
 *  Train FTRL model on a training dataset.
 */
-void Ftrl::train(const DataTable* dt) {
-  // Define number of features that equal to one bias term
-  // plus `dt->ncols - 1` columns. We assume that the target column
-  // is the last one.
-  n_features = dt->ncols;
+void FtrlModel::fit(const DataTable* dt) {
+  // Define number of features assuming that the target column is the last one.
+  n_features = dt->ncols - 1;
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, there could also be a problem with the frame used for prediction. I've been thinking about the best way to do a sanity check for both of them. We probably need to enforce the frame used for predictions to have one column less, and the same column types (may be only distinguish "strings" and "numerics") as the frame used for training.

c/extras/py_ftrl.cc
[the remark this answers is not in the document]
st-pasha: And there's nothing wrong with accepting "a mess" as a model. If a user wants to provide some weird initial values for the model, let them.
Only check the conditions that are strictly necessary. Like, verify that the coefficients do not have any NAs, and that n-values are all non-negative.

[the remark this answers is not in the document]
oleksiyskononenko: Yes, we can accept a mess if needed. This check doesn't prevent it. What I wanted to do here is to have user be aware about it. So this won't happen just by an accident, because instead of a real model one just passed a dataset for predictions (since this one happens to have 2 `float64` columns). When you have proper names it is more likely that this frame is either already exported model, or user spent time by looking at the dataset.

c/extras/ftrl.cc
```diff
@@ -165,28 +142,28 @@ dtptr Ftrl::test(const DataTable* dt) {
...
   double wTx = 0;
   for (size_t j = 0; j < x_size; ++j) {
     size_t i = x[j];
-    if (fabs(z[i]) <= l1) {
+    if (fabs(z[i]) <= fmp.l1) {
       w[i] = 0;
     } else {
-      w[i] = (signum(z[i]) * l1 - z[i]) / ((b + sqrt(n[i])) / a + l2);
+      w[i] = (signum(z[i]) * fmp.l1 - z[i]) /
+             ((fmp.b + sqrt(n[i])) / fmp.a + fmp.l2);
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, sure, for the moment this just mimics the algo explained in the paper, so it was easier for me to debug. We can do some refactoring.

c/extras/ftrl.cc
oleksiyskononenko: Yes, exactly, it is a sparse representation of `x`. Do you have a suggestion for a better name?

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, but this will have a message that will specify which parameter is not numeric, while the `a.to_double()` will only have a some generic error, right?

c/extras/ftrl.cc
oleksiyskononenko: Yes, you're right. I actually already observed this crash :) Will fix it.

c/extras/py_ftrl.cc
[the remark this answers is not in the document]
st-pasha: You can say so in text. For example, `d` and `n_epochs` are positive too, but you don't specify them as "unsigned". Besides, nothing bad will happen if you allow negative seeds...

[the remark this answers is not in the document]
oleksiyskononenko: Yes, need to specify this for `d` and `n_epochs` too. Sure, negative seed won't break anything because at some point it will be converted to the positive one, but negative `d` and `n_epochs`, for example, won't work. If your convention is just to say it in text, I will change it to be like that. Thanks.

PR #1418, opened 2018-11-16
Fix crash in test_setfns_random

### LINKED ISSUE #1414: Crash in tests/test_sets.py::test_setfns_random[3699163247]
Seed 3699163247 causes the test to fail with a segmentation fault.

st-pasha:
The `sort` function now always returns an array rowindex, never a SliceRowindex.

Closes #1414


st-pasha:
yep, that's why it's "tiny"

PR #1420, opened 2018-11-17
Adjust the styling of Frame's html repr

### LINKED ISSUE #1419: datatable's css style for Jupyter looks bad with dark theme
The style was designed with the default, "light" theme in mind. However, Jupyter supports multiple color schemes, and we should support those other themes too...

st-pasha:
Better Frame styling in both Light and Dark themes.
We no longer use any hard-coded colors, relying instead on Jupyter CSS variables such as `var(--jp-ui-font-color3)` etc. As the values of these variables change with the theme, hopefully, datatable's Frame will remain reasonably well-looking across all possible themes. 

Closes #1419

### DIFF
--- a/c/frame/_repr_html_.cc
+++ b/c/frame/_repr_html_.cc
@@ -1,22 +1,67 @@
 //------------------------------------------------------------------------------
 // Copyright 2018 H2O.ai
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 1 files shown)


c/frame/_repr_html_.cc
```diff
@@ -188,40 +240,51 @@ class HtmlWidget {
 
     void render_styles() {
       if (styles_emitted) return;
-      html << "<style type='text/css'>\n"
+      std::time_t now = std::time(nullptr);
+      std::tm* t = std::localtime(&now);
+      bool xd = (t->tm_mon == 11);
+      bool vd = (t->tm_mon == 4) && (t->tm_wday == 4) &&
+                (t->tm_mday >= 15) && (t->tm_mday <= 21);
```
oleksiyskononenko: :)

PR #1427, opened 2018-11-19
Fix Frame rendering style in docs

### LINKED ISSUE #1426: Update Frame's style in documentation
Now that the style of Frame's HTML representation has been updated (#1419), it's necessary to update the style in the docs too, so that they match.

st-pasha:
- The styling of Frame display in docs is now the same as in Jupyter;
- Row numbers are now displayed comma-separated;
- Comma-separation is not applied to numbers in the range 1000-9999.

Closes #1426

### DIFF
--- a/c/frame/_repr_html_.cc
+++ b/c/frame/_repr_html_.cc
@@ -150,7 +150,9 @@ class HtmlWidget {
 
     void render_data_row(size_t i) {
       html << "    <tr>";
-      html << "<td class='row_index'>" << i << "</td>";
+      html << "<td class='row_index'>";
+      render_comma_separated(i);
+      html << "</td>";
       for (size_t j = 0; j < ncols; ++j) {
         if (j == cols0) {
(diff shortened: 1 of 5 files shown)


c/frame/_repr_html_.cc
```diff
@@ -178,18 +180,18 @@ class HtmlWidget {
 
 
     void render_table_footer() {
-      html << "  <div class='footer'>";
+      html << "  <div class='footer'>\n";
       render_frame_dimensions();
```
oleksiyskononenko: What is the reason you use `\n`s here, and don't use them in `render_data_row()`? I guess they're only required for HTML code to look nice, when opened through "View Source" or something like that?

st-pasha: Yes, from HTML standpoint the newlines don't matter. However, we put that html into the docs (see  docs/using-datatable.rst for example), and we'd like to have at least some amount of readability there.

Issue #1434, opened 2018-11-20
Review and refactor RowIndex usage in Aggregator


oleksiyskononenko:
It looks like there are quite many places in Aggregator, where `RowIndex` is used without any checks for its actual type. Since it could be `RI_ARR32`, `RI_ARR64` or `RI_SLICE`, indices must be accessed in a different way, while in the majority of situations we do `indices32()`.

This must be reviewed and refactored. Probably, adding a new method, called something like `inidices()`, to `RowIndex` class, that will take care about all the different possibilities. We should also move to `RowIndex::strided_loop(...)` for iterations.

PR #1435, opened 2018-11-20
Sort exemplars even when aggregation is bypassed

### LINKED ISSUE #1429: Sort exemplars even when aggregation is bypassed
When we do binning, final exemplars are automatically sorted based on their bin ids. When we bypass aggregation for data sets that have `nrows < min_rows`, this doesn't happen as there are no bin ids. It means that exemplars are saved in the same order, as they go in the original data set. 

However, it seems like `vis-data-server` expects exemplars be always sorted for 1D continuous case, even when we don't do any aggregation. Aggregator can do this, even though no real aggregation is required in this case.

oleksiyskononenko:
Sort exemplars by the first column, when the aggregation is bypassed.

Closes #1429


c/extras/aggregator.cc
```diff
@@ -254,12 +254,28 @@ void Aggregator::aggregate_exemplars(DataTable* dt,
...
+      ri_exemplars.extract_into(temp);
+      ri_exemplars_indices = temp.data();
+    } else if (ri_exemplars.isarr64()){
+      throw ValueError() << "RI_ARR64 is not supported for the moment";
+    }
+
+    auto d_members = static_cast<int32_t*>(dt_members->columns[0]->data_w());
+    for (size_t i = 0; i < dt->nrows; ++i) {
+      d_members[ri_exemplars_indices[i]] = static_cast<int32_t>(i);
+    }
```
st-pasha: If you want to just iterate over the rowindex, the best way to do it is
```
ri_exemplars.strided_loop(0, dt->nrows, 1, 
  [&](int64_t i) {
    d_members[ri_exemplars_indices[i]] = static_cast<int32_t>(i);
  });
```

oleksiyskononenko: Thanks! It's not just one place where I'm using this functionality, will carefully review and refactor all of these shortly as a part of #1434.

oleksiyskononenko: Btw, it seems that I just need `strided_loop2` here...

PR #1441, opened 2018-11-21
Add methods Frame.head() and Frame.tail()

### LINKED ISSUE #1307: implement head and tail method
as in pandas `.head`/`.tail` which nicely corresponds to R methods with same names and default R's data.table print

st-pasha:
Closes #1307

### DIFF
--- a/c/frame/py_frame.cc
+++ b/c/frame/py_frame.cc
@@ -1,21 +1,28 @@
 //------------------------------------------------------------------------------
 // Copyright 2018 H2O.ai
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


c/frame/py_frame.cc
```diff
@@ -160,6 +188,26 @@ void Frame::_clear_types() const {
 }
 
 
+oobj Frame::head(const PKArgs& args) {
+  size_t n = args[0].is_undefined()? 10 : args[0].to_size_t();
+  if (n > dt->nrows) n = dt->nrows;
+  py::otuple aa(2);
+  aa.set(0, py::orange(0, n, 1));
+  aa.set(1, py::None());
```
oleksiyskononenko: What this `py::None()` corresponds to?

st-pasha: that's `None` in python. So, I'm creating a tuple `(range(0, n), None)` here.

PR #1460, opened 2018-12-01
FTRL column hashers, improved API and tests

### LINKED ISSUE #1393: Add tests for FTRL algo
This should include tests for at least
- hash reproducibility;
- single thread and multi-thread learning;
- saving/opening a trained model.

### LINKED ISSUE #1425: Take care about FTRL model parameters
We can now get and set an FTRL model itself, i.e. vectors `z` and `n`. However, we also need to take care about the parameters we've used to train our model, i.e. `alpha`, `beta`, etc. We could use a Python tuple to get and set all of them at once, also there should be a way to relate the model itself with the set of parameters it was trained on.

oleksiyskononenko:
- use column hashers to hash the data, this improves performance by up to 50% and reduces the memory usage;
- capability to get/set parameters by using a Python `namedtuple`;
- streamline API; 
- add tests to cover most of functionality.

Closes #1425
Closes #1393


c/extras/py_ftrl.cc
```diff
@@ -254,156 +384,173 @@ void Ftrl::set_model(robj model) {
...
 
 
-oobj Ftrl::get_b() const {
-  return py::ofloat(fm->get_b());
+oobj PyFtrl::get_params() const {
+  strpair tuple_info;
+  tuple_info.first = "Params";
+  tuple_info.second = "FTRL model parameters";
+
+  py::onamedtupletype nttype(tuple_info, Type::params_info);
```
st-pasha: This should really be a static member of `PyFtrl` class, initialized on first access.

c/python/namedtuple.cc
```diff
@@ -0,0 +1,79 @@
+//------------------------------------------------------------------------------
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// © H2O.ai 2018
+//------------------------------------------------------------------------------
```
st-pasha: how about MIT?

c/python/obj.h
```diff
@@ -128,6 +128,7 @@ class _obj {
     oobj get_attr(const char* attr) const;
     bool has_attr(const char* attr) const;
     oobj get_item(const py::_obj& key) const;
+    static oobj import(const char* module, const char* symbol);
```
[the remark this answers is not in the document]
oleksiyskononenko: Sure, thanks, I still need to review the code changes by myself and perform several adjustments. Made this PR just to see how the tests I've added behave on different systems.

c/extras/py_ftrl.cc
```diff
@@ -254,156 +384,173 @@ void Ftrl::set_model(robj model) {
...
+  nt.set(6, get_inter());
+
+  return std::move(nt);
 }
 
 
-oobj Ftrl::get_l1() const {
-  return py::ofloat(fm->get_l1());
+oobj PyFtrl::get_default_params() const {
+  py::otuple params(7);
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, sure, need to convert this to namedtuple and also wanted to have a namedtuple copy of parameters in the class. Just like you do for column names. Btw, every time one of the column names is adjusted, you also sync this to your memoized copy, right?

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, sure. Still have some work to be done on the tuple/namedtuple/docs here.

c/python/namedtuple.cc
oleksiyskononenko: Guess it came from your `tuple.cc` file that I used as a reference. Will change it shortly. Btw, what about if we just replace it automatically in all the relevant files, so that we have a consistent license in all the places?

tests/extras/test_ftrl.py
```diff
@@ -0,0 +1,410 @@
...
+def test_ftrl_construct_wrong_n_epochs():
+    with pytest.raises(TypeError) as e:
+        ft = core.Ftrl(n_epochs = 10.0)
+    assert ("Argument `n_epochs` in Ftrl() constructor should be an integer, instead "
+            "got <class 'float'>" == str(e.value))
+
+
+def test_ftrl_construct_wrong_inter():
+    with pytest.raises(TypeError) as e:
+        ft = core.Ftrl(inter = 2)
```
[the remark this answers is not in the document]
st-pasha:
```python
from tests import noop

...

def test_ftrl_construct_wrong_inter():
    with pytest.raises(TypeError) as e:
        noop(core.Ftrl(inter = 2))
```

c/extras/ftrl.cc
```diff
@@ -23,74 +23,221 @@
...
+/*
+*  Template class to hash strings.
+*/
+template <typename T>
+class HashString : public Hash {
+  private:
+    const char* strdata;
+    const T* offsets;
+  public:
+    HashString(const Column*);
```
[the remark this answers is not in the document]
st-pasha:
```C++
explicit HashString(const Column*);
```

oleksiyskononenko:
> Hmm, I think I'm gonna disable comments on Codacity, or it may get too annoying :)

Actually I fixed many of Codacity detected issues, like trailing white space, and was about to see the effect, but they all disappeared somehow yesterday. Though, some of them, in my opinion, don't have much of sense.

c/extras/ftrl.cc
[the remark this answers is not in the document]
oleksiyskononenko: Thanks, yes, this is something I already fixed, will be pushing soon.

tests/extras/test_ftrl.py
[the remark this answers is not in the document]
oleksiyskononenko: If I just remove `ft =`, it will start complaining on something else?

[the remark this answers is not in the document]
st-pasha: Don't know, I guess it depends whether they can figure out that `core.Ftrl()` returns a value

c/python/obj.h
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/extras/py_ftrl.cc
```diff
@@ -23,127 +23,212 @@
...
+void PyFtrl::m__init__(PKArgs& args) {
+  FtrlParams fp = Ftrl::params_default;
+  bool defined_params = !(args[0].is_undefined() || args[0].is_none());
+  bool defined_alpha = !(args[1].is_undefined() || args[1].is_none());
+  bool defined_beta = !(args[2].is_undefined() || args[2].is_none());
+  bool defined_lambda1 = !(args[3].is_undefined() || args[3].is_none());
+  bool defined_lambda2 = !(args[4].is_undefined() || args[4].is_none());
+  bool defined_d = !(args[5].is_undefined() || args[5].is_none());
+  bool defined_n_epochs= !(args[6].is_undefined() || args[6].is_none());
+  bool defined_inter = !(args[7].is_undefined() || args[7].is_none());
```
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/extras/py_ftrl.cc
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/python/namedtuple.h
```diff
@@ -0,0 +1,52 @@
...
+    onamedtupletype(strpair, std::vector<strpair>);
+    ~onamedtupletype();
+
+    friend class onamedtuple;
+};
+
+
+class onamedtuple : public otuple {
+  public:
+    onamedtuple(onamedtupletype*);
```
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/extras/py_ftrl.cc
oleksiyskononenko: Fixed

c/extras/ftrl.cc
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

tests/extras/test_ftrl.py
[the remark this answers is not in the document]
oleksiyskononenko: Fixed with `noop` to be safe

c/extras/ftrl.cc
```diff
@@ -363,129 +347,139 @@ DataTable* FtrlModel::get_model() {
...
 }
 
 
-void FtrlModel::set_a(double a_in) {
-  if (fmp.a != a_in) {
-    fmp.a = a_in;
+void Ftrl::set_alpha(double a_in) {
+  if (params.alpha != a_in) {
+    params.alpha = a_in;
     init_model();
```
st-pasha: Why doing `init_model()` here? This is expensive (esp. if you need to set multiple parameters), and probably unnecessary.

c/extras/ftrl.h
```diff
@@ -19,17 +19,23 @@
...
-    double b;
-    double l1;
-    double l2;
+
+struct FtrlParams {
+    double alpha;
+    double beta;
+    double lambda1;
+    double lambda2;
     uint64_t d;
```
st-pasha: why is this not `size_t` ?

c/extras/py_ftrl.cc
```diff
@@ -23,127 +23,220 @@
...
-  FtrlModelParams fmp = FtrlModel::fmp_default;
-
-  if (!(args[0].is_undefined() || args[0].is_none())) {
-    fmp.a = args[0].to_double();
-  }
+PKArgs PyFtrl::Type::args___init__(0, 0, 10, false, false,
+                                   {"params", "alpha", "beta", "lambda1",
+                                   "lambda2", "d", "n_epochs", "inter",
+                                   "hash_type", "seed"},
+                                   "__init__", nullptr);
```
st-pasha: You should allow `params` to be an unnamed argument too (or perhaps, only unnamed?) `core.Ftrl(model_params)`

c/extras/py_ftrl.cc
```diff
@@ -23,127 +23,220 @@
...
     Number of epochs to train for.
 inter : bool
     If feature interactions to be used or not.
-hash_type : int
-    Hashing method to use for strings:
-    `0` - std::hash;
-    `1` - Murmur2;
-    `2` - Murmur3.
-seed: int
-    Seed to be used for Murmur hash functions.
```
st-pasha: Why were these removed here, yet they are still parsed in the constructor?

c/extras/py_ftrl.cc
```diff
@@ -23,127 +23,220 @@
...
+  mm.add<&PyFtrl::fit, args_fit>();
+  mm.add<&PyFtrl::predict, args_predict>();
+  mm.add<&PyFtrl::reset_model, args_reset_model>();
+  mm.add<&PyFtrl::reset_params, args_reset_params>();
 }
 
 
-PKArgs Ftrl::Type::args_fit(1, 0, 0, false, false, {"frame"}, "fit",
+PKArgs PyFtrl::Type::args_fit(1, 0, 0, false, false, {"frame"}, "fit",
 R"(fit(self, frame)
```
st-pasha: The standard scikit signature is `fit(X, y)` (note that `X` is capitalized). The target column `y` is provided by the user separately

c/extras/py_ftrl.cc
```diff
@@ -152,21 +245,33 @@ Train an FTRL model on a dataset.
...
+  }
+  if (dt_train->columns[dt_train->ncols - 1]->stype() != SType::BOOL ) {
+    throw ValueError() << "Last column in a training frame must have a `bool` "
+                       << "type";
+  }
+  ft->fit(dt_train);
 }
 
 
-PKArgs Ftrl::Type::args_predict(1, 0, 0, false, false, {"frame"}, "predict",
```
st-pasha: Standard scikit signature is `predict(X)`.

c/extras/ftrl.cc
```diff
@@ -19,91 +19,136 @@
...
 */
-FtrlModel::FtrlModel(FtrlModelParams fmp_in)
+Ftrl::Ftrl(FtrlParams params_in)
 {
   // Set model parameters
-  fmp = fmp_in;
+  params = params_in;
   // Create and initialize model datatable and weight vector.
   create_model();
   init_model();
```
st-pasha: Model instantiation should be delayed until it becomes needed.

c/extras/py_ftrl.cc
```diff
@@ -212,198 +328,247 @@ Returns
...
+----------
+    None
+
+Returns
+----------
+    None
+)");
+
+
+void PyFtrl::reset_params(const PKArgs&) {
```
st-pasha: Why is this method needed? I could just write `ft.params = Ftrl().params`.

c/extras/py_ftrl.cc
```diff
@@ -212,198 +328,247 @@ Returns
...
+  params.set(4, get_d());
+  params.set(5, get_n_epochs());
+  params.set(6, get_inter());
+  return std::move(params);
 }
 
 
-oobj Ftrl::get_l1() const {
-  return py::ofloat(fm->get_l1());
+oobj PyFtrl::get_default_params() const {
```
st-pasha: Also not needed: same as `Ftrl().params`

tests/extras/test_ftrl.py
```diff
@@ -0,0 +1,410 @@
...
+        ft.inter = 2
+    assert ("Expected a boolean, instead got <class 'int'>" == str(e.value))
+
+
+def test_ftrl_set_wrong_params_type():
+    ft = core.Ftrl()
+    params = test_params._replace(alpha = "1.0")
+    with pytest.raises(TypeError) as e:
+        ft.params = params
+    assert ("Expected a float, instead got <class 'str'>" == str(e.value))
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, missed this one. What is the best way to customize the error message? Using a custom `error_manager` for each parameter?

tests/extras/test_ftrl.py
```diff
@@ -0,0 +1,410 @@
...
+def test_ftrl_model_none():
+    ft = core.Ftrl()
+    assert ft.model == None
+
+
+def test_ftrl_get_set_reset_model():
+    ft = core.Ftrl(params = test_params)
+    model = dt.Frame([[random.random() for i in range(test_params.d)],
+                     [random.random() for i in range(test_params.d)]],
+                     names=['z', 'n'])
```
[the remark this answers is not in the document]
oleksiyskononenko: Right, will add a check for that too.

tests/extras/test_ftrl.py
```diff
@@ -0,0 +1,410 @@
...
+                   d = test_params.d, n_epochs = test_params.n_epochs,
+                   inter = test_params.inter)
+    assert ft.params == (test_params.alpha, test_params.beta,
+                         test_params.lambda1, test_params.lambda2,
+                         test_params.d, test_params.n_epochs, test_params.inter)
+
+
+#-------------------------------------------------------------------------------
+# Test getters, setters and reset methods for FTRL parameters
+#-------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, also it seems that for a combinations of `alpha`, `beta` and `lambda2` there could be  `n_i` for which denominator will be zero...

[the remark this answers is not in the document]
oleksiyskononenko: Need to check what are the restrictions for these parameters, so that may be it is not the case.

c/python/namedtuple.cc
```diff
@@ -0,0 +1,97 @@
...
+  py::otuple args(2);
+  args.set(0, py::ostring(tuple_info.first));
+  args.set(1, argnames);
+
+  auto type = namedtuple.call(args);
+  auto res = std::move(type).release();
+
+  // Set namedtuple doc
+  PyObject_SetAttrString(res, "__doc__",
+                         py::ostring(tuple_info.second).to_borrowed_ref());
```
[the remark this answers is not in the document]
oleksiyskononenko: You mean something like this?

`  py::ostring docstr = py::ostring(tuple_info.second);
  PyObject_SetAttrString(res, "__doc__",
                         docstr.to_borrowed_ref());`

c/datatablemodule.cc
```diff
@@ -235,7 +235,7 @@ PyInit__datatable()
     init_jay();
 
     py::Frame::Type::init(m);
-    py::Ftrl::Type::init(m);
+    py::PyFtrl::Type::init(m);
```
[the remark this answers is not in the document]
oleksiyskononenko: Because I already have `Ftrl` class defined in `ftrl.h`, so would prefer to have different name here. In the current framework, can I simply move `PyFtrl` out of `py` namespace?

c/extras/ftrl.cc
oleksiyskononenko: Because I don't know what will happen when one restarts from a model that was created based on a different `alpha` (also `beta`, `lambda1`, etc).

c/extras/ftrl.h
oleksiyskononenko: Because `uint64_t` is guaranteed to be 64-bit, while `size_t` is not (may be in most of the cases they're identical, but from my understanding there is still a possibility they may differ). Since the Murmur hasher I'm currently using is designed for 64-bits `uint64_t`, I just want to be safe.

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, it will look nicer than. Thanks for suggestion.

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, I don't think we will need them after working on #1391. I will clean the rest of the code, because these parameters don't have any effect on the current hashers in any case.

c/extras/py_ftrl.cc
oleksiyskononenko: Do you know if this signature is also something standard in DAI for other algos?

c/extras/py_ftrl.cc
oleksiyskononenko: Just a convenience method.

c/extras/py_ftrl.cc
oleksiyskononenko: Yes, this one can be removed.

c/datatablemodule.cc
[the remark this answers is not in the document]
oleksiyskononenko: Yes, I understand there will be no conflict, what I don't like is using the same name for two different things, even if they're in different namespaces. Seems like a potential source of confusion.

c/extras/ftrl.cc
st-pasha: The "expensive" part refers to the fact that when setting the model's parameters, you're calling `set_alpha()`, `set_beta()`, etc -- and each of them does `init_model()` over and over again.

The "unnecessary" comment is less clear. On one hand, we can prohibit changing the model's parameters altogether once it has been trained. Or alternatively, allow the parameters to be changed, and then start training with the existing model data. It is up to the user whether to reset the data or not. However, I feel that force-clearing the data without the user's consent is the worst choice in this case.

c/extras/ftrl.h
st-pasha: See `types.cc`: datatable only compiles on platforms where `sizeof(size_t) == sizeof(int64_t) == sizeof(void*) == 8`.

c/extras/ftrl.cc
oleksiyskononenko: I agree on the first part. My comment was more about why we may want to discard the trained model, when some parameters are changed.

Yes, when a user changes the individual parameters one by one, it now does unnecessary job of cleaning the model again, and again. I've been actually thinking about changing it to something like this:

- since now one can use `params` to set all the input parameters at once, we may not actually need all the individual getters and setters;
- we can keep the model state when the parameters change, because there is always a `reset_model()` method to initialize the model;
- we can add an additional argument to `set_params` that will control whether or not the model must be initialized, this argument could be `false` by default. In this case it will be up to the user to decide what to do with the trained model.

c/extras/py_ftrl.cc
st-pasha: it's used in DAI's transformers, and also in `h2o`s models

c/extras/ftrl.h
oleksiyskononenko: Ah, good to know.

c/extras/py_ftrl.cc
oleksiyskononenko: Gotcha, then let me make it closer to the standard.

c/extras/py_ftrl.cc
st-pasha: Convenience methods are good for frequently used operations. For example, we have `frame.head()`, which is a shortcut for `frame[:10, :]`. But I don't think resetting parameters to default is that useful to warrant a dedicated method.

oleksiyskononenko: Frankly, I'm not sure about all the use cases for the moment... anyways, it doesn't seem to be a big deal to keep it or remove. What I guess is more important are the values we should use for the default parameters, or may be don't have default parameters at all... Right now I just used a common sense and experience gained through the tests to set them up.

st-pasha: Well, there is a cost in properly documenting/testing every new method, which is why I was suggesting to remove it.
As for the default values of the parameters, don't they have something in the paper regarding which parameters they found optimal for one of their experiments?

oleksiyskononenko: I will easily remove it if it makes other's life simpler :) 

Regarding the default parameters, this is the only thing they say:
`The optimal value of α can vary a fair bit  depending  on  the  features  and  dataset,  and
β =  1  is usually good enough; this simply ensures that early learning rates are not too high.`

So I believe there are no "magic" numbers that will work for all the cases. I will just keep it as they're now, and we will see.

c/datatablemodule.cc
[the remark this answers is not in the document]
st-pasha: This is up to you, but keep in mind that we tend to use this kind of convention in other places, when we need to expose an object to python. Things like `py::base_expr`, `py::Frame`, `py::Column`, `py::Rowindex`, etc. This convention is not followed always, but that's mostly due to historic circumstances: there is an element of evolution to this. Let me explain how it came out to be:

Originally we too started with "Py" prefixes: there were "PyDatatable", "PyColumn", "PyRowindex", etc. But this system was confusing in 3 respects:
- it is exactly the same convention as used by the core python. Thus, there is a small chance of accidentally creating an object with the same name as an existing python object. Or vice versa, a new version of Python may add an object that clashes with our existing object. 
- for a person reading the code it may be difficult to understand which of the objects are from python standard library, and which "belong" to datatable.
- prefixing every name with Py is exactly like creating a "fake" namespace. They do it in C because C lacks "real" namespaces, but why should we fake it in C++?

And if you still feel like there is a potential for confusion here, one way to address it is to avoid using the `namespace` keyword. Thus, you'd be forced to always write explicitly `py::Ftrl` or `dt::Ftrl`, which is unambiguous.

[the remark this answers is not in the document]
oleksiyskononenko: Thanks for the detailed explanation. Yes, I actually never use `namespace` keyword, because then there is a high chance to get into a trouble. 

What you say makes perfect sense and I would prefer to follow the convention. Will convert it to`py::Ftrl` and `dt::Ftrl` shortly.

c/extras/py_ftrl.cc
oleksiyskononenko: Fixed

tests/extras/test_ftrl.py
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/extras/ftrl.h
oleksiyskononenko: Guess I will leave this as `uint64_t` for the moment, because otherwise `static_cast` will be needed for each call of Murmur hasher. Will review this in the next PR, may be some adjustments are also needed to the hasher code.

PR #1465, opened 2018-12-04
Minor code fixes recommended by Codacity

st-pasha:


st-pasha:
For some reason, they are lagging in their updates. Maybe tomorrow.

PR #1470, opened 2018-12-08
Namedtuple functionality

oleksiyskononenko:
As a part of #1425, add `onamedtuple` and `onamedtupletype` classes to deal with Python namedtuples.

Also, several new "to" methods to deal with negative and positive values are added to `oobj` class. They are required for #1460.


st-pasha:
The `obj.to_*()` methods are designed to convert from `PyObject*`s into native C++ types. However, a "positive double" is not a C++ type. This conversion conflates type checking which *must* be done in `obj` class using python API, with value checking which can be easily done outside of the `obj` class. That class is already too big, let's not bloat it unnecessarily.

Consider this: if we add "double > 0" type today, then tomorrow someone will ask for a "double in interval [0, 1)" type, or an "integer from 1 to 5", etc. This is not the direction we want to go...

oleksiyskononenko:
Well, I can easily keep it in `py::ftrl`, but thought positive/negative checks (they seem to be more generic than the examples you provide) could also be useful in other places.

st-pasha:
@oleksiyskononenko I have no doubt that "> 0" could be useful in other places. But the question is: should this check be the job of the `obj` class? My approach so far was to handle only **type checks** in `obj`. This already makes the class unbearably big... If we erase this line, then where will we draw it next?

That said, there is nothing preventing you from creating a new helper class containing common functions for value-checking. So far I've only been doing it with helper functions at file level, but we can do it globally too.

oleksiyskononenko:
Sure, if `obj` was designed to make type checks only, we should probably create a new class for validation of parameters. Btw, what about the `Arg` class, also only type checks?

PR #1471, opened 2018-12-08
namedtuple API adjustments

st-pasha:
- prefer C++ `using` instead of `typedef`;
- fixed docstring of parameter `lambda2`;
- use `std::string`s instead of `const char*` for namedtuple construction;
- rename `Params` -> `FtrlParams`;
- defer creation of `FtrlParams` named tuple type until it is actually needed;
- add another constructor for `onamedtuple`, without the docstrings;
- add copy/move constructor for `onamedtuple` and `onamedtupletype`.

### DIFF
--- a/c/extras/py_ftrl.cc
+++ b/c/extras/py_ftrl.cc
@@ -32,7 +32,7 @@ PKArgs Ftrl::Type::args___init__(1, 0, 7, false, false,
                                  "lambda2", "d", "n_epochs", "inter"},
                                  "__init__", nullptr);
 
-std::vector<strpair> Ftrl::params_fields_info = {
+static std::vector<strpair> params_fields_info = {
   strpair("alpha", "`alpha` in per-coordinate FTRL-Proximal algorithm"),
   strpair("beta", "`beta` in per-coordinate FTRL-Proximal algorithm"),
   strpair("lambda1", "L1 regularization parameter"),
@@ -42,9 +42,13 @@ std::vector<strpair> Ftrl::params_fields_info = {
   strpair("inter", "Parameter that controls if feature interactions to be used "
                    "or not")
 };
-strpair Ftrl::params_info = strpair("Params", "FTRL model parameters");
-onamedtupletype Ftrl::params_nttype(Ftrl::params_info,
-                                    Ftrl::params_fields_info);
+
+static onamedtupletype& _get_params_namedtupletype() {
+  static onamedtupletype ntt(
+    strpair("FtrlParams", "FTRL model parameters"),
+    params_fields_info);
+  return ntt;
+}
 
 
 void Ftrl::m__init__(PKArgs& args) {
@@ -124,7 +128,7 @@ const char* Ftrl::Type::classname() {
 
 const char* Ftrl::Type::classdoc() {
   return R"(Follow the Regularized Leader (FTRL) model with hashing trick.
-    
+
 See this reference for more details:
 https://www.eecs.tufts.edu/~dsculley/papers/ad-click-prediction.pdf
 
@@ -139,7 +143,7 @@ lambda1 : float
 lambda2 : float
     L2 regularization parameter.
 d : int
-    Number of bins to be used after the hashing trick. 
+    Number of bins to be used after the hashing trick.
 n_epochs : int
     Number of epochs to train for.
 inter : bool
@@ -217,11 +221,11 @@ Train an FTRL model on a dataset.
 Parameters
 ----------
 X: Frame
-    Datatable frame of shape (nrows, ncols) to be trained on. 
+    Datatable frame of shape (nrows, ncols) to be trained on.
 
 y: Frame
-    Datatable frame of shape (nrows, 1), i.e. the target column. 
-    This column must have a `bool` type. 
+    Datatable frame of shape (nrows, 1), i.e. the target column.
+    This column must have a `bool` type.
 
 Returns
 ----------
@@ -278,12 +282,12 @@ Make predictions for a dataset.
 Parameters
 ----------
 X: Frame
-    Datatable frame of shape (nrows, ncols) to make predictions for. 
+    Datatable frame of shape (nrows, ncols) to make predictions for.
     It must have the same number of columns as the training frame.
 
 Returns
 ----------
-    A new datatable frame of shape (nrows, 1) with a prediction 
+    A new datatable frame of shape (nrows, 1) with a prediction
     for each row of frame X.
 )");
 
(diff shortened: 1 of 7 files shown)


c/extras/py_ftrl.cc
```diff
@@ -24,60 +24,60 @@
...
-onamedtupletype Ftrl::params_nttype(Ftrl::params_info,
-                                    Ftrl::params_fields_info);
+static const char* doc_alpha    = "`alpha` in per-coordinate FTRL-Proximal algorithm";
+static const char* doc_beta     = "`beta` in per-coordinate FTRL-Proximal algorithm";
+static const char* doc_lambda1  = "L1 regularization parameter";
+static const char* doc_lambda2  = "L2 regularization parameter";
+static const char* doc_d        = "Number of bins to be used for the hashing trick";
+static const char* doc_n_epochs = "Number of epochs to train a model";
+static const char* doc_inter    = "Parameter that controls if feature interactions to be used or not";
+
```
oleksiyskononenko: My idea was to iterate over this vector to handle parameters automatically at some point. I really don't like that we have to specify `alpha`, `beta`, and others many times in different parts of `py::Ftrl` class. May not be possible to automate it by 100%, but at would like to reduce number of duplicates at least.

c/extras/py_ftrl.cc
```diff
@@ -217,11 +189,11 @@ Train an FTRL model on a dataset.
 Parameters
 ----------
 X: Frame
-    Datatable frame of shape (nrows, ncols) to be trained on. 
+    Datatable frame of shape (nrows, ncols) to be trained on.
```
oleksiyskononenko: So should we say `datatable frame` or just `frame` in documentation and error messages?

c/python/namedtuple.cc
```diff
@@ -27,50 +27,68 @@
...
+    args_prop.set(1, py::None());
+    args_prop.set(2, py::None());
+    for (size_t i = 0; i < field_docs.size(); ++i) {
+      args_itemgetter.set(0, py::oint(i));
+      args_prop.set(0, itemgetter.call(args_itemgetter));
+      args_prop.set(3, py::ostring(field_docs[i]));
+      auto prop = property.call(args_prop);
+      PyObject_SetAttrString(res, field_names[i].c_str(),
+                             prop.to_borrowed_ref());
+    }
```
oleksiyskononenko: If we don't use `strpair` to pass field names and field docs, then number of names could be different than the number of docs. For example, there is nothing preventing `field_docs.size()` to be larger than `nfields`. Do you know what will happen in this case?

c/extras/py_ftrl.cc
```diff
@@ -304,9 +276,10 @@ oobj Ftrl::predict(const PKArgs& args) {
 
   size_t n_features = dtft->get_n_features();
   if (dt_X->ncols != n_features && n_features != 0) {
-    throw ValueError() << "Can only predict on a frame that has "<< n_features
-                       << " column(s), i.e. has the same number of features as "
-                       << "was used for model training";
+    throw ValueError() << "Can only predict on a Frame that has " << n_features
+                       << " column" << (n_features == 1? "" : "s")
```
oleksiyskononenko: I guess at some point I saw a `plural` function in datatable, but when I wanted to use it here, I couldn't find it. Do I confuse it with something, or there is really something like that?

c/extras/py_ftrl.cc
st-pasha: I'm not sure how this could work? Parameters have different types, different value constraints, tied to different fields in the Ftrl class, their getters/setters must each be separate functions, etc. If you do come up with an interesting approach here, it would be very quick to put the docstrings into an array.

oleksiyskononenko: I'm not sure either for the moment. Some thoughts:
- most of getters and setters follow exactly the same pattern in both their names (`get_smth`  /`set_smth`), and their implementation (simple return / assignment). If there is a getter or setter that is different, it could be supplied as a custom method. Others, in my case this is a majority of parameters, could probably be automated;
- if we have a set of predefined validation rules (like sign check, equal to zero check, within an interval, etc.) this will probably cover most of the constrains. Again, if there is something that requires a custom validator, it could be set as an external method;
- yes, types should be supplied in the same array along with the parameter names and their descriptions.

c/extras/py_ftrl.cc
oleksiyskononenko: I see, but there are other packages that provide different types of frames, like for example Pandas, right? Or we assume if a user imports `datatable`, she or he knows what is meant by frame here?

c/python/namedtuple.cc
oleksiyskononenko: Any particular reason why you prefer not to use `std::pair`?

c/extras/py_ftrl.cc
oleksiyskononenko: Ah, that explains why I couldn't find it at some point :)

c/extras/py_ftrl.cc
st-pasha: If the user is reading `datatable` documentation, we can assume that he/she can guess that the subject is datatable Frame :) The only exception may be when we're talking about interoperability with other libraries, for example `pandas`. Then it could make sense to explicitly say "pandas DataFrame" or "datatable Frame".

oleksiyskononenko: Ok, let's hope for the best :) I will then clean all the places like this.

c/python/namedtuple.cc
st-pasha: Yes, because I wanted to support a common use-case when there are no doc-strings. For example, consider converting a `Frame` into a list of namedtuples. In this case there are only the names of the fields. Of course, I could have created strpairs with empty docstrings, but this seems like a rather awkward transform for no good reason.

st-pasha: Though on a second thought, we could have used a `struct { std::string name, doc; }` there and achieved the same result provided the struct had suitable set of constructors.

oleksiyskononenko: I see, but don't you have to create an empty `field_docs` now? Due to this `xassert(field_docs.size() == nfields)`, it seems that you require `field_docs` to have the same size as `field_names`, i.e. one need to create an empty `std::vector` of doc strings.

oleksiyskononenko: Ah, sorry missed the `!field_docs.empty()` condition. Gotcha.

oleksiyskononenko: Yes, that's also a possibility...

st-pasha: How about now?

oleksiyskononenko: Like it much better now.

PR #1480, opened 2018-12-13
Review and refactor validation in `py::Ftrl` class

oleksiyskononenko:
- remove `to_*_positive` and `to_*_not_negative` methods from the both `py::_obj` and `py::Arg`;
- create a namespace `py::Validator` instead, that has a list of functions for data validation;
- minor adjustments to the FTRL tests;
- minor adjustments to text and cosmetics.


c/extras/py_ftrl.cc
```diff
@@ -38,7 +40,7 @@ static const char* doc_lambda1  = "L1 regularization parameter";
 static const char* doc_lambda2  = "L2 regularization parameter";
 static const char* doc_d        = "Number of bins to be used for the hashing trick";
 static const char* doc_n_epochs = "Number of epochs to train a model";
-static const char* doc_inter    = "Parameter that controls if feature interactions to be used or not";
+static const char* doc_inter    = "If feature interactions to be used or not";
```
st-pasha: That doesn't sound grammatically correct...

oleksiyskononenko: Yes, thanks. Fixed it with "Switch to enable second order feature interaction", sounds better?

PR #1482, opened 2018-12-14
Add support for pickling Ftrl objects

### LINKED ISSUE #1476: Add support for pickling datatable.core.Ftrl objects
`TypeError: can't pickle datatable.core.Ftrl objects`

oleksiyskononenko:
- make `Ftrl` class to be a part of `datatable.models`;
- add support for pickling `Ftrl` objects that contain the both parameters and the model. Since namedtuples cannot be easily pickled, we pickle parameters as a normal tuple adding internal methods to get/set a normal tuple of parameters. These getters and setters are not exposed to the user;
- add `w` weight initialization in the model setter, otherwise it would not possible to fit / predict after unpickling;
- add the corresponding tests;
- minor cleaning and refactoring.

Closes #1476

### DIFF
--- a/c/extras/dt_ftrl.cc
+++ b/c/extras/dt_ftrl.cc
@@ -43,17 +43,17 @@ Ftrl::Ftrl(FtrlParams params_in) :
   params(params_in),
   n_features(0),
   n_inter_features(0),
+  w(nullptr),
   model_trained(false)
 {
 }
 
 
 void Ftrl::create_model() {
-  w = doubleptr(new double[params.d]());
   Column* col_z = Column::new_data_column(SType::FLOAT64, params.d);
   Column* col_n = Column::new_data_column(SType::FLOAT64, params.d);
   dt_model = dtptr(new DataTable({col_z, col_n}, model_cols));
-  init_zn();
+  init_weights();
   reset_model();
 }
 
@@ -65,9 +65,10 @@ void Ftrl::reset_model() {
 }
 
 
-void Ftrl::init_zn() {
+void Ftrl::init_weights() {
   z = static_cast<double*>(dt_model->columns[0]->data_w());
   n = static_cast<double*>(dt_model->columns[1]->data_w());
+  w = doubleptr(new double[params.d]());
 }
 
 
@@ -383,7 +384,7 @@ size_t Ftrl::get_n_epochs() {
 void Ftrl::set_model(DataTable* dt_model_in) {
   dt_model = dtptr(dt_model_in->copy());
   set_d(dt_model->nrows);
-  init_zn();
+  init_weights();
   n_features = 0;
   model_trained = true;
 }
--- a/c/extras/dt_ftrl.cc
+++ b/c/extras/dt_ftrl.cc
@@ -37,13 +37,11 @@ const FtrlParams Ftrl::params_default = {0.005, 1.0, 0.0, 1.0,
 *  Set up FTRL parameters and initialize weights.
 */
 Ftrl::Ftrl(FtrlParams params_in) :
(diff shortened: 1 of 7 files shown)


c/extras/dt_ftrl.cc
```diff
@@ -43,17 +43,17 @@ Ftrl::Ftrl(FtrlParams params_in) :
   params(params_in),
   n_features(0),
   n_inter_features(0),
+  w(nullptr),
```
st-pasha: `std::unique_ptr` already by default initializes with `nullptr` -- no need to say that explicitly

tests/extras/test_ftrl.py
```diff
@@ -574,3 +575,41 @@ def test_ftrl_fit_predict_string():
...
+
+
+#-------------------------------------------------------------------------------
+# Test pickling
+#-------------------------------------------------------------------------------
+
+def test_ftrl_pickling():
+    ft = Ftrl(d = 10)
+    df_train = dt.Frame([[i for i in range(ft.d)]])
+    df_target = dt.Frame([[True for i in range(ft.d)]])
```
st-pasha: You could also write
```
df_train = dt.Frame(range(ft.d))
df_target = dt.Frame([True] * ft.d)
```

tests/extras/test_ftrl.py
```diff
@@ -43,9 +43,9 @@
 tparams = Params(alpha = 1, beta = 2, lambda1 = 3, lambda2 = 4, d = 5,
                      n_epochs = 6, inter = True)
 
-tmodel = dt.Frame([[random.random() for i in range(tparams.d)],
-                      [random.random() for i in range(tparams.d)]],
-                      names=['z', 'n'])
+tmodel = dt.Frame([[random.random()] * tparams.d,
+                   [random.random()] * tparams.d],
```
st-pasha: Here, the old statement was more correct:
`[random.random() for _ in range(N)]` creates a list of N different random numbers;
`[random.random()] * N` creates a list of N same numbers, although that number is chosen randomly.

tests/extras/test_aggregate.py
```diff
@@ -492,9 +492,9 @@ def aggregate_nd(nd):
     nrows = 1000
     div = 50
     column = [i % div for i in range(nrows)]
-    matrix = [column for i in range(nd)]
+    matrix = [column] * nd
     out_types = [ltype.int] * nd + [ltype.int]
-    out_value = [[i for i in range(div)]] * nd + \
+    out_value = [list(range(div))] * nd + \
                 [[nrows // div for i in range(div)]]
```
st-pasha: could be `[nrows // div] * div` too

tests/extras/test_ftrl.py
```diff
@@ -365,8 +365,8 @@ def test_ftrl_set_wrong_shape_model():
 
 def test_ftrl_set_wrong_type_model():
     ft = Ftrl(tparams)
-    model = dt.Frame([["foo" for i in range(tparams.d)],
-                      [random.random() for i in range(tparams.d)]],
+    model = dt.Frame([["foo"] * tparams.d,
+                      [random.random()] * tparams.d],
```
st-pasha: same as above

tests/extras/test_aggregate.py
oleksiyskononenko: yeah, this one I already fixed, see next commit

tests/extras/test_ftrl.py
oleksiyskononenko: Both will probably work in this case, however, may be for the purpose of testing different random numbers are better. Will revert it back, thanks.

Issue #1496, opened 2018-12-19
Support for i selector an integer column with value -1


st-pasha:
If `x` is an integer-valued Frame then `DT[x, :]` selects rows `x.to_list()` from the frame. Currently, `x` cannot contain negative values or values greater or equal than `DT.nrows`.

This issue proposes to allow value `-1` in `x`, which would signify selection of an empty row (similar to join operations). We already allow "-1" in a RowIndex internally, so there is no reason to prohibit this use.

With this proposal there will be a guarantee that a RowIndex extracted from a Frame can always be applied back to that Frame.

PR #1497, opened 2018-12-19
Improve test coverage for i_node.cc

st-pasha:
- Many tests in `test_dt_rows.py` used old-fashioned style `DT(rows=...)` of selecting elements from a Frame. Consequently, those tests were not running for the new-style selectors `DT[i, j]` expressions. As a result, several bugs were inadvertently introduced during the latest round of refactoring, as the tests effectively ignored the new code. This PR addresses this issue, and transforms all tests in `test_dt_rows.py` into the "new-style" selectors of the form `DT[i, j]`. Eventually the old-style calling method `DT(...)` will be deprecated and later removed.

- After the tests were enabled, few more bugs were discovered and fixed, in particular bugs involving "multislice" `i` selectors;

- Fixed coverage script `llvm-gcov.sh`;

- Handle correctly all reducer nodes, including sd, min, max and mean;

- Several tests were added in order to achieve 100% coverage of `i_node.cc`;

- In some tests error messages were improved.

### DIFF
--- a/c/rowindex_array.cc
+++ b/c/rowindex_array.cc
@@ -403,7 +403,7 @@ void ArrayRowIndexImpl::init_from_integer_column(const Column* col) {
   }
   int64_t imin = col->min_int64();
   int64_t imax = col->max_int64();
-  if (imin < 0) {
+  if (imin < -1) {
     throw ValueError() << "Row indices in integer column cannot be negative";
   }
   min = static_cast<size_t>(imin);
--- a/ci/llvm-gcov.sh
+++ b/ci/llvm-gcov.sh
@@ -1,2 +1,21 @@
 #!/bin/bash
-exec "$LLVM4$LLVM5$LLVM6/bin/llvm-cov" gcov "$@"
+
+if [ -n "$LLVM7" ]; then
+	exec "$LLVM7/bin/llvm-cov" gcov "$@"
+
+elif [ -n "$LLVM6" ]; then
+	exec "$LLVM6/bin/llvm-cov" gcov "$@"
+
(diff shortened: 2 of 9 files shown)


c/rowindex_array.cc
```diff
@@ -403,7 +403,7 @@ void ArrayRowIndexImpl::init_from_integer_column(const Column* col) {
   }
   int64_t imin = col->min_int64();
   int64_t imax = col->max_int64();
-  if (imin < 0) {
+  if (imin < -1) {
```
oleksiyskononenko: So `imin` is allowed to be `-1`?

ci/llvm-gcov.sh
```diff
@@ -1,2 +1,21 @@
 #!/bin/bash
-exec "$LLVM4$LLVM5$LLVM6/bin/llvm-cov" gcov "$@"
+
+if [ -n "$LLVM7" ]; then
```
oleksiyskononenko: Should we update documentation online saying that LLVM7 is also supported? I guess I checked it yesterday, and it was only 4/5/6.

c/rowindex_array.cc
oleksiyskononenko: Ah, this seems to be connected to #1496?

st-pasha: yes

ci/llvm-gcov.sh
st-pasha: Well, we never tested with LLVM7, so it's too early to claim support. However, eventually when we add the support, this script won't have to be modified.

PR #1505, opened 2018-12-24
Make FTRL algo to work properly on views

### LINKED ISSUE #1502: [bug] FTRL log loss increases with number of epochs
Hi,

I'm trying to use datatable FTRL proximal algo on a dataset and it behaves strangely.
LogLoss increases with the number of epochs.

Here is the code I use : 

```
train_dt = dt.fread('dt_ftrl_test_set.csv.gz')
features = [f for f in train_dt.names if f not in ['HasDetections']]
for n in range(10):
    ftrl = Ftrl(nepochs=n+1)
    ftrl.fit(train_dt[:, features], train_dt[:, 'HasDetections'])
    print(log_loss(np.array(train_dt[trn_, 'HasDetections'])[:, 0], np.array(ftrl.predict(train_dt[trn_, features]))))
``` 
The output is 
```
0.6975873940617929
0.7004277294410224
0.7030339011892597
0.705290424565774
0.7072685897773024
0.7091474008277487
0.7108282513596036
0.7123130263929156
0.713890830846544
0.7151695514165213
```
my own version of FTRL trains correctly with the following output:
```
time_used:0:00:01.026606	epoch: 0   rows:10001	t_logloss:0.59638
time_used:0:00:01.715622	epoch: 1   rows:10001	t_logloss:0.52452
time_used:0:00:02.436984	epoch: 2   rows:10001	t_logloss:0.48113
time_used:0:00:03.158367	epoch: 3   rows:10001	t_logloss:0.44260
time_used:0:00:03.851369	epoch: 4   rows:10001	t_logloss:0.39633
time_used:0:00:04.553488	epoch: 5   rows:10001	t_logloss:0.38197
time_used:0:00:05.264179	epoch: 6   rows:10001	t_logloss:0.35380
time_used:0:00:05.973398	epoch: 7   rows:10001	t_logloss:0.32839
time_used:0:00:06.688121	epoch: 8   rows:10001	t_logloss:0.32057
time_used:0:00:07.394217	epoch: 9   rows:10001	t_logloss:0.29917
```
- Your environment?
I'm on ubuntu 16.04, clang+llvm-7.0.0-x86_64-linux-gnu-ubuntu-16.04, python 3.6, 
datatable is compiled from source.

let me know if you need more.

I guess I'm missing something but could not find anything in the unit tests.

Thanks for your help.

P.S. : make test results and the dataset I use are attached.
[datatable_make_test_results.txt](https://github.com/h2oai/datatable/files/2702636/datatable_make_test_results.txt)
[dt_ftrl_test_set.csv.gz](https://github.com/h2oai/datatable/files/2702639/dt_ftrl_test_set.csv.gz)

oleksiyskononenko:
- take into account rowindices in column hashers, so that FTRL algo trains and predicts on correct data even if rowindex is not empty;
- add a relevant test.

Closes #1502

### DIFF
--- a/c/extras/dt_ftrl.cc
+++ b/c/extras/dt_ftrl.cc
@@ -54,8 +54,8 @@ Ftrl::Ftrl(FtrlParams params_in) :
 void Ftrl::fit(const DataTable* dt_X, const DataTable* dt_y) {
   define_features(dt_X->ncols);
 
-  if (!is_dt_valid(dt_model, params.d, 2)) create_model();
-  if (!is_dt_valid(dt_fi, nfeatures, 1)) create_fi();
+  is_dt_valid(dt_model, params.d, 2)? init_weights() : create_model();
+  is_dt_valid(dt_fi, nfeatures, 1)? init_fi() : create_fi();
 
   // Create column hashers.
   create_hashers(dt_X);
@@ -96,7 +96,8 @@ void Ftrl::fit(const DataTable* dt_X, const DataTable* dt_y) {
 dtptr Ftrl::predict(const DataTable* dt_X) {
   xassert(model_trained);
   define_features(dt_X->ncols);
-  if (is_dt_valid(dt_fi, nfeatures, 1)) create_fi();
+  init_weights();
+  is_dt_valid(dt_fi, nfeatures, 1)? init_fi() : create_fi();
 
   // Re-create hashers as stypes for training dataset and predictions
   // may be different
@@ -164,12 +165,13 @@ void Ftrl::create_model() {
   Column* col_z = Column::new_data_column(SType::FLOAT64, params.d);
   Column* col_n = Column::new_data_column(SType::FLOAT64, params.d);
   dt_model = dtptr(new DataTable({col_z, col_n}, model_colnames));
-  init_weights();
+  w = doubleptr(new double[params.d]());
   reset_model();
 }
 
 
 void Ftrl::reset_model() {
+  init_weights();
   if (z == nullptr || n == nullptr) return;
   std::memset(z, 0, params.d * sizeof(double));
   std::memset(n, 0, params.d * sizeof(double));
@@ -181,14 +183,12 @@ void Ftrl::init_weights() {
   if (dt_model == nullptr) return;
   z = static_cast<double*>(dt_model->columns[0]->data_w());
   n = static_cast<double*>(dt_model->columns[1]->data_w());
-  w = doubleptr(new double[params.d]());
 }
 
 
 void Ftrl::create_fi() {
   Column* col_fi = Column::new_data_column(SType::FLOAT64, nfeatures);
   dt_fi = dtptr(new DataTable({col_fi}, {"feature_importance"}));
-  init_fi();
   reset_fi();
 }
 
@@ -200,6 +200,7 @@ void Ftrl::init_fi() {
 
 
 void Ftrl::reset_fi() {
+  init_fi();
   if (fi == nullptr) return;
   std::memset(fi, 0, nfeatures * sizeof(double));
 }
@@ -411,6 +412,7 @@ size_t Ftrl::get_nepochs() {
 void Ftrl::set_model(DataTable* dt_model_in) {
   dt_model = dtptr(dt_model_in->copy());
   set_d(dt_model->nrows);
+  w = doubleptr(new double[params.d]());
   init_weights();
   ncols = 0;
   nfeatures = 0;


c/extras/dt_ftrl.cc
```diff
@@ -64,33 +64,28 @@ void Ftrl::fit(const DataTable* dt_X, const DataTable* dt_y) {
   auto c_y = static_cast<BoolColumn*>(dt_y->columns[0]);
   auto d_y = c_y->elements_r();
 
-  // Do training for `nepochs`.
-  for (size_t i = 0; i < params.nepochs; ++i) {
-    double total_loss = 0;
+  const RowIndex ri_X = dt_X->rowindex;
```
oleksiyskononenko: @st-pasha, from what I understand, rowindex is deprecated for `DataTable`, right? Then which column rowindex one should use here and also for `predict`?

tests/extras/test_ftrl.py
```diff
@@ -606,6 +606,26 @@ def test_ftrl_fit_predict_from_setters():
...
 
+def test_ftrl_fit_predict_view():
+    ft = Ftrl(d=100)
+    df_train = dt.Frame([random.random() for _ in range(ft.d)])
+    df_target = dt.Frame([bool(random.getrandbits(1)) for _ in range(ft.d)])
+    rows = range(ft.d//2, ft.d)
+
+    ft.fit(df_train[rows,:], df_target[rows,:])
+    predictions = ft.predict(df_train[rows,:])
+    model = dt.Frame(ft.model.to_dict())
```
oleksiyskononenko: @st-pasha, this is something I don't understand. If I just do `model = dt.Frame(ft.model)`, then `model` is still a reference to `ft.model`. Why is it so?

st-pasha: `dt.Frame(DT)` creates a shallow copy of `DT`. This is equivalent to `DT.copy()`. What makes you think it doesn't work?

c/extras/dt_ftrl.cc
st-pasha: This is correct. DataTable's rowindex will get removed in the near future, and should not be used. Each column may have its own rowindex. You could detect and process specially the case when all columns have the same rowindex (it is quite common), but, in general, different columns are allowed to have different rowindices.

c/extras/dt_ftrl.cc
```diff
@@ -64,33 +64,28 @@ void Ftrl::fit(const DataTable* dt_X, const DataTable* dt_y) {
...
-
-        double loss = logloss(p, y);
-        #pragma omp atomic update
-        total_loss += loss;
-        if ((j+1) % REPORT_FREQUENCY == 0) {
-          printf("Training epoch: %zu\tRow: %zu\tPrediction: %f\t"
-                 "Current loss: %f\tAverage loss: %f\n",
-                 i, j+1, p, loss, total_loss / (j+1));
+      ri_X.iterate(ith, dt_X->nrows, nth,
+        [&](size_t i, size_t j) {
```
st-pasha: When iterating over a RowIndex you should also be prepared to encounter `j == RowIndex::NA` indicating that the value in that column is NA.

tests/extras/test_ftrl.py
oleksiyskononenko: Ah, I see. I don't say it doesn't work. I just didn't expect `dt.Frame(DT)` to return a shallow copy. What is the best way to make a deep copy then?

c/extras/dt_ftrl.cc
oleksiyskononenko: I see. So when we do something like `df[r1:r2,:]`, all columns will have the same rowindex, but if, for instance, we sort by two columns it may not be the case?

st-pasha: When sorting by 2 columns the rowindex will also be the same.
You can get different rowindices by: cbinding 2 frames that have different rowindices; selecting from a Frame some regular columns + some computed columns, e.g. `DT[:100, [f.A, f.A+f.B]]`; performing a join; etc.

oleksiyskononenko: I see, if rowindices can be different for different columns, iterating over just one will not work...

tests/extras/test_ftrl.py
st-pasha: I believe `copy.deepcopy()` just pickles and then unpickles the object. The more important question is why do you need a deep copy? The datatable should work correctly even with shallow copies.

oleksiyskononenko: Because I need a copy of the model here, otherwise if I simply do

```
model = dt.Frame(ft.model)
ft.reset
```

`model` is reset also

c/extras/dt_ftrl.cc
st-pasha: `i`th row of the result is the `j`th row of the source. Indices `i` are always consecutive (at least in single-threaded mode): 0, 1, 2, 3, ... Indices `j` can be anything, including `-1` (which is NA) indicating that there is no source row for that `i`th place.

This can happen for example when (1) the number of rows of a frame is increased; (2) several frames with different number of rows are force-cbound into a single frame; (3) an outer join is performed but some rows are not matched in either of the joined frames.

tests/extras/test_ftrl.py
st-pasha: I'll need to review that code, but it seems like there is a bug there. I would expect that simple `model = ft.model; ft.reset()` (without any copying) would preserve the variable `model`.

oleksiyskononenko: If I just do (for `d=4`, i.e. model size is 4x2)

```
    model = ft.model
    print("model: ", model.to_list())
    ft.reset()
    print("model after reset: ", model.to_list())
```

this is what I get

```
model:  [[-0.5, 0.0, 0.0, 0.5], [0.25, 0.0, 0.0, 0.25]]
model after reset:  [[0.0, 0.0, 0.0, 0.0], [0.0, 0.0, 0.0, 0.0]]
```

To me it looks consistent, because the model is returned as a shallow copy from FTRL: `return dt_model->copy();`. If you think there is a bug in my code, or somewhere else, please let me know. Thanks!

st-pasha: I've looked into this a little, and I think there could be a problem with pointers `z` and `n`. Even though they are acquired as `column->data_w()` pointers, they are kept for an arbitrary long period of time. During that time the pointers may become invalid (in particular, they may become invalid for writing if you create a copy of the Frame). It might be better to re-acquire these pointers at the beginning of each python call that needs them.

st-pasha: Are you sure? Here's what I'm getting:
```
>>> import random
>>> import datatable as dt
>>> from datatable.models import Ftrl
>>> ft = Ftrl(d=4)
>>> df_train = dt.Frame([random.random() for _ in range(ft.d)])
>>> df_target = dt.Frame([bool(random.getrandbits(1)) for _ in range(ft.d)])
>>> ft.fit(df_train, df_target)
>>> model = ft.model
>>> model.to_list()
[[0.9553244374605763, 0.0, 0.0, 0.0], [1.0001815001713965, 0.0, 0.0, 0.0]]
>>> ft.reset()
>>> model.to_list()
[[0.9553244374605763, 0.0, 0.0, 0.0], [1.0001815001713965, 0.0, 0.0, 0.0]]
>>> print(ft.model)
None
```

oleksiyskononenko: I've recompiled datatable from scratch and it seems like this problem has been fixed. I will update tests. Thanks!

Actually there was also another problem when I just did `df_train_range = dt.Frame(df_train[rows,:])`, it kind of inherited rowindex and `df_train_range` was still a view. That is why I had to do `df_train_range = dt.Frame(df_train[rows,:].to_list())`. Is it something we expect? If yes, then what is the best way to create a real frame based on the view?

st-pasha: for any frame you can write `frame.materialize()`, and it will be converted from "view" into a "real" data frame, in-place.

PR #1509, opened 2018-12-28
Fix split_into_nhot() when applied to a view

### LINKED ISSUE #1507: split_into_nhot() doesn't work properly on views
`split_into_nhot()` works just fine on the entire frame and views that start from zero row
```
>>> import datatable as dt
>>> from datatable import split_into_nhot
>>> df = dt.Frame(["a", "b"])
>>> split_into_nhot(df)
      a   b
---  --  --
 0    1   0
 1    0   1
>>> split_into_nhot(df[0,:])
      a
---  --
 0    1
```

If a view doesn't start at the zero row, it returns wrong results
```
>>> split_into_nhot(df[1,:])
      a
---  --
 0    1
```

Expected output in this case is
```
>>> split_into_nhot(df[1,:])
      b
---  --
 0    1
```

Need this to be fixed to finalize #1499.

st-pasha:
Function `dt.split_into_nhot(frame)` now works correctly when `frame` is a view.

Closes #1507

### DIFF
--- a/tests/munging/test_str.py
+++ b/tests/munging/test_str.py
@@ -126,3 +126,14 @@ def test_split_into_nhot_long(seed, st):
     assert set(f1.names) == set(fr.names)
     f1 = f1[..., fr.names]
     assert f1.to_list() == fr.to_list()
+
+
+def test_split_into_nhot_view():
+  f0 = dt.Frame(A=["cat,dog,mouse", "mouse", None, "dog, cat"])
+  f1 = dt.split_into_nhot(f0[::-1, :])
+  f2 = dt.split_into_nhot(f0[3, :])
+  assert set(f1.names) == {"cat", "dog", "mouse"}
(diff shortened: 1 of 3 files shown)


tests/munging/test_str.py
```diff
@@ -126,3 +126,14 @@ def test_split_into_nhot_long(seed, st):
     assert set(f1.names) == set(fr.names)
     f1 = f1[..., fr.names]
     assert f1.to_list() == fr.to_list()
+
+
+def test_split_into_nhot_view():
+  f0 = dt.Frame(A=["cat,dog,mouse", "mouse", None, "dog, cat"])
+  f1 = dt.split_into_nhot(f0[::-1, :])
```
oleksiyskononenko: Does `f0[::-1, :]` mean all the rows taken in a reverse order?

st-pasha: yes, same rules as with a regular python array

PR #1522, opened 2019-01-09
Add support for multinomial regression and linear regression for numerical columns

### LINKED ISSUE #1499: Support for multinomial logistic regression in FTRL
For the moment FTRL model we implemented can only do binomial regression, i.e. two possible outcomes. This should be generalized to multinomial case, i.e. classification for multiclass problems. 

Based on the existing capabilities, we can implement [One-vs.-rest](https://en.wikipedia.org/wiki/Multiclass_classification#One-vs.-rest) strategy.

### LINKED ISSUE #1518: Support regression for numeric targets in FTRL model
For the moment, a target column type is restricted to `bool`, as we only do binomial regression. When a target column type is numeric, we should also do regression, i.e. `fit()` on and `predict()` for a continuous range of targets.

oleksiyskononenko:
Depending on the target column type do the following:
- for numeric columns do regression, for which `identity()` link function is used by both `fit()` and `predict` methods;
- for boolean columns do binomial regression, for which`sigmoid()` link function is used by both `fit()` and `predict()` methods;
- for string columns do multinomial regression, which is implemented as a set of binomial classifiers. As a consequence, `sigmoid()` link function is used by `fit()` and `softmax()` function is applied by the `predict()` method.

Add tests for most of the use-cases, refactor.

Closes #1499
Closes #1518


oleksiyskononenko:
@st-pasha are you sure we should call it "linear regression" here?

st-pasha:
I thought you picked the terminology from some paper?
If not, I'd avoid the term "regression" completely: FTRL for binary target, categorical target, and continuous target variable.

oleksiyskononenko:
Well, for the binomial case it is indeed called "logistic regression" in the original paper, there is no paper that discussed multinomial and continuous cases. So I would say it is a regression, but not sure we can call it linear for the case of continuous targets. We need to reformulate it for the docs.

PR #1524, opened 2019-01-09
Implement updating values within string columns

### LINKED ISSUE #1523: NotImplementedError: StringColumn::replace_values() not implemented
After refactor - this feature was lost, it seems. Needed for DAI.

st-pasha:
This PR makes it possible to modify a string column in-place, via expression `DT[i, j] = ...`. The expression itself was implemented long time ago; but now it also works when applied to string columns.

Closes #1523

### DIFF
--- a/c/column_string.cc
+++ b/c/column_string.cc
@@ -326,25 +326,58 @@ void StringColumn<T>::replace_values(
     RowIndex replace_at, const Column* replace_with)
 {
   reify();
-  if (!replace_with) {
+  Column* rescol = nullptr;
+
+  if (replace_with && replace_with->stype() != stype()){
+    replace_with = replace_with->cast(stype());
+  }
+  // This could be nullptr too
+  auto repl_col = static_cast<const StringColumn<T>*>(replace_with);
+
+  if (!replace_with || replace_with->nrows == 1) {
+    CString repl_value;  // Default constructor creates an NA string
+    if (replace_with) {
+      T off0 = repl_col->offsets()[0];
+      if (!ISNA<T>(off0)) {
+        repl_value = CString(repl_col->strdata(), static_cast<int64_t>(off0));
+      }
+    }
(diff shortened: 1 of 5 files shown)


c/column_string.cc
```diff
@@ -326,25 +326,58 @@ void StringColumn<T>::replace_values(
...
   }
-  throw NotImplError() << "StringColumn::replace_values() not implemented";
+
+  xassert(rescol);
+  StringColumn<T>* scol = static_cast<StringColumn<T>*>(rescol);
+  std::swap(mbuf, scol->mbuf);
+  std::swap(strbuf, scol->strbuf);
+  delete rescol;
+  if (stats) stats->reset();
 }
```
oleksiyskononenko: So you have to re-create a column in this case, because it is not FW?

st-pasha: Yes, because all strings are stored contiguously end-to-end.
If you wanted to replace a single string for example, you could have copied the initial part of the string buffer, then the new string, then the remaining string buffer. But you'd still need to adjust all offsets. 
So basically any string manipulation requires modifying the entire string buffer and offsets buffer.

Which means that doing many small edits on a string column is very expensive. Conceivably, we could create a special type of "EditableStringColumn" which could for example use Rope data structure, or even a simple Gap buffer (the way most text editors are implemented). But that's a non-trivial amount of work.

oleksiyskononenko: Yep, that's what I thought. So there is always a trade-off between how effective we are at storing vs manipulating the strings.

st-pasha: Our storage schema is dictated by the fact that we want to have the same storage mechanism in memory as on disk. Btw Apache Arrow stores string columns in almost the same way (with the exception of NA strings).

oleksiyskononenko: I see, and may be most of the times we need these strings in a r/o mode, i.e. replacements are rare.

PR #1532, opened 2019-01-10
Do not allow `None` to be assigned to float FTRL parameters

oleksiyskononenko:
Otherwise Python`None` gets converted to C++ `NaN`, and makes no sense for the FTRL algo.


st-pasha:
wow, that's a lot of commits for a 16-line change :)

oleksiyskononenko:
This what Eclipse is doing when you switch from one branch to another, and then merge `master` onto the new branch :)

oleksiyskononenko:
Thanks. I didn't notice `math.nan` can also come through.

st-pasha:
Maybe add a test then that it works as expected?

oleksiyskononenko:
Yes, I already tested it manually, adding these tests to `extras/test_ftrl.py`. Btw, do you know a good word for "either positive or zero"? What I mean is that I would prefer to replace this error message `ValueError: Value cannot be negative: nan` with something like `ValueError: Value should be some_word_that_is_equivalent_to_either_positive_or_zero: nan`.

st-pasha:
You could try:
```
ValueError: Invalid value for parameter `alpha` of FTRL model: nan
ValueError: Parameter `alpha` of FTRL model should be positive or zero, got nan
```
There is also word "nonnegative", but I find it awkward.

st-pasha:
Are you going to forbid infinities too? (`math.inf`)

oleksiyskononenko:
`nonnegative` may work also, just not sure if it is the word that is commonly used in this context.
What would be also good to do, is to have this message consistent with this one in `obj.cc`.

```
Error _obj::error_manager::error_int_negative(PyObject*) const {
  return ValueError() << "Integer value cannot be negative";
}
```

What do you think? I am inclining to "greater than or equal to zero". But if you prefer I can also do "should be positive or zero".

oleksiyskononenko:
Yes, probably need to forbid infinities too...

tests/extras/test_ftrl.py
```diff
@@ -298,32 +299,128 @@ def test_ftrl_set_wrong_inter_type():
 # Test getters and setters for wrong values of individual FTRL parameters
 #-------------------------------------------------------------------------------
 
-def test_ftrl_set_wrong_alpha_value():
+def test_ftrl_set_zero_alpha_value():
```
st-pasha: You can simply parametrize this:
```python
@pytest.mark.parametrize('x', [0.0, math.nan, math.inf, None])
def test_ftrl_set_bad_alpha_value(x):
    ft = Ftrl()
    with pytest.raises(ValueError) as e:
        ft.alpha = x
    assert ("Value should be positive: %s" % str(x) == str(e.value))
```

oleksiyskononenko: yes, forgot to re-factor it at the very end...

PR #1553, opened 2019-01-14
Traverse exemplars in a quasi-random order in ND aggregator

oleksiyskononenko:
When exemplars are traversed in the order they appear in the `exemplars` vector, first exemplars will always have a priority over the last ones, hence, first exemplars "absorb" most of the members. In this case one can not distinguish real outliers with the exemplars (the last ones) that just had no chances to get some members.

One of the ways to fix it is to traverse exemplars in a random order, so that each exemplar gets the "same" chances to "absorb" a particular member. Here we use modular arithmetic to traverse exemplars in quasi-random order, as this approach seems to have negligible overhead, when comparing to `std::random_shuffle`, and we do not really need to be fully random, but just different enough to not inflate some particular exemplars.

### DIFF
--- a/c/extras/aggregator.cc
+++ b/c/extras/aggregator.cc
@@ -578,7 +578,7 @@ void Aggregator::group_nd(const dtptr& dt, dtptr& dt_members) {
   // Exemplar counter, if doesn't match thread local value, it means
   // some new exemplars were added (and may be even `delta` was adjusted)
   // meanwhile, so restart is needed for the `test_member` procedure.
-  size_t ecounter;
+  size_t ecounter = 0;
 
   #pragma omp parallel num_threads(nth0)
   {
@@ -588,6 +588,7 @@ void Aggregator::group_nd(const dtptr& dt, dtptr& dt_members) {
     double distance;
     doubleptr member = doubleptr(new double[ndims]);
     size_t ecounter_local;
+    std::vector<size_t> sample;
 
     try {
       // Main loop over all the rows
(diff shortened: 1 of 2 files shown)


c/extras/aggregator.cc
```diff
@@ -641,6 +667,25 @@ void Aggregator::group_nd(const dtptr& dt, dtptr& dt_members) {
 }
 
 
+void Aggregator::fill_coprimes(size_t n, std::vector<size_t>& coprimes) {
+  coprimes.clear();
+  for (size_t i = 1; i <= n; ++i) {
+    if (gcd(i, n) == 1) coprimes.push_back(i);
```
st-pasha: Eratosthenes Sieve is a much better approach to finding coprimes:
- create a bitmask of length `n`, filled with 0s;
- starting with `i = 2`:
  - if `mask[i]` is 1 then continue;
  - if `n` is divisible by `i`, then set all elements of the bitmask that are multiples of `i` to 1;
- continue until `i*i < n`.
In the end, all elements of bitmask that are 0s are coprime to `n`.

oleksiyskononenko: Yes, sure, current approach can be optimized, I was more concerned about the final `members_count` distribution for the moment. For the aggregation purpose we can even stick to the list of prime numbers that are smaller than the typical `nd_max_bins`, i.e. ~1000, and then just use these prime numbers as the "seed".

oleksiyskononenko: Actually, it seems that we can just use any prime number that is larger than `nd_max_bins`.

oleksiyskononenko: Yes, sorry. I probably kept in mind the approach of `[n / 2; n]` for the range of prime numbers.

oleksiyskononenko: Ok, I've changed it to Eratosthenes sieve, stopping at `i == n / 2`. Could probably optimized even further, but still good for our purpose.

PR #1563, opened 2019-01-16
Fix a bug in true division of integers

### LINKED ISSUE #1562: Wrong result during division of int8 / int16
```python
from datatable import *
DT = Frame(A=[-8], B=[1677])
DT[:, f.A / f.B]
```
produces
```
           C0
---  --------
 0   -1.00477

[1 row x 1 column]
```

st-pasha:
The issue was that the logic for correcting integer division (offsetting by -1 when dividing a negative number by a positive) was incorrectly applied for true division as well.

Closes #1562

### DIFF
--- a/c/expr/binaryop.cc
+++ b/c/expr/binaryop.cc
@@ -195,7 +195,7 @@ inline static VT op_div(LT x, RT y) {
   VT vx = static_cast<VT>(x);
   VT vy = static_cast<VT>(y);
   VT res = vx / vy;
-  if ((vx < 0) != (vy < 0) && vx != res * vy) {
+  if (std::is_integral<VT>::value && (vx < 0) != (vy < 0) && vx != res * vy) {
     --res;
   }
   return res;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - A string column produced from a partial join now materializes correctly
   (#1556).
 
+- Fixed incorrect result during "true division" of integer columns, when one
+  of the values was negative and the other positive (#1562).
+
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


c/expr/binaryop.cc
```diff
@@ -195,7 +195,7 @@ inline static VT op_div(LT x, RT y) {
   VT vx = static_cast<VT>(x);
   VT vy = static_cast<VT>(y);
   VT res = vx / vy;
-  if ((vx < 0) != (vy < 0) && vx != res * vy) {
+  if (std::is_integral<VT>::value && (vx < 0) != (vy < 0) && vx != res * vy) {
```
oleksiyskononenko: - What is the reason we need to decrement `res` here?

- Am I correct that `vx` and `vy` could be integer, and `res` float? If yes, then are you sure the condition `vx != res * vy` won't always be true due to floating point arithmetics on the right hand side?

st-pasha: This accounts for the difference in division operator between python in C. In C, x/y truncates the result towards 0, whereas in python x/y returns the floor. Thus, in C `(-1)/2 == 0`, while in python `(-1)//2 == -1`.

st-pasha: If `res` is float, then `std::is_integral<VT>::value` will be false, and the if condition will be ignored.

oleksiyskononenko: Ah, yes. With this change it will not pass through.

Btw, what are the cases when we divide integer over integer without casting them to float?

oleksiyskononenko: Then what is the case you provided the above example for? I mean when we may end up with integers being divided as `(-1) / 2`?

st-pasha: In python integer division is `//`. So, `(-1) // 2 == -1`, while `(-1) / 2 == -0.5`.

PR #1564, opened 2019-01-17
Add tests for setting column to NA

### LINKED ISSUE #1555: Setting a column's values to NA does not preserve stype
Whenever a column has its values set to `None` its stype is converted to `bool8`. I would expect the original `stype` to be preserved
```
>>> X = dt.Frame(a=list("abc"))
>>> X.stypes
(stype.str32,)
>>> X[:, 'a'] = None
>>> X.stypes
(stype.bool8,)
```

st-pasha:
When a column is set to `None`, its stype will now be preserved.
The necessary change was introduced in #1561, this PR merely adds tests.

Closes #1555

### DIFF
--- a/tests/munging/test_assign.py
+++ b/tests/munging/test_assign.py
@@ -1,10 +1,28 @@
 #!/usr/bin/env python
-# © H2O.ai 2018; -*- encoding: utf-8 -*-
-#   This Source Code Form is subject to the terms of the Mozilla Public
-#   License, v. 2.0. If a copy of the MPL was not distributed with this
-#   file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# -*- coding: utf-8 -*-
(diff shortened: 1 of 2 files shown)


tests/munging/test_assign.py
```diff
@@ -99,3 +117,35 @@ def test_assign_empty_frame():
...
+    DT = dt.Frame(A=range(5))
+    DT[:, f.A] = None
+    DT.internal.check()
+    assert DT.stypes == (dt.int32,)
+    assert DT.to_list() == [[None] * 5]
+
+
+def test_assign_none_new():
+    DT = dt.Frame(A=range(5))
+    DT[:, "B"] = None
```
oleksiyskononenko: Would be nice if we had a similar way of adding a new column in C++ :)

PR #1583, opened 2019-01-23
FTRL API adjustments

oleksiyskononenko:
- rename `d` to `nbins`;
- rename `inter` to `interactions`;
- add a `feature_name` column to `feature_importances` frame, that will reflect on the actual feature name for which we calculated `feature_importance`;
- adjust and add more tests;
- adjust documentation.

### DIFF
--- a/c/extras/dt_ftrl.cc
+++ b/c/extras/dt_ftrl.cc
@@ -131,7 +131,7 @@ void Ftrl::reset_fi() {
 
 void Ftrl::define_features(size_t ncols_in) {
   ncols = ncols_in;
-  size_t n_inter_features = (params.inter)? ncols * (ncols - 1) / 2 : 0;
+  size_t n_inter_features = (params.interactions)? ncols * (ncols - 1) / 2 : 0;
   nfeatures = ncols + n_inter_features;
 }
 
@@ -179,7 +179,7 @@ hashptr Ftrl::create_colhasher(const Column* col) {
 
 
 /*
-*  Hash each element of the datatable row, do feature interaction if requested.
+*  Hash each element of the datatable row, do feature interactionsaction if requested.
 */
 void Ftrl::hash_row(uint64ptr& x, size_t row) {
   for (size_t i = 0; i < ncols; ++i) {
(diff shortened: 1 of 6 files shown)


c/extras/dt_ftrl.cc
```diff
@@ -179,24 +179,24 @@ hashptr Ftrl::create_colhasher(const Column* col) {
 
 
 /*
-*  Hash each element of the datatable row, do feature interaction if requested.
+*  Hash each element of the datatable row, do feature interactionsaction if requested.
```
st-pasha:
```suggestion
*  Hash each element of the datatable row, do feature interactions if requested.
```

c/extras/py_ftrl.cc
```diff
@@ -334,6 +339,33 @@ void Ftrl::fit(const PKArgs& args) {
     default:             throw TypeError() << "Cannot predict for a column "
                                            << "of type `" << stype_y << "`";
   }
+
+  if (feature_names == nullptr) {
+    size_t nfeatures = (*dtft)[0]->get_nfeatures();
+    size_t ncols = dt_X->ncols;
+    std::vector<std::string> column_names = dt_X->get_names();
```
st-pasha:
```suggestion
    const std::vector<std::string>& column_names = dt_X->get_names();
```
to avoid unnecessary copy

c/extras/py_ftrl.cc
```diff
@@ -334,6 +339,33 @@ void Ftrl::fit(const PKArgs& args) {
...
+
+    dt::fixed_height_string_col scol(nfeatures);
+    dt::fixed_height_string_col::buffer sb(scol);
+    sb.commit_and_start_new_chunk(0);
+    for (auto feature_name : column_names) {
```
st-pasha:
```suggestion
    for (const auto& feature_name : column_names) {
```

c/extras/dt_ftrl.cc
oleksiyskononenko: Sure, thanks, thought I fixed all these issues.

c/extras/py_ftrl.cc
oleksiyskononenko: Yep, also missed.

oleksiyskononenko:
Re-built docs on readthedocs.

PR #1588, opened 2019-01-24
slice-valued i expression, combined with a groupby

### LINKED ISSUE #1585: slice_in class should work properly in the presence of a groupby
The `slice_in` node extracts a slice of rows from the Frame. In the presence of a groupby, it should extract a slice of rows from each group. The semantics of a slice is to produce an empty set if the slice boundaries are outside of the range of the sequence. Thus, if a slice produces 0 results for some group, that group will simply be omitted from the result.

st-pasha:
This PR allows the syntax such as
```
DT[:5, :, by("id")]
```
to select the first 5 rows for each group when the frame is grouped by the "id" column.

Slices with positive, negative, and zero step are supported.

Closes #1585

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -115,7 +115,9 @@ def assert_equals(frame1, frame2):
         assert frame1.stypes == frame2.stypes, (
             "The left frame has stypes %r, while the right has stypes %r"
             % (frame1.stypes, frame2.stypes))
-        assert frame1.to_list() == frame2.to_list()
+        assert frame1.to_list() == frame2.to_list(), (
+            "The frames have different data:\n"
+            "LHS = %r\nRHS = %r" % (frame1.to_list(), frame2.to_list()))
     else:
         assert frame1.shape == frame2.shape
         assert same_iterables(frame1.names, frame2.names)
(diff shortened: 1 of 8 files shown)


tests/__init__.py
```diff
@@ -115,7 +115,9 @@ def assert_equals(frame1, frame2):
         assert frame1.stypes == frame2.stypes, (
             "The left frame has stypes %r, while the right has stypes %r"
             % (frame1.stypes, frame2.stypes))
-        assert frame1.to_list() == frame2.to_list()
+        assert frame1.to_list() == frame2.to_list(), (
+            "The frames have different data:\n"
+            "LHS = %r\nRHS = %r" % (frame1.to_list(), frame2.to_list()))
```
oleksiyskononenko: Won't this produce too much of output if frames are different? For example for some tests with big data. May be we could have a parameter that controls verbosity of this function?

oleksiyskononenko: Sure, but I mean the tests. I'm using this function for some moderate size data, guess I saw some of your large data tests, if test fails, this will produce a lot of noise :)

st-pasha: And once a test fails and produces a lot of noise, there will be an incentive to go and fix the reporting routine to output a more meaningful diff. This is what I mean by "solve problems only as they appear". The current change is still an improvement: showing all the data is better than not showing any.

PR #1595, opened 2019-01-25
Fix rbinding frames with negative step slice

### LINKED ISSUE #1594: Incorrect result after rbinding a view frame
```python
from datatable import *
DT = Frame(range(5))
DT = DT[::-1, :]
DT.nrows = 4
_ = DT.rbind(DT)
assert DT.to_list() == [[4, 3, 2, 1, 4, 3, 2, 1]]
```
However, the Frame contains data `[[4, 3, 2, 3, 4, 3, 2, 3]]`.

st-pasha:
- The negative step slice was incorrectly accounted for during frame materialization;
- Improved error printout for data mismatch in random attack script.

Closes #1594

### DIFF
--- a/c/column_fw.cc
+++ b/c/column_fw.cc
@@ -141,7 +141,7 @@ void FwColumn<T>::reify() {
   // If the rowindex is absent, then the column is already materialized.
   if (!ri) return;
   bool simple_slice = ri.isslice() && ri.slice_step() == 1;
-  bool ascending    = ri.isslice() && ri.slice_step() > 0;
+  bool ascending = ri.isslice() && static_cast<int64_t>(ri.slice_step()) > 0;
 
   size_t elemsize = sizeof(T);
   size_t newsize = elemsize * nrows;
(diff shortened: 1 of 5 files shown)


c/column_fw.cc
```diff
@@ -141,7 +141,7 @@ void FwColumn<T>::reify() {
   // If the rowindex is absent, then the column is already materialized.
   if (!ri) return;
   bool simple_slice = ri.isslice() && ri.slice_step() == 1;
-  bool ascending    = ri.isslice() && ri.slice_step() > 0;
+  bool ascending = ri.isslice() && static_cast<int64_t>(ri.slice_step()) > 0;
```
oleksiyskononenko: When I look at return type of `ri.slice_step()`, it is `size_t`. Then what is your definition of a negative slice step?

st-pasha: For example `step = 18446744073709551615 = size_t(-1)` is considered negative, because for any reasonable starting index `start`, we will have `start + step < start` using `size_t` arithmetics. I can test whether the step is negative by casting into `int64_t` and checking the sign of that.

oleksiyskononenko: I see, so you're using `size_t` arithmetics everywhere for the slices, and basically `std::numeric_limits<size_t>::max() - i` will result in `-(i+1)` step?

st-pasha: Yes, though I don't do subtraction in code, just casting between `int64_t` and `size_t`

PR #1613, opened 2019-01-29
Implement multi-iter protocol for DataTable

st-pasha:
In #1188, each columns in the Frame may potentially have its own distinct rowindex. 
For row-based iteration over a frame, we need a convenient method to perform such iteration, given that there are multiple columns each with its own rowindex.
This PR implements such functionality.

### DIFF
--- a/c/datatable.cc
+++ b/c/datatable.cc
@@ -145,25 +145,25 @@ void DataTable::delete_all() {
 }
 
 
+
 // Split all columns into groups, by their `RowIndex`es
-static std::pair<std::vector<RowIndex>, std::vector<std::vector<size_t>>>
-_split_columns_by_rowindices(const DataTable* dt)
-{
-  std::vector<RowIndex> rowindices;
-  std::vector<std::vector<size_t>> colindices;
-  for (size_t i = 0; i < dt->ncols; ++i) {
-    RowIndex r = dt->columns[i]->rowindex();
-    size_t j = 0;
-    for (; j < rowindices.size(); ++j) {
-      if (rowindices[j] == r) break;
+std::vector<RowColIndex> DataTable::split_columns_by_rowindices() const {
+  std::vector<RowColIndex> res;
+  for (size_t i = 0; i < ncols; ++i) {
+    RowIndex r = columns[i]->rowindex();
+    bool found = false;
+    for (auto& item : res) {
+      if (item.rowindex == r) {
+        found = true;
+        item.colindices.push_back(i);
+        break;
+      }
     }
-    if (j == rowindices.size()) {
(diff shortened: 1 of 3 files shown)


c/datatable.cc
```diff
@@ -145,25 +145,25 @@ void DataTable::delete_all() {
...
-_split_columns_by_rowindices(const DataTable* dt)
-{
-  std::vector<RowIndex> rowindices;
-  std::vector<std::vector<size_t>> colindices;
-  for (size_t i = 0; i < dt->ncols; ++i) {
-    RowIndex r = dt->columns[i]->rowindex();
-    size_t j = 0;
-    for (; j < rowindices.size(); ++j) {
-      if (rowindices[j] == r) break;
+std::vector<RowColIndex> DataTable::split_columns_by_rowindices() const {
```
oleksiyskononenko: May be a better name for this method is `group_columns_by_rowindices()`?

c/datatable.h
```diff
@@ -183,4 +209,30 @@ RowIndex natural_join(const DataTable* xdt, const DataTable* jdt);
 
 //==============================================================================
 
+template <void (*F1)(size_t i, size_t j),
+          void (*FN)(size_t i, const intvec& js)>
+void DataTable::iterate_rows(size_t row0, size_t row1) {
+  if (row1 == size_t(-1)) {
+    row1 = nrows;
+  }
+  std::vector<RowColIndex> rcs = split_columns_by_rowindices();
```
oleksiyskononenko: Btw, what about storing this `rcs` as a property in `DataTable`, and invalidate it when there are some relevant changes to the columns? Then, if several passes over the data is needed (for example when we do `nepochs > 1` in FTRL), `rcs` won't have to be calculated multiple times.

c/datatable.cc
st-pasha: sure

c/datatable.h
oleksiyskononenko: Yes, I can imagine there are many places in the code where RowIndex could be affected. By the way, when we had `RowIndex` as a `DataTable` property, how this one was changed when one did `column->reify()`? 

I guess, `rcs` could also be an optional parameter to `iterate_rows`, if provided it should be used as is and not re-calculated. Actually, for FTRL I store `RowIndex` for each column within a corresponding hasher object, and then apply it while hashing. Do you think this is better/worse in terms of performance comparing this this new iterator?

st-pasha: With your approach there is a cost of one virtual function call per column. With my approach there is a cost of copying the value `j` into once per column. Alternatively this could be implemented as double lookup.

PR #1625, opened 2019-01-31
max/min for an empty column now returns None

### LINKED ISSUE #1624: max/min of an empty column should return None
```
>>> f = dt.Frame(a=[float("nan"), 3], b=['a', 'b'])
>>> f
      a  b 
---  --  --
 0       a 
 1    3  b 

[2 rows x 2 columns]

>>> g = f[:, {'c': dt.max(dt.f['a'])}, dt.by('b')]
>>> g
     b      c
---  --  ----
 0   a   -inf
 1   b      3

[2 rows x 2 columns]
```

`[0, 'c']` should be `NA`.

st-pasha:
This PR changes the behavior of functions `max()` and `min()` when they are applied to empty or all-NA columns. Previously those functions returned `-Inf` and `+Inf` respectively (or, for integer columns, `-MAX` and `+MAX`). Now these functions will return `None`.

The reasons for the change are discussed in #1624, but in summary they are:
- in mathematics max and min of an empty set do not exist, only supremum and infimum;
- for integers returning a large integer looks like a bug, and we prefer the behavior of floats to match the behavior of integers;
- numpy returns nan for max/min of array of nans (with a warning);
- R returns NA for max/min of a vector of NAs (however, it returns ±∞ if option `na.rm=TRUE` is used).

Closes #1624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,6 +191,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   the frame in-place, so it was confusing to both update in-place and return the
   original frame (#1610).
 
+- `min()` / `max()` over an empty or all-NA column now returns `None` instead of
+  +Inf / -Inf respectively (#1624).
+
 
 ### Deprecated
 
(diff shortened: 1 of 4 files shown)


oleksiyskononenko:
Do we want to issue any warning when max/min is applied to an empty column?

st-pasha:
Do we? NA is a perfectly valid value in `datatable`, it is quite possible to get a column having only NA values in it, especially during grouping. Why is this a warning situation? If the user gets the warning, would they do anything different? Can anything bad happen if the user is not warned? If the user is ok with current behavior, how would they silence the warning (esp. when running in "warnings are errors" mode)?

PR #1629, opened 2019-02-01
#952 latest stable artifacts generation

abal5:
WIP for #952

### DIFF
--- a/ci/Jenkinsfile.groovy
+++ b/ci/Jenkinsfile.groovy
@@ -699,19 +699,19 @@ ansiColor('xterm') {
                                         s3cmd put -P release/dist/*.whl ${S3_URL_STABLE}/datatable-${versionText}/
                                         s3cmd put -P release/dist/*.tar.gz ${S3_URL_STABLE}/datatable-${versionText}/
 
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile.groovy
```diff
@@ -54,7 +54,9 @@ SOURCE_DIR = "/home/0xdiag"
 TARGET_DIR = "/tmp/pydatatable_large_data"
 BUCKET = 'h2o-release'
 STABLE_FOLDER = 'datatable/stable'
+LATEST_STABLE = 'datatable/latest_stable'
```
mmalohlava: what is different between stable and latest_stable?

jangorecki:
@abal5 could have those links mentioned in README?

st-pasha:
Nope, this doesn't work:

1. the wheels are inside the `isRelease()` section, which means they trigger not after every PR but only during the release (which so far happens once every few months);
2. the wheel name is not even recognized by pip as valid:
```
$  pip install https://h2o-release.s3.amazonaws.com/datatable/latest_stable/datatable-0.latest-p35-linux_x86_64.whl
datatable-0.latest-p35-linux_x86_64.whl is not a valid wheel filename.
```
3. I can specify the wheel name in the requirements file, but this is very unportable: the python version and the platform is hard-coded, and the process will break if either of those change.

Issue #1310, opened 2018-09-25
Switch back to the Apache-v2 license


st-pasha:
The absolute majority of Python packages are using Apache, MIT, BSD, or similar open licenses. It would be courteous to the broader Python community, and invite broader collaboration/contribution, if we did as well.

Historically, this project has been Apache from the [very first commit](https://github.com/h2oai/datatable/commit/83f4b1efb19e5ceb456f64b5dafb312691ec2af0). However, sometime before the public release, we switched to MPL-2 license. The idea was to have the same license as R data.table project (which at that time switched from GPL to MPL too). Unfortunately, we failed to grasp the primary difference between R and Python communities at that point: the majority of R packages are licensed as GPL, and within such environment, an MPL-licensed project can be integrated freely and will be seen as more open compared to others. On the contrary, within Python community, an MPL license is more restrictive and will be eyed with suspicion. In fact, MPL license creates a perfectly tangible barrier: ASF includes this license into the [Category B list](https://www.apache.org/legal/resolved.html#category-b) of software that can only be integrated in binary, but not in source code form.

Please, share your thoughts/comments.

mattdowle:
My last comment on the R-data.table project members list was :  

> I've spoken to Pasha today. Indeed my fears are realized. He believes that the only part of pydatatable that came from data.table is one file: fread. He believes all the rest he did himself. This is why fread is the only file which is kind of different to the rest in terms of the licensing at the top. Everything else in pydatatable has copyright of H2O. This means that because H2O is the copyright holder, H2O can change the license of pydatatable without asking any contributor to data.table. That is Pasha's view I believe.
Pasha - can you confirm?

This remains unanswered.  Can you confirm?

This came up recently here : h2oai/datatable#1232
The license change for data.table was extensive:  https://github.com/Rdatatable/data.table/pull/2456

Note that I explicitly left the door open to Apache in https://github.com/Rdatatable/data.table/pull/2456. But my problem with you is that you've been sneaky. You tricked the data.table contributors by rewriting data.table and licensing fread differently to the rest of it. You ignored all those discussions. You ignored the agreement. You're saying that https://github.com/Rdatatable/data.table/pull/2456 didn't matter, that it was a waste of time, and you don't need to discuss with the past contributors to data.table in R.

Otherwise, what was the point of https://github.com/Rdatatable/data.table/pull/2456?

st-pasha:
[...]

PR #1633, opened 2019-02-04
Documentation update

st-pasha:


docs/contrib.rst
```diff
@@ -0,0 +1,16 @@
+
+Contributing
+------------
+
+``datatable`` is an open source project released under the Mozilla Public Licence v2. Open Source projects live by their user and developer communities. We welcome and encourage your contributions of any kind!
```
oleksiyskononenko: So `datatable` is still MPL2, while FTRL (at least this is how we put it on the conference slides) can be considered as MIT?

docs/quick-start.rst
```diff
@@ -0,0 +1,308 @@
...
+    DT.to_pandas()
+    DT.to_numpy()
+
+A frame can also be converted into python native data structures: a dictionary,
+keyed by the column names; a list of columns, where each column is itself a
+list of values; or a list of rows, where each row is a tuple of values::
+
+    DT.to_dict()
+    DT.to_list()
+    DT.to_tuples()
```
oleksiyskononenko: Is there any reason as to why we create a list of tuples and not tuple of tuples?

st-pasha: Think about this from the types perspective: there is a type `List[T]` for a list of objects of type `T`; then there is also type `Tuple[T1, T2, ..., Tn]` for a tuple of objects of types `T1`, `T2`, ..., `Tn`. So it is natural to use lists for a sequence of same-typed elements of arbitrary length, and it is natural to use tuples for a fixed-length sequence of items of different types.

Though in python it is possible to put anything into a list, or to create a tuple of any length -- this is much less natural, and goes against principles of other typed languages.

It could be argued though that `DT.to_list()` should really return a tuple of columns instead of a list, but I'm not sure if it's worth it...

docs/contrib.rst
st-pasha: Yes, whatever that means. Since MPL is not a repository-level license, it only applies to those files that were explicitly marked as MPL.
We have a task #1310 to switch entirely to MIT, but I can't proceed with that without @arnocandel 's permission.

PR #1655, opened 2019-02-13
Create mechanism for defining python interface methods more robustly

st-pasha:
Adding new macro `ADD_METHOD(mm, method, args);`, which should eventually replace all the existing template functions `mm.add<...>()`. I hope that the new approach will be more robust than the current (in particular, less chances of compiler errors). In addition, it is conceptually simpler and produces smaller code footprint than the current approach. Also, the `args` variable can now be declared statically in a file, there is no need to define it in the class.

This PR only converts 2 methods `Frame.head()` and `Frame.tail()` to the new approach, as a "test-drive". A subsequent PR will take care to convert all the existing usages.

### DIFF
--- a/c/frame/py_frame.cc
+++ b/c/frame/py_frame.cc
@@ -30,6 +30,53 @@ namespace py {
 PyObject* Frame_Type = nullptr;
 
 
+
+//------------------------------------------------------------------------------
+// Head & tail
+//------------------------------------------------------------------------------
+
+static PKArgs fn_head(
+    1, 0, 0, false, false,
+    {"n"}, "head",
+R"(head(self, n=10)
+--
+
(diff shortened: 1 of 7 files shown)


oleksiyskononenko:
Yes, seems like there will be less problems with macros comparing to what we currently have. Several questions:

- will the old way of defining methods continue to work, or we need to rewrite all of the `mm.add` instances asap?

- are we going to change getters and setters also?

st-pasha:
The old way of defining `mm.add` still works, as evidenced by the fact that everything compiles and all tests are green. However, I'd like to remove those methods eventually, after transitioning the code.

We should probably do something similar with getters & setters too.

PR #1657, opened 2019-02-13
Simplify ExtType<T>::Methods class

st-pasha:
Convert all code to use new `ADD_METHOD` macro; remove old templates `Methods::add<...>`. As a consequence, the `ExtType<T>::Methods` class is now much simpler.

The purpose of this refactoring is that complicated templates are not always compiled reliably by all compilers, which causes problems such as #1369. By switching to much simpler macros, these problems should be largely alleviated.

### DIFF
--- a/c/extras/py_ftrl.cc
+++ b/c/extras/py_ftrl.cc
@@ -36,10 +36,6 @@ PKArgs Ftrl::Type::args___init__(0, 2, 7, false, false,
                                  "lambda2", "nbins", "nepochs", "interactions"},
                                  "__init__", nullptr);
 
-NoArgs Ftrl::Type::fn___getstate__("__getstate__", nullptr);
-PKArgs Ftrl::Type::fn___setstate__(1, 0, 0, false, false, {"state"},
-                                   "__setstate__", nullptr);
-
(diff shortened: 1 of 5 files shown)


c/extras/py_ftrl.cc
```diff
@@ -610,7 +578,7 @@ Returns
 )");
 
 
-void Ftrl::reset(const NoArgs&) {
+void Ftrl::reset(const PKArgs&) {
```
oleksiyskononenko: So there is no `NoArgs` anymore?

st-pasha: Yeah, I don't think it has any use really: you can always declare `PKArgs(0,0,0,false,false)`. On the other hand, removing `NoArgs` and merging `PKArgs` with parent `Args` will simplify the class hierarchy and avoid virtual methods.

oleksiyskononenko: I've used `PKArgs(0,0,0,false,false)` some time ago, but then we changed that part to `NoArgs`. At some point I realized it was a good indicator that method takes no parameters, one don't need to go into details and can get such an info just from the `*Args` name. Not a big deal though...

PR #1658, opened 2019-02-13
Getters / setters revamp

st-pasha:
This PR changes changes the way getters / setters are declared, similar to #1657.

Now in order to declare a getter [+setter], you have to create a static `GSArgs` variable (similar to `PKArgs` for methods). This variable only carries the name of the property and its docstring. The advantage of the new approach is that now the documentation for a property can be declared right next to the property itself, which is a big plus.
In addition, the amount of generated code is now smaller, and we don't use templates as heavily. Instead, there are now 2 macros:
```
ADD_GETTER(gs, &Class::getterfn, args);
ADD_GETSET(gs, &Class::getterfn, &Class::setterfn, args);
```

### DIFF
--- a/c/extras/py_ftrl.cc
+++ b/c/extras/py_ftrl.cc
@@ -36,29 +36,6 @@ PKArgs Ftrl::Type::args___init__(0, 2, 7, false, false,
                                  "lambda2", "nbins", "nepochs", "interactions"},
                                  "__init__", nullptr);
 
-static const char* doc_alpha        = "`alpha` in per-coordinate FTRL-Proximal algorithm";
-static const char* doc_beta         = "`beta` in per-coordinate FTRL-Proximal algorithm";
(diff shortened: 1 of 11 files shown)


c/extras/py_ftrl.cc
```diff
@@ -783,133 +802,44 @@ oobj Ftrl::get_colname_hashes() const {
...
-                         << dt_model_in->ncols << " column"
-                         << (dt_model_in->ncols == 1? "": "s");
+//------------------------------------------------------------------------------
+// .alpha
+//------------------------------------------------------------------------------
 
-    }
+static GSArgs args_alpha(
+  "alpha",
+  "`alpha` in per-coordinate FTRL-Proximal algorithm");
```
oleksiyskononenko: I like this way of setting up docstrings much more than it was before.

PR #1669, opened 2019-02-19
Leak-proofing

### LINKED ISSUE #1596: memory leak and speed concerns
```
import numpy as np
import lightgbm_gpu as lgb
import scipy
import pandas as pd
from sklearn.utils import shuffle
from h2oaicore.metrics import def_rmse
import datatable as dt

def set_dt_col(train_dt, name, value):
    if isinstance(name, int):
        name = train_dt.names[name]
    train_dt[:, name] = dt.Frame(value)
    return train_dt

nrow = 4000
ncol = 5000
X = np.random.randn(nrow, ncol)
y = np.random.randn(nrow)
model = lgb.LGBMRegressor(objective='regression', n_jobs=20)  # 40 very slow
model.fit(X, y)

X_dt = dt.Frame(X)
cols_actual = list(X_dt.names)

do_numpy = False
score_f = def_rmse
preds = model.predict(X)
main_metric = score_f(actual=y, predicted=preds)
seed = 1234
def go():
    feature_importances = {}
    for n in range(ncol):
        print(n, flush=True)
        if do_numpy:
            shuf = shuffle(X[:,n].ravel())
            X_tmp = X # .copy()
            X_tmp[:,n] = shuf
            new_preds = model.predict(X_tmp)
            metric = score_f(actual=y, predicted=new_preds)
            col = "C" + str(n)
            feature_importances[col] = main_metric - metric
        else:
            col = cols_actual[n]
            shuf = shuffle(X_dt[:, col].to_numpy().ravel(), random_state=seed)
            X_tmp = set_dt_col(dt.Frame(X_dt), col, shuf)
            new_preds = model.predict(X_tmp)

            metric = score_f(actual=y, predicted=new_preds)
            feature_importances[col] = main_metric - metric
    return feature_importances

print(go())
```

Related to permutation variable importance.

If do_numpy = False, so it uses dt, then I see the resident memory slowly creep up from about 0.8GB to 1.6GB at n=1800 etc.  By n=4000 it's using 2.7GB.

If I use do_numpy = True, so it uses no dt, then I see resident memory never change over all n.

I thought at one point I only saw with LightGBM and not xgboost, but I'm not sure.

Unit tests like this numpy version by Microsoft show LightGBM not itself leaking: 
https://github.com/Microsoft/LightGBM/issues/1968

These 2 cases aren't doing exactly the same thing in that the numpy version keeps shuffling the same original X, while the dt version I think has essentially 2 copies, but the other original X_dt columns are not modified.  But @st-pasha you can confirm.

One can add the X_tmp = X.copy(), but it's not quite fair.  It makes a full copy, while dt should get away with only overwriting a single column.

Perhaps the flaw is how we are using dt and the frames?

st-pasha:
This PR attempts to make the process of finding leaks easier. To that end, 2 macros are defined: `TRACK` and `UNTRACK`. They are used wherever any dynamic memory is allocated / deallocated. The idea is that by the end of running a certain set of tests, and after destroying all objects created, there should be no more "tracked" objects left of any kind. If there are, then they were leaked. The current set of tracked objects can be queried using `dt.lib._datatable.get_tracked_objects()`.

These macros are only defined in debug mode, so there should be no performance impact for regular (non-debug) usage.

This approach helped discover several instances where small amount of memory was leaked:
- in Frame constructor, if there was an exception during the construction;
- in Column::from_buffer, the `view` object was sometimes not properly freed;
- python objects defined via `ext_type.h` were not freed properly upon destruction.

Closes #1596

### DIFF
--- a/c/datatablemodule.cc
+++ b/c/datatablemodule.cc
@@ -190,23 +190,29 @@ static std::unordered_map<void*, PtrInfo> tracked_objects;
 
 
 void TRACK(void* ptr, size_t size, const char* name) {
-  if (tracked_objects.count(ptr)) {
-    throw RuntimeError() << "Pointer " << ptr << " is already tracked. Old "
-        "pointer contains " << tracked_objects[ptr].to_string() << ", new: "
-        << (PtrInfo {size, name}).to_string();
+  #pragma omp critical
+  {
+    if (tracked_objects.count(ptr)) {
+      std::cerr << "ERROR: Pointer " << ptr << " is already tracked. Old "
+          "pointer contains " << tracked_objects[ptr].to_string() << ", new: "
+          << (PtrInfo {size, name}).to_string();
+    }
+    tracked_objects.insert({ptr, PtrInfo {size, name}});
   }
-  tracked_objects.insert({ptr, PtrInfo {size, name}});
 }
 
 
 void UNTRACK(void* ptr) {
-  if (tracked_objects.count(ptr) == 0) {
-    // UNTRACK() is usually called from a destructor, so cannot throw any
-    // exceptions there :(
-    std::cerr << "ERROR: Trying to remove pointer " << ptr
-              << " which is not tracked\n";
+  #pragma omp critical
+  {
+    if (tracked_objects.count(ptr) == 0) {
+      // UNTRACK() is usually called from a destructor, so cannot throw any
+      // exceptions there :(
+      std::cerr << "ERROR: Trying to remove pointer " << ptr
+                << " which is not tracked\n";
+    }
+    tracked_objects.erase(ptr);
   }
-  tracked_objects.erase(ptr);
 }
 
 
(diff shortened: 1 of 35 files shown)


c/datatablemodule.cc
```diff
@@ -193,6 +263,9 @@ void DatatableModule::init_methods() {
   #ifdef DTTEST
     init_tests();
   #endif
+  #ifdef DTDEBUG
```
oleksiyskononenko: Am I correct, that when one does `make build`, `DTDEBUG` becomes undefined? If so, then what will happen to `make test` in a release mode, if some of the Python tests invoke `get_tracked_objects()`?

st-pasha: `get_tracked_objects()` will not be available without the debug mode, you'll get an `AttributeError` exception. So any test that tries to use that method should properly check for its availability. For example, `dt.internal.in_debug_mode()` can help with that.

PR #1670, opened 2019-02-19
Fix `aggregate(...)` signature in help

oleksiyskononenko:
Make `aggregate(...)` signature be consistent with the aggregator functionality.

### DIFF
--- a/c/models/aggregator.cc
+++ b/c/models/aggregator.cc
@@ -113,8 +113,8 @@ static oobj aggregate(const PKArgs& args) {
   bool defined_double_precision = !args[10].is_none_or_undefined();
 
   if (undefined_dt) {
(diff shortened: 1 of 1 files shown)


c/models/aggregator.cc
```diff
@@ -33,7 +33,9 @@ static PKArgs args_aggregate(
   },
   "aggregate",
 
-R"(aggregate(dt, min_rows, n_bins, nx_bins, ny_bins, nd_max_bins, max_dimensions, seed, progress_fn, nthreads, double_precision)
+R"(aggregate(dt, min_rows=500, n_bins=500, nx_bins=50, ny_bins=50,
```
st-pasha: Call `dt` -> `frame` instead, so that there is no confusion with `import datatable as dt` ?

oleksiyskononenko: Done

PR #1673, opened 2019-02-20
Add option to display a Frame interactively

### LINKED ISSUE #1668: R Reticulate Support - Option to remove interactivity
I'm working with RStudio on an [issue](https://github.com/rstudio/reticulate/issues/431) involving the `datatable` Python API. We are using a library called `reticulate`, which does not currently support the interactivity shown in the image below.

image

Is it possible to have an option to turn off this interactivity so we can use via the `reticulate` within the RStudio IDE similar to how `pandas` is used?

st-pasha:
- Added option `dt.options.display.interactive` to control whether Frames will be displayed in interactive mode in a console or not.
- The default value of `dt.options.display.interactive` changed to `False`.
- Added a helper class `py::obool`.

Closes #1668

### DIFF
--- a/c/options.cc
+++ b/c/options.cc
@@ -28,6 +28,7 @@ int32_t sort_nthreads = 1;
 bool fread_anonymize = false;
 int64_t frame_names_auto_index = 0;
 std::string frame_names_auto_prefix = "C";
+bool display_interactive = false;
 bool display_interactive_hint = true;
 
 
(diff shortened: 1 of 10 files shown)


c/options.cc
```diff
@@ -193,6 +197,9 @@ static py::oobj get_option(const py::PKArgs& args) {
   } else if (name == "frame.names_auto_prefix") {
     return py::ostring(frame_names_auto_prefix);
 
+  } else if (name == "display.interactive") {
+    return py::oint(display_interactive);
```
oleksiyskononenko: Any reason there is no `py::obool`?

st-pasha: i guess just because nobody added it

st-pasha: fixed

oleksiyskononenko: cool, that was fast :)

PR #1696, opened 2019-03-01
Prevent overflow when constructing str32 columns

### LINKED ISSUE #1694: column.replace() may fail if the resulting string column is too big
```
>>> DT = dt.Frame(a=["a"] * 10000000)
>>> DT.replace("a", "A" * 250)
AssertionError: Assertion 'sb.size() == (mb.get_element<T>(n) & ~GETNA<T>())' failed in "c/column_string.cc", line 38
```
The problem here is that the size of the column created exceeds the limits of `int32`. The proper course of action in this case is to create the STR64 column instead.

st-pasha:
Direct construction of `StringColumn<T>` objects is now prohibited (except for the cases when a blank column is created). Instead, we use `new_string_column()` function which returns either `StringColumn<uint32_t>*` or `StingColumn<uint64_t>*`, whichever is more appropriate. 

This function also takes care of "fixing" the offsets array in case it overflown during the construction. This works provided that no individual string in the column exceeds 2Gb in size.

Closes #1694

### DIFF
--- a/c/column.h
+++ b/c/column.h
@@ -588,7 +588,6 @@ template <typename T> class StringColumn : public Column
 
 public:
   StringColumn(size_t nrows);
-  StringColumn(size_t nrows, MemoryRange&& offbuf, MemoryRange&& strbuf);
   void save_to_disk(const std::string& filename,
                     WritableBuffer::Strategy strategy) override;
   void replace_buffer(MemoryRange&&, MemoryRange&&) override;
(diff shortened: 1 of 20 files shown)


c/column.h
```diff
@@ -118,14 +118,17 @@ class Column
   size_t nrows;
 
 public:
+  static constexpr size_t MAX_STRING_SIZE = 0x7FFFFFFF;
+  static constexpr size_t MAX_STR32_BUFFER_SIZE = 0x7FFFFFFF;
+  static constexpr size_t MAX_STR32_NROWS = 0x7FFFFFFF;
```
oleksiyskononenko: If it is `size_t`, why we only do `0x7FFFFFFF`? Do we have `int32_t` coming into effect somewhere?

st-pasha: This is the maximum size for STR32 column. If actual size is bigger, it has to become STR64

PR #1704, opened 2019-03-06
Cast functionality improvements

### LINKED ISSUE #1665: Allow casting columns with the existing rowindex
`Column::cast` should also allow casting a column with rowindex.

### LINKED ISSUE #1695: Conversion to string may fail if the column is too big
```
>>> DT = dt.Frame(range(300000000))
>>> DT[:, "A"] = dt.str32(f[0])
ValueError: The size of new_strbuf does not correspond to the last offset of new_offbuff: expected 2588888890, got 441405242
```
The total size of string data in the column exceeds `int32`, but the converter is not smart enough to automatically convert to `str64`.

### LINKED ISSUE #1705: Memory leak in DT[i, j] expression
Minimal reproducible example:
```python
import datatable as dt
from datatable.lib._datatable import get_tracked_objects

def test():
    DT1 = dt.Frame(A=range(10000))
    DT2 = DT1[:, dt.f.A * 2]

test()
print(get_tracked_objects())
```

st-pasha:
- All cast functionality moved to "frame/cast.cc" (with the exception of `cast_expr` class in "exprs/base_expr.cc", which we might eventually move too);
- Implement casting a column with an existing rowindex;
- Fixed leak when converting string columns into pandas;
- Conversion into string type now happens via common `generate_string_column()` function from parallel.h;
- When converting a large column into str32, it can no longer overflow. Instead, a str64 column will be returned.
- Fixed memory leak when computing any DT[i,j] expression;
- Fixed memory leak in `dt::generate_string_column` (buffers were not properly destroyed).

Closes #1695
Closes #1665
Closes #1705


oleksiyskononenko:
What happens to rowindex, when we cast to the same stype as the original column has? Does the casted column still carry the same rowindex or it gets materialized?

st-pasha:
It gets materialized

oleksiyskononenko:
@st-pasha ic, this is what I also observed... this won't change at some point?

PR #1711, opened 2019-03-07
FTRL model re-factoring and new capabilities

### LINKED ISSUE #1498: Support early stopping in FTRL algo
Allow a validation dataset to be provided as an optional parameter in `py::Ftrl::fit` method, and subsequently in `dt::Ftrl::fit`. If provided, it should be used for early stopping, taking priority over the `nepochs` parameter.

Prerequisite: #1574

### LINKED ISSUE #1574: Move learning functionality from `py::Ftrl` to `dt::Ftrl`
Adding multinomial classification capabilities resulted in `py::Ftrl` class not just controlling `dt::Ftrl`, but also doing irrelevant things like detecting a problem type, creating binomial classifiers or averaging feature importance information.

We should keep it as 
- `py::Ftrl` class provides Python related functionality, and calls `dt::Ftrl` for training and predicting;
- `dt::Ftrl` does all the FTRL-related work for all the different problem types.

In this case we will not need as many `dt::Ftrl`s as the number of classes, one `dt::Ftrl` could solve any problem. This will allow us to save on many costly operations like hashing, and would allow to properly estimate the loss.

oleksiyskononenko:
This PR introduces a major re-factoring of the FTRL model, namely:
- all the FTRL fitting/predicting functionality is moved from `py::` to `dt::`, making it essentially one model, rather than a set of binomial classifiers. In this case we save on hashing in the case of multinomial classification, and also can control the loss;
- FTRL model can now stop early, if there is no improvement during the training process. The loss is calculated every `nepochs_validate` epochs, that is now an input parameter in the `fit(...)` method, and if doesn't decrease, training stops; 
- FTRL model can now do all the calculations in single or double precision arithmetic, depending on the constructor input parameter `double_precision`;
- FTRL multinomial model now automatically detects all the labels in the target column, and only trains on them. Plus, on the so-called "negative" label, that has all the negative targets. When the model trains on another dataset and sees a new label, it creates a copy of the "negative" label weights and starts training from there;
- other re-factoring and improvements, added tests.

The re-factored FTRL model is now more efficient in terms of memory (~x3), and performance (~20-30% depending on a dataset).

Closes #1574
Closes #1498


c/models/dt_ftrl_real.h
```diff
@@ -0,0 +1,1078 @@
...
+
+template <typename T>
+void FtrlReal<T>::create_model() {
+  size_t nlabels = labels.size();
+  if (nlabels == 0) ValueError() << "Model cannot have zero labels";
+
+  size_t ncols = 2 * nlabels;
+  colvec cols(ncols);
+  for (size_t i = 0; i < ncols; ++i) {
+    cols[i] = Column::new_data_column(stype<T>::get_stype(), nbins);
```
st-pasha: why not simply `new RealColumn<T>(nbins)`?

c/models/dt_ftrl_real.h
```diff
@@ -0,0 +1,1078 @@
...
+
+
+template <typename T>
+dtptr FtrlReal<T>::create_dt_p(size_t nrows) {
+  size_t nlabels = labels.size();
+  if (nlabels == 0) ValueError() << "Prediction frame cannot have zero labels";
+
+  colvec cols_p(nlabels);
+  for (size_t i = 0; i < nlabels; ++i) {
+    cols_p[i] = Column::new_data_column(stype<T>::get_stype(), nrows);
```
st-pasha: can be replaced with `new RealColumn<T>(nrows)`

c/models/utils.h
```diff
@@ -21,12 +21,87 @@
...
+    return SType::FLOAT32;
+  }
+};
+
+
+template<> struct stype<double> {
+  static SType get_stype() {
+    return SType::FLOAT64;
+  }
+};
```
st-pasha: Generally, there is no 1-to-1 correspondence between stypes and element types `T`.
We have template `element_t<stype>` (declared in column.h) that provides a mapping `stype -> T`, but the inverse is ill-defined, especially once we start adding new types such as datetime, uint64, categorical, array, etc.

c/models/utils.h
```diff
@@ -21,12 +21,87 @@
...
+
+void calculate_coprimes(size_t, std::vector<size_t>&);
+
+/*
+*  Sigmoid function.
+*/
+template<typename T>
+inline T sigmoid(T x) {
+  T one = static_cast<T>(1.0);
```
st-pasha: should be `constexpr`

c/models/utils.h
```diff
@@ -21,12 +21,87 @@
...
+/*
+* Calculate logloss(p, y) = -(y * log(p) + (1 - y) * log(1 - p)),
+* where p is a prediction and y is the actual target:
+* - apply minmax rule, so that p falls into the [epsilon; 1 - epsilon] interval,
+*   to prevent logloss being undefined;
+* - simplify the logloss formula to more compact branchless code.
+*/
+template<typename T>
+inline T log_loss(T p, bool y) {
+  T epsilon = std::numeric_limits<T>::epsilon();
```
st-pasha: `constexpr`

c/models/dt_ftrl_real.h
```diff
@@ -0,0 +1,1078 @@
...
+    void set_interactions(bool) override;
+    void set_double_precision(bool) override;
+    void set_labels(strvec) override;
+};
+
+
+/*
+*  Set up parameters and initialize weights.
+*/
+template <typename T>
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, sure. Forgot to do it at the very end. Now it's done.

c/models/dt_ftrl_real.h
```diff
@@ -0,0 +1,1078 @@
...
+void FtrlReal<T>::fill_ri_data(const DataTable* dt,
+                               std::vector<const RowIndex>& ri,
+                               std::vector<const U*>& data) {
+	size_t ncols = dt->ncols;
+  ri.reserve(ncols);
+  data.reserve(ncols);
+	for (size_t i = 0; i < ncols; ++i) {
+  	data.push_back(static_cast<const U*>(dt->columns[i]->data()));
+	  ri.push_back(dt->columns[i]->rowindex());
+	}
```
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/models/dt_ftrl_real.h
```diff
@@ -0,0 +1,1078 @@
...
+			T loss_diff = loss_global_prev - loss_global;
+			if (c && loss_diff <= std::numeric_limits<T>::epsilon()) break;
+			// If loss decreases, save current loss and continue training.
+			loss_global_prev = loss_global;
+			loss_global = 0;
+		}
+	}
+  double epoch = static_cast<double>(chunk_end) / dt_X->nrows;
+	return epoch;
+}
```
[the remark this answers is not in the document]
oleksiyskononenko: Fixed

c/models/dt_ftrl_real.h
```diff
@@ -0,0 +1,1078 @@
...
+  	data.push_back(static_cast<const U*>(dt->columns[i]->data()));
+	  ri.push_back(dt->columns[i]->rowindex());
+	}
+}
+
+/*
+*  Fit model on a datatable.
+*/
+template <typename T>
+template <typename U /* column data type */, typename F /* link function */,
```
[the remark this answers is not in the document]
oleksiyskononenko: Right now for multinomial case I do a call `fit<int8_t>(nepochs_val, sigmoid<T>, log_loss<T>)`, so basically treat its loss just like an average of losses for a set of binomial classifiers. And you are right, in this case `F` and `G` can be deduced from `T` and `U`. 

However, when I do predictions I do something slightly different. I use softmax function for multinomial case, that is equivalent of using `std::exp` link function, and then doing normalization. I'm not sure for the moment if I also need to do softmax in `fit(...)`, and, if I do, should I continue using `log_loss` as a loss function or not. If I need to change it, then I will need this `F` and `G`.

PR #1715, opened 2019-03-08
Added median() function

### LINKED ISSUE #1530: implement median function
Already mentioned in #290 as _Group-by operators that involve sorting: median, quantile, rank_. Filling here so I can have boolean status when `median` is done as it is currently required for q6 in new set of questions in groupby benchmark.

st-pasha:
- In `Column` class added method `sort_grouped` which allows to sort a column within the existing groupby structure;
- Reworked `expr/reduce.cc`: now all reduce functions are collected into a map (op,stype)->reducer, which simplifies the logic and makes it easier to add new reducers;
- Added reducer `median()` which computes the median of the values in a column. The return type is always float64 (except when the input is float32, the output is also float32). The median ignores any NA values;
- Fixed a bug where `expr_reduce::resolve()` was not returning proper stype of the result (was marked as FIXME).

Closes #1530


jangorecki:
on `G1_1e7_1e2_0_0` I am getting
```
>>> ans = x[:, {"median_v3": median(f.v3), "sd_v3": sd(f.v3)}, by(f.id2, f.id4)]
Segmentation fault (core dumped)
```

st-pasha:
Thanks @jangorecki , I'll try to investigate

jangorecki:
@st-pasha should I create new issue for that? or it should be resolved already?

PR #1721, opened 2019-03-10
Upgrade the wheel URL in requirements_docs.txt

### LINKED ISSUE #1720: Documentation for 0.8.0 cannot be built
Currently the documentation for 0.8.0 is missing entirely: https://datatable.readthedocs.io/en/v0.8.0/

The reason for this is because documentation cannot be built off branch `v0.8.0` due to RTD's system upgrade: they recently switched from Python3.5 to Python3.7. The branch v0.8.0 had file "requirements_docs.txt" which referred to a datatable wheel for python 3.5. This wheel fails to install on python 3.7, obviously, so the entire "v0.8.0" version of the docs cannot be built.

One possible solution would be to update v0.8.0 branch to use a different wheel, suitable for py3.7.

st-pasha:
Ideally, we should be able to say `datatable==0.8.0` in the requirements file, but it won't work because we can't publish Linux wheels on PyPI. Instead, we'll be using the hardcoded wheel URL referring to our S3 bucket.

Closes #1720

### DIFF
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -1,3 +1,3 @@
 sphinx>=1.8
 nbsphinx>=0.3.5
-https://s3.amazonaws.com/artifacts.h2o.ai/releases/ai/h2o/pydatatable/0.7.0.dev492/x86_64-centos7/datatable-0.7.0.dev492-cp35-cp35m-linux_x86_64.whl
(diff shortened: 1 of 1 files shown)


st-pasha:
It appears this can be built on readthedocs just fine, so merging.

st-pasha:
I've rebuilt the docs v0.8.0, now it looks fine: https://datatable.readthedocs.io/en/v0.8.0/index.html

PR #1723, opened 2019-03-13
Allow for arbitrary feature interactions in FTRL algo

### LINKED ISSUE #1397: Allow for an arbitrary feature interactions in FTRL algo
For the moment there is an argument `inter` that controls if 2-nd order feature interaction should be used for learning or not. When number of features is large, doing full 2-nd order interaction is not practical, as it scales with `n_features^2`. Instead, we can allow for a way to learn only for "interesting" feature combinations, may be even for the orders higher than 2.

oleksiyskononenko:
In this PR we replace full second order feature interactions with a possibility to do arbitrary feature combinations:

- replace a boolean `.interactions` parameter with a list of feature lists to do interactions for;
- store column names in `py::Ftrl` class to ensure the names are consistent between training and predictions;
- adjust tests, minor re-factoring and cosmetics.

Closes #1397


docs/ftrl.rst
```diff
@@ -124,11 +122,27 @@ can be accessed as
...
+by FTRL model. User can provide additional features by specifying
+a list of feature interactions, for instance as
+
+::
+
+  ftrl_model.interactions = [["C0", "C1", "C3"], ["C2", "C5"]]
+
+where ``C*`` are column names from a training dataset. `interactions`
+should be set before a call to `fit()` method, and can not be changed
+once the model is trained.
```
st-pasha: Not entirely clear from this doc what kind of interactions are created: is this 2 interactions `C0*C1*C3` and `C2*C5`; or 6 interactions `C0*C2, C0*C5, C1*C2, C1*C5, C3*C2, C3*C5` ?

c/models/py_ftrl.cc
```diff
@@ -149,13 +152,48 @@ void Ftrl::m__dealloc__() {
...
+  interactions.reserve(py_iter.size());
+
+  for (auto py_interaction : py_iter) {
+    size_t nfeatures = py_interaction.to_pylist().size();
+    sizetvec interaction;
+    interaction.reserve(nfeatures);
+    for (auto py_feature : py_interaction.to_oiter()) {
+      std::string feature = py_feature.to_string();
+
+      auto it = find(colnames.begin(), colnames.end(), feature);
```
st-pasha: You could create a blank `DataTable` object with `colnames` as names, and then use `DataTable::xcolindex()` to convert each column name into its index. This would use a dictionary in order to do value look up, so would be slightly faster, and also emit a nice error message checking for possible misspellings of a column name (e.g. "column `c0` not found, did you mean `C0`?")

docs/ftrl.rst
oleksiyskononenko: Thanks, clarified this part. May need to work on a more detailed manual at some point.

c/models/py_ftrl.cc
oleksiyskononenko: That's a good idea, however, what about the custom error messages in this case? For example, now I emit

```
        throw ValueError() << "Feature " << py_feature << " is used for "
                              "interactions, however, it is missing in the "
                              "training frame";
```

If I use `DataTable`, it will just emit something general?

oleksiyskononenko:
@st-pasha I first thought I need copies of all those to return, and now I see that in reality I only need a reference. Converted all these methods to `const vector&`.

Issue #1699, opened 2019-03-02
Enhancement of using-datatable.rst documentation. Add importing operator "by", "sort", examples, etc.


Viktor-Demin:
```
import datatable as dt
from datatable import f, min, max,mean
import math
DT1 = dt.Frame(A=range(5), B=[1.7, 3.4, 0, None, -math.inf], C = [1,1,1,2,2],
               stypes={"A": dt.int64})


DT1[:,mean(f.A),by(f.C)]
```

Returns:

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-62-1e6b6dedb0de> in <module>()
      6 
      7 
----> 8 DT1[:,mean(f.A),by(f.C)]
      9 

NameError: name 'by' is not defined
```

version 0.8.0

oleksiyskononenko:
@Cheburusska one should also import `by`, then it will work:

```
>>> import datatable as dt
>>> from datatable import f, min, max, mean, **by**
>>> import math
>>> DT1 = dt.Frame(A=range(5), B=[1.7, 3.4, 0, None, -math.inf], C = [1,1,1,2,2],
...                stypes={"A": dt.int64})
>>> 
>>> 
>>> DT1[:,mean(f.A),by(f.C)]
      C   C0
---  --  ---
 0    1  1  
 1    2  3.5

[2 rows x 2 columns]
```

Cheburusska:
@oleksiyskononenko you are very right and now it works. I believe it is worth to add in documentation to avoid confusing. I would to do it myself and contribute to this project. Do we have any manual for that?

oleksiyskononenko:
@Cheburusska Sure, here is the [manual](https://datatable.readthedocs.io/en/latest/?badge=latest), and this is the [groupby](https://datatable.readthedocs.io/en/latest/quick-start.html#by) section under `Getting started` that already has this info. 

Guess the source of confusion is that there is also `Using datatable` section (that is the older version of `Getting started`) that may not be that detailed. The idea was to move all the items from `Using datatable` to `Getting started` making them as detailed as possible. This is now WIP, and you are welcome to contribute, of course.

st-pasha:
Hi @Cheburusska, thank you for your feedback.
After using the `datatable` for a while, importing these symbols becomes almost a second nature. Which is why it's so easy to forget to write about it in the documentation. So I absolutely agree that this is something that should be added.

As for where to add this info, I imagine this can be done in several places:
[...]

PR #1726, opened 2019-03-14
#1699 Enhancement of documentation

Viktor-Demin:
closes https://github.com/h2oai/datatable/issues/1699
1) Merged quick-start.rst and using datatable.rst. and last one was deleted to avoid confusing
2) added a lot of new examples which hardcoded as raw html until ipython directive is fixed
3) created tutorial.ipynb - same as quick-start.rst but it can be executed. Maybe we can share it with users to give them ability of fast start? if not it can be deleted. 
4) fixed link in readme previous one didn't lead to anything


st-pasha:
(btw, I'm on vacation currently, so I'd be able to take a look at this PR only later this week)

Cheburusska:
@st-pasha thank you for notification. Have a nice vacation =)

Cheburusska:
@st-pasha could you please review this request if you returned from vacation?

Cheburusska:
@st-pasha kind reminder.

Cheburusska:
@st-pasha @oleksiyskononenko could you please tell me what is a standard process of contribution? Maybe I can do something to help with review? Or maybe there are some problems with my request which I should fix? Maybe there is something what blocks my requests?

st-pasha:
The major change in this PR is to merge `using-datatable.rst` and `quick-start.rst`, however such merger is not actually desirable. These two docs serve a very distinct purpose: "quick-start" is supposed to be a very quick introduction into the capabilities of datatable, almost like a cheatsheet, with lots of code and few explanations. It could benefit from being made more succinct, more broad, and having more cross-references to other documents. On the other hand, "using-datatable" is supposed to be a more in-depth exposition, almost like a Wikipedia article. The contents of "using-datatable" may be even split across multiple pages, if a single page becomes too long.

It is also unnecessary to replace quotation marks with "smart quotes": typography is a good thing, but such replacements are already done by a plugin, so they are superfluous in code.

Lastly, a large portion of this PR is an ipynb example file. Examples are good, and we need them, but in this case the example is just a doc page converted into a Jupyter notebook, which is probably not what we want.

Issue #1727, opened 2019-03-18
Include gcc compiled datatable in Jenkins CI


mmalohlava:
## Feature 
Jenkins CI now use only clang to compile datatable. The goal is to use gcc and publish datatable artifacts compiled by gcc.

st-pasha:
Completed in #1810

PR #1744, opened 2019-03-25
Dt gcc jenkins

abal5:
#1727 Using Tom's branch as base to test/build on Jenkins for gcc.


st-pasha:
This should have been based on top of master.
Replacing with #1746

PR #1746, opened 2019-03-25
Fix warnings/errors when compiling with gcc

st-pasha:
- centos7 docker image now always uses gcc for compilation;
- `monkey_patch_compiler` is now executed on all platforms, and it attempts to remove compilation flags that are harmful;
- added `FALLTHROUGH` macro which uses either clang or gcc "fallthrough" attribute;
- some warnings ignored;
- move constructor of `Error()` does copying on gcc4.8, since it didn't support moving string streams;
- disabled tests of error messages from bad regular expressions: these messages are compiler-specific;
- setup.py now checks the python version first, to give a reasonable error message if the user accidentally tries to compile datatable under python2.7


st-pasha:
There are still problems with gcc4.8.5 compiler(wrong test results for regular expressions); for now use Clang compiler instead.

PR #1747, opened 2019-03-26
Use static schedule in dt::run_parallel()

st-pasha:
In tests of cast functionality, this improved functionality significantly.

### DIFF
--- a/c/utils/parallel.cc
+++ b/c/utils/parallel.cc
@@ -26,16 +26,16 @@ namespace dt {
 
 void run_parallel(rangefn run, size_t nrows)
 {
-  // `min_nrows_per_thread`: avoid processing less than this many rows in each
+  // `min_nrows_per_chunk`: avoid processing less than this many rows in each
   // thread, reduce the number of threads if necessary.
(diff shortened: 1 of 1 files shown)


oleksiyskononenko:
Confirm that this change fixes Aggregator performance. Testing FTRL now.

c/utils/parallel.cc
```diff
@@ -26,42 +26,59 @@ namespace dt {
...
-                        static_cast<int>(nrows / min_nrows_per_thread));
+                        static_cast<int>(nrows / min_nrows_per_chunk));
     xassert(nth0 > 0);
     OmpExceptionManager oem;
     #pragma omp parallel num_threads(nth0)
     {
       size_t ith = static_cast<size_t>(omp_get_thread_num());
       size_t nth = static_cast<size_t>(omp_get_num_threads());
-      size_t batchsize = min_nrows_per_batch * nth;
+      size_t nchunks = std::min(static_cast<size_t>(nth0),
```
oleksiyskononenko: We should probably use `nth` and not `nth0` here?

Issue #1736, opened 2019-03-22
Eliminate dependency on OpenMP


st-pasha:
We rely on OpenMP to parallelize the code. This solution has its advantages and drawbacks:

### PROs
- This is very simple to do: a single `#pragma omp` is sufficient to parallelize a loop;
- OMP implementation is well-tested and robust, with very few bugs;

### CONs
- The easiest approach is not always the best. OpenMP is not 100% efficient, especially when it comes to ordered loops;
- If a problem occurs in OpenMP code, it is very hard to debug;
- Bugs in auto-generated OpenMP code are impossible to fix, because this code is added at the compiler level (there will be bugs in our own implementation too, and probably much more, but at least we'd be able to find and fix them);
- There is a known problem where forking a process with an existing OpenMP pool will cause the child process to deadlock;
- OpenMP support requires us to distribute an openmp dynamic library, which itself is problematic, and for example prevents us from creating "many-linux" distribution on PyPI;
- If another library uses OpenMP too (eg. numpy), there may be a problem with using these libraries together;
- OpenMP provides a number of useful parallel constructs out of the box, but they cannot be extended easily.

## Roadmap

- [x] Use atomics instead of OpenMP in `shared_bmutex`;
- [x] **Implement basic thread pool**;
- [x] Implement parallel for loop with static scheduling (see "parallel.cc");
- [x] Implement parallel for loop with dynamic scheduling (#1772);
- [x] Implement parallel for ordered loop using a thread pool (#1775);
- [x] Implement atomic float and double (with arithmetics);
- [x] Implement atomic min and max functions (#1763);
- [x] Implement parallel_for_static nested inside parallel_region (#1762);
- [x] Implement parallel_for_dynamic nested inside parallel_region (#1776);
- [x] Avoid critical section in `TRACK()` and `UNTRACK()`;
- [x] Avoid `#pragma omp` in column_fw.cc;
- [x] Avoid `#pragma omp` in column_string.cc;
- [x] Eliminate `#pragma omp` in rowindex_array.cc;
- [x] Get rid of `#pragma omp` in sort.cc;
- [x] Remove `#pragma omp`s in stats.cc;
- [x] Use new thread pool in csv writer;
- [x] Get rid of `#pragma omp` in expr/reduce.cc;
- [x] Stop using `#pragma omp` in expr/string_expr.cc;
- [x] Avoid `#pragma omp` in frame/join.cc;
- [x] Expel `#pragma omp` from to_numpy.cc;
[...]

PR #1751, opened 2019-03-29
Thread pool implementation

st-pasha:
This PR adds an internal thread pool + supporting classes, which will eventually be used to replace OMP parallelism.

Currently supported functionality:
- increase/reduce number of threads in the pool;
- `dt::run_once_per_thread()` function can be used for plain parallelization (similar to plain `#pragma omp parallel`);
- generic scheduler mechanism, which can be extended to implement other types of parallelism, such as parallel for or parallel ordered for;
- python function `dt.internal.get_thread_ids()` return a list of strings which are implementation-dependent thread ids (can be used for debugging, maybe);
- added macro `CACHELINE_SIZE` and struct `cache_aligned<T>`.

WIP for #1736


c/parallel/thread_scheduler.cc
```diff
@@ -0,0 +1,156 @@
...
+  return &tsleep[index];
+}
+
+void thread_sleep_scheduler::wait_until_finish() {
+  // This is the only non-joinable task. Don't do anything here.
+}
+
+void thread_sleep_scheduler::awaken(thread_scheduler* next) {
+  size_t i = index;
+  size_t j = (i + 1) & 1;
```
oleksiyskononenko: What is this `j` index?

c/parallel/thread_scheduler.h
```diff
@@ -0,0 +1,183 @@
...
+ * The master thread that called `awaken(job)` will then call `job.join()`,
+ * and it is the responsibility of thread_scheduler `job` to wait until all
+ * threads have finished execution and were put back to sleep. Thus, the master
+ * thread ensures that all threads are sleeping again before the next call to
+ * `awaken`.
+ */
+class thread_sleep_scheduler : public thread_scheduler {
+  private:
+    thread_sleep_task tsleep[2];
+    size_t index = 0;  // switches between 0 or 1
```
oleksiyskononenko: Could you explain what this `index` is? Btw, if it only takes 0/1 values, may be we make it `bool`?

c/datatablemodule.cc
```diff
@@ -142,6 +145,28 @@ static py::oobj has_omp_support(const py::PKArgs&) {
 }
 
 
+static py::PKArgs args_get_thread_ids(
+    0, 0, 0, false, false, {}, "get_thread_ids",
+R"(Return internal ids of all threads used internally by datatable)");
```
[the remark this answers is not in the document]
st-pasha: fixed

c/parallel/thread_scheduler.cc
st-pasha: see my new commit

PR #1752, opened 2019-03-30
FTRL algo and docs improvements

oleksiyskononenko:
In this PR we improve documentation to reflect the latest FTRL developments, and also
- add a `negative_class` parameter to FTRL constructor, that controls algo behavior in the case of multinomial classification. By default `negative_class`  is `False`, so that there is no "_negative" label that is created, and when we encounter new labels their weights are zero initialized. When `negative_class` is set to `True`, then in addition to existing labels a special "_negative" label is created and trained on. When we encounter new labels, their weight coefficients are initialized with a shallow copies of the "_negative" coefficients;
- change default value for `lambda2` to zero, make default parameters consistent between the code, docstrings and `.rst` file;
- add an additional `sort` parameter to `split_into_nhot` function. If set to `True`, column names are returned in an alphabetical order, otherwise their order can not be guaranteed due to parallelization;
- fix minor bug in `run_parallel`, where `nchunks` was calculated based on `nth0` and not on the actual `nth` number of threads;
- overall re-factoring and cleaning, improving test coverage.


c/models/utils.h
```diff
@@ -21,18 +21,40 @@
...
+
+/*
+*  Create list of sorting indexes.
+*/
+template <typename T>
+sizetvec sort_index(const std::vector<T> &v) {
+
+  // Initialize original index locations
+  sizetvec index(v.size());
+  iota(index.begin(), index.end(), 0);
```
st-pasha: better use `std::iota` and `std::sort` here, otherwise some compilers will complain

c/str/split_into_nhot.cc
```diff
@@ -165,6 +166,21 @@ DataTable* split_into_nhot(Column* col, char sep) {
...
+    std::vector<std::string> outnames_sorted(ncols);
+    std::vector<Column*> outcols_sorted(ncols);
+    std::vector<size_t> colindex = sort_index<std::string>(outnames);
+    for (size_t i = 0; i < outnames.size(); ++i) {
+      size_t j = colindex[i];
+      outnames_sorted[i] = outnames[j];
+      outcols_sorted[i] = outcols[j];
+    }
+    outcols = outcols_sorted;
+    outnames = outnames_sorted;
```
st-pasha: `outnames = std::move(outnames_sorted)` to avoid copy, same with `outcols`

c/str/split_into_nhot.cc
```diff
@@ -165,6 +166,21 @@ DataTable* split_into_nhot(Column* col, char sep) {
...
 
+  // Re-order columns, so that column names go in alphabetical order.
+  if (sort) {
+    size_t ncols = outcols.size();
+    std::vector<std::string> outnames_sorted(ncols);
+    std::vector<Column*> outcols_sorted(ncols);
+    std::vector<size_t> colindex = sort_index<std::string>(outnames);
+    for (size_t i = 0; i < outnames.size(); ++i) {
+      size_t j = colindex[i];
+      outnames_sorted[i] = outnames[j];
```
st-pasha: `outnames_sorted[i] = std::move(outnames[j]);` to avoid copying

c/str/split_into_nhot.cc
```diff
@@ -165,6 +166,21 @@ DataTable* split_into_nhot(Column* col, char sep) {
 
   oem.rethrow_exception_if_any();
 
+  // Re-order columns, so that column names go in alphabetical order.
+  if (sort) {
+    size_t ncols = outcols.size();
+    std::vector<std::string> outnames_sorted(ncols);
+    std::vector<Column*> outcols_sorted(ncols);
+    std::vector<size_t> colindex = sort_index<std::string>(outnames);
+    for (size_t i = 0; i < outnames.size(); ++i) {
```
st-pasha: `outnames.size()` -> `ncols` for brevity

oleksiyskononenko: Thanks, yes, all fixed.

PR #1755, opened 2019-04-01
Implement class atomic<double> and atomic<float>

st-pasha:
`std::atomic<double>` is specialized only in C++17, not in C++11. In C++11 there is generic `std::atomic<T>`, but it lacks arithmetic operators, which are very important for our use case. Hence, we implement our own atomic class, which reuses atomic ints at bit level.

WIP for #1736

### DIFF
--- a/c/models/dt_ftrl_real.cc
+++ b/c/models/dt_ftrl_real.cc
@@ -20,6 +20,7 @@
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
 #include "models/dt_ftrl_real.h"
+#include "parallel/atomic.h"
 #include "utils/macros.h"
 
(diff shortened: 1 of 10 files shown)


c/models/dt_ftrl_real.cc
```diff
@@ -388,20 +388,19 @@ double FtrlReal<T>::fit(T(*linkfn)(T), T(*lossfn)(T,U)) {
               }
             }
           }
-          #pragma omp atomic
           loss += loss_local;
```
oleksiyskononenko: With your implementation of `dt::atomic`, will it do `loss += loss_local` atomically or we need to use some `loss.operation()` here?

PR #1759, opened 2019-04-01
Correct behavior of thread pool during forking

### LINKED ISSUE #1757: Delay initializing thread pool until actually needed
Currently we create the thread pool immediately when the library initializes, but it would probably be beneficial to postpone this initialization until slightly later. This would allow common idiom
```
import datatable as dt
dt.options.nthreads = 1
```
to work, without having a large pool created first, only to destroy it later.

### LINKED ISSUE #1758: Fortify the thread pool against forking
When a process is forked (but not exec'd), the child spawns without its parent's threads; which leads to a broken thread pool: whenever any job is executed, we expect the threads to wake up and start picking the tasks -- however, if there are no threads then no tasks are ever executed and the program freezes.

We can guard against this using `pthread_atfork()` library function.

st-pasha:
This PR introduces two changes to the thread pool:

1. The threads are now initialized lazily, that is on the first run that they are actually required. For example function `get_thread_ids()` will always cause the threads to spawn, whereas `dt::run_parallel()` may not, if the number of rows is small enough to run single-threadedly.

2. Fixed problems with datatable hanging after the process was forked by an external module such as `multiprocessing`. Now we re-initialize the thread pool in every child process, ensuring that they do not accidentally share mutexes / condition variables / threads. 

Closes #1758 
Closes #1757

### DIFF
--- a/tests/test_dt.py
+++ b/tests/test_dt.py
@@ -176,6 +176,20 @@ def test_coverage():
         core.test_coverage()
 
 
+def test_multiprocessing_threadpool():
+    # Verify that threads work properly after forking (#1758)
+    import multiprocessing as mp
+    from datatable.internal import get_thread_ids
+    parent_threads = get_thread_ids()
+    n = 4
+    pool = mp.Pool(processes=n)
+    child_threads = pool.starmap(get_thread_ids, [()] * n, chunksize=1)
+    assert len(child_threads) == n
+    for chthreads in child_threads:
+        assert len(parent_threads) == len(chthreads)
+        assert chthreads != parent_threads
+
+
 def test_internal_shared_mutex():
     from datatable.lib import core
     if hasattr(core, "test_shmutex"):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed conversion to numpy of a view Frame which contains NAs (#1738).
 
+- `datatable` can now be safely used with `multiprocessing`, or other modules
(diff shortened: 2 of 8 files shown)


tests/test_dt.py
```diff
@@ -176,6 +176,20 @@ def test_coverage():
...
+    import multiprocessing as mp
+    from datatable.internal import get_thread_ids
+    parent_threads = get_thread_ids()
+    n = 4
+    pool = mp.Pool(processes=n)
+    child_threads = pool.starmap(get_thread_ids, [()] * n, chunksize=1)
+    assert len(child_threads) == n
+    for chthreads in child_threads:
+        assert len(parent_threads) == len(chthreads)
+        assert chthreads != parent_threads
```
oleksiyskononenko: When `get_thread_ids()` is called, does it always return threads in the same order? If by accident the order is different between child and parent, but threads are the same, may be it is safer to do `assert Counter(chthreads) == Counter(parent_threads)`?

st-pasha: Thread ids are always returned in the same order, in the sense that repeated calls to `get_thread_ids()` will always produce the same list. However, thread ids in child processes will necessarily be different from the thread ids of the parent, because children do not inherit threads from the parent. Therefore I test `chthreads != parent_threads` (it is allowed, though very unlikely, for some thread ids to accidentally be the same in 2 processes, but for all of them to be same is practically impossible). 
The standard does guarantee that all thread ids in a single process are unique, so there is no point in using `Counter`.

oleksiyskononenko: I see, if they always come in the same order then it's ok.

PR #1767, opened 2019-04-03
Remove #pragma omp uses in aggregator.cc

st-pasha:
WIP for #1736


c/models/aggregator.cc
```diff
@@ -549,14 +561,17 @@ void Aggregator<T>::group_1d_categorical() {
...
 
-  #pragma omp parallel for schedule(dynamic)
-  for (size_t i = 0; i < grpby0.ngroups(); ++i) {
-    size_t off_i = static_cast<size_t>(offsets0[i]);
-    size_t off_i1 = static_cast<size_t>(offsets0[i+1]);
-    for (size_t j = off_i; j < off_i1; ++j) {
-      d_members[ri0[j]] = static_cast<int32_t>(i);
-    }
-  }
+  // TODO: use dynamic instead?
```
oleksiyskononenko: It looks like we don't have many use cases for categorical aggregation, so can leave with static for the moment.

oleksiyskononenko:
Thanks, let me test it on larger datasets and some dozens of threads.

c/models/aggregator.cc
```diff
@@ -730,7 +750,7 @@ void Aggregator<T>::group_nd() {
   if (do_projection) pmatrix = generate_pmatrix(ncols);
 
   // Figuring out how many threads to use.
-  size_t nth0 = get_nthreads(nrows);
+  size_t nth = std::min(get_nthreads(nrows), dt::get_num_threads());
```
oleksiyskononenko: `get_nthreads()` uses `config::nthreads` to determine number of threads, and then did `#pragma omp parallel num_threads(nth0)`. Since now `dt::parallel_region` doesn't accept number of threads as a parameter, how can we change number of threads?

st-pasha: i'll add that as a parameter

oleksiyskononenko:
Did some benchmarks for the airlines dataset (nrows = 123534969, ncols = 31) on Linux with 40 threads.

- ND aggregation: thread_pool ~34[s] vs OpenMP ~31[s];
- 1D aggregation for all the columns: thread_pool ~135[s] vs OpenMP ~93[s].

oleksiyskononenko:
Timing to do `pytest ./tests/models/test_aggregate.py`:

- thread_pool: 0.31[s]
- OpenMP: 0.13[s]

These are mostly 1 and 2D tests, so seems like we need to take a look at their performance.

st-pasha:
I've added `nthreads` parameter in #1771, so you can use that to limit the number of threads for a particular parallel region.
Generally, I'm ok with taking a small performance hit right now (can tweak the parameters later), the goal of eliminating OpenMP is too urgent to linger on small details right now.

oleksiyskononenko:
Well, to me 50% performance drop for 1D doesn't seem small. 1D+2D are dominating visualization time in DAI. `nthreads` won't probably fix it, because airlines is quite big, so we need all the threads. Let me check where the bottleneck is.

PR #1770, opened 2019-04-03
Add mantissa_nbits parameter to FTRL algo

oleksiyskononenko:
Add `mantissa_nbits` parameter to FTRL algo that specifies how many mantissa bits should be used for hashing float values.

### DIFF
--- a/c/models/column_hasher.cc
+++ b/c/models/column_hasher.cc
@@ -22,16 +22,16 @@
 #include "models/column_hasher.h"
 
 
-/*
-*  Abstract Hasher class constructor and destructor.
-*/
+/**
+ *  Abstract Hasher class constructor and destructor.
+ */
 Hasher::Hasher(const Column* col) : ri(col->rowindex()) {}
 Hasher::~Hasher() {}
 
 
-/*
-*  Hash booleans by casting them to `uint64_t`.
-*/
+/**
+ *  Hash booleans by casting them to `uint64_t`.
+ */
 HasherBool::HasherBool(const Column* col) : Hasher(col) {
   values = dynamic_cast<const BoolColumn*>(col)->elements_r();
 }
@@ -45,9 +45,9 @@ uint64_t HasherBool::hash(size_t row) const {
 }
 
 
-/*
-*  Hash integers by casting them to `uint64_t`.
-*/
+/**
+ *  Hash integers by casting them to `uint64_t`.
+ */
 template <typename T>
 HasherInt<T>::HasherInt(const Column* col) : Hasher(col) {
   values = static_cast<const T*>(col->data());
@@ -63,10 +63,10 @@ uint64_t HasherInt<T>::hash(size_t row) const {
 }
 
 
-/*
-*  Hash floats by interpreting their bit representation as `uint64_t`,
-*  also do mantissa binning here.
-*/
+/**
+ *  Hash floats by interpreting their bit representation as `uint64_t`,
+ *  also do mantissa binning here.
+ */
 template <typename T>
 HasherFloat<T>::HasherFloat(const Column* col, unsigned char shift_nbits_in) :
 Hasher(col), shift_nbits(shift_nbits_in) {
@@ -85,9 +85,9 @@ uint64_t HasherFloat<T>::hash(size_t row) const {
 }
 
 
-/*
-*  Hash strings using Murmur hash function.
-*/
+/**
+ *  Hash strings using Murmur hash function.
+ */
 template <typename T>
 HasherString<T>::HasherString(const Column* col) : Hasher(col){
   auto scol = dynamic_cast<const StringColumn<T>*>(col);
--- a/c/models/column_hasher.cc
+++ b/c/models/column_hasher.cc
@@ -20,6 +20,7 @@
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
 #include "models/column_hasher.h"
+#include <float.h>
 
 
 /*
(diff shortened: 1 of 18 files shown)


c/models/column_hasher.cc
```diff
@@ -20,17 +20,18 @@
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
 #include "models/column_hasher.h"
+#include <float.h>
 
 
 /*
-* Abstract Hasher class constructor and destructor.
+*  Abstract Hasher class constructor and destructor.
```
st-pasha: FYI, I usually prefer to indent these comments as
```
/**
 * content...
 */
```
This way if the comment gets big enough it can be collapsed in Sublime Text.
You may, of course, use your own style if you prefer it that way.

c/models/py_ftrl.cc
```diff
@@ -51,28 +52,30 @@ void Ftrl::m__init__(PKArgs& args) {
...
     if (defined_alpha || defined_beta || defined_lambda1 ||
-        defined_lambda2 || defined_nbins || defined_nepochs ||
-        defined_double_precision || defined_negative_class) {
+        defined_lambda2 || defined_nbins || defined_mantissa_nbits ||
+        defined_nepochs || defined_double_precision || defined_negative_class) {
 
       throw TypeError() << "You can either pass all the parameters with "
             << "`params` or any of the individual parameters with `alpha`, "
-            << "`beta`, `lambda1`, `lambda2`, `nbins`, `nepochs`, "
+            << "`beta`, `lambda1`, `lambda2`, `nbins`, `mantissa_nbits`, `nepochs`, "
```
st-pasha: I wonder if we should perhaps allow this use case? `Ftrl(params=my_params, alpha=0.01)` would mean "use these parameters, but override alpha". Would be really easy to implement too: just delete "else" on line 112 below.

c/models/column_hasher.cc
```diff
@@ -20,17 +20,18 @@
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
 #include "models/column_hasher.h"
+#include <float.h>
```
st-pasha: I would avoid including any non-standard headers unless absolutely necessary.
Plus, it would be better to define `DBL_MANT_DIG` as an explicit constexpr anyways: at least then there is no doubt about its value (which should be 52).

tests/models/test_ftrl.py
```diff
@@ -181,6 +190,15 @@ def test_ftrl_construct_wrong_nbins_value():
             == str(e.value))
 
 
+@pytest.mark.parametrize('value, message',
+                         [[-1, "Argument `mantissa_nbits` in Ftrl() constructor cannot be negative: -1"],
+                         [54, "Argument `mantissa_nbits` in Ftrl() constructor should be less than 54, got: 54"]])
```
st-pasha: mantissa is only 52 bits in doubles

tests/models/test_ftrl.py
```diff
@@ -358,6 +385,16 @@ def test_ftrl_set_wrong_nbins_value():
     assert ("Value should be positive: 0" == str(e.value))
 
 
+@pytest.mark.parametrize('value, message',
+                         [[-1, "Integer value cannot be negative"],
+                         [54, "Value should be less than 54, got: 54"]])
```
st-pasha: 0..52

c/models/column_hasher.cc
oleksiyskononenko: Sure, let's convert all the comments to the format Sublime understands. Btw, any reason you sometimes use different format? Like [here](https://github.com/h2oai/datatable/blob/master/c/writebuf.cc#L32-L63), for example.

c/models/py_ftrl.cc
oleksiyskononenko: This is very easy to be done, yes. Though I'm not sure it is a good idea to allow two `alpha`s to be passed, someone may do `Ftrl( alpha=0.01, params=my_params)` and easily be confused with what value is actually used :)

c/models/column_hasher.cc
oleksiyskononenko: If we are sure that on all the platforms, where datatable runs, `DBL_MANT_DIG == 52` then let me define it somewhere in ftrl part.

oleksiyskononenko: Btw, I'm not sure `float.h` is not standard, it is just an [old name](https://en.cppreference.com/w/cpp/header/cfloat) for the standard `<cfloat>`.

tests/models/test_ftrl.py
oleksiyskononenko: Yes, typo when going from `float_nbits` to `mantissa_nbits`, let me fix it.

tests/models/test_ftrl.py
oleksiyskononenko: Same here...

c/models/column_hasher.cc
st-pasha: No reason, just happens >.>

c/models/py_ftrl.cc
st-pasha: Well, it's not actually that ambiguous: the interpretation where first `alpha=0.01` is applied and then overwritten with `params=` would mean the first alpha is completely useless. Whereas if you say `params=` is applied first and then `alpha=`, then the first params= is still useful: it sets up all other parameters.

c/models/column_hasher.cc
st-pasha: This is part of IEEE754 standard for doubles (https://en.wikipedia.org/wiki/Double-precision_floating-point_format). This is true on all platforms where doubles use IEEE754 format; and if they don't, then this approach would not be correct anyways.
Also note that DBL_MANT_DIG is defined as 53, which is not what we want.

oleksiyskononenko: Yes, I noticed that too, not sure why it is 53 :) Anyways, I've already defined it in dt::Ftrl class.

c/models/py_ftrl.cc
oleksiyskononenko: Well, we don't distinguish the order of `alpha` and `params`, so in both examples user ends up with the same set of parameters set for FTRL.  So it still looks ambiguous to me and may confuse people. What you say could already be achieved with just 

```
model = Ftrl(my_params)
model.alpha = 0.01
```

This looks much cleaner to me and doesn't require additional explanation.

c/models/column_hasher.cc
oleksiyskononenko: np :) I converted it to the above format just adding one more space for indentation, as it looks to me that then `*` is better separated from the content. Still ok for Sublime.

PR #1771, opened 2019-04-03
Changes to dt::parallel_* functions

st-pasha:
- `dt::parallel_region`'s argument is now `void()`, i.e. a function without arguments. If a caller does need the current thread's number, they could ask via `dt::get_thread_num()`;
- `dt::parallel_region` now has an optional `nthreads` parameter, specifying how many threads to use for this parallel region. If the value is 0 or above the current number of threads in the pool, all available threads will be used;
- `dt::parallel_for_static` is now a template, sparing the user to write an inner loop `for(i=i0;i<i1;++i)...`.

WIP for #1736


c/parallel/api.h
```diff
@@ -45,12 +45,32 @@ size_t get_hardware_concurrency() noexcept;
...
+template <typename F>
+void parallel_for_static(size_t nrows, F f) {
+  _parallel_for_static(nrows, 4096,
+    [&](size_t i0, size_t i1) {
+      for (size_t i = i0; i < i1; ++i) f(i);
+    });
+}
+
+template <typename F>
+void parallel_for_static(size_t nrows, size_t min_chunk_size, F f) {
```
oleksiyskononenko: Any reason why you created two separate functions here and didn't just do 

```
template <typename F>
void parallel_for_static(size_t nrows, F f, size_t min_chunk_size = 4096)
```

Is it because you wanted `F f` to be always the last parameter?

st-pasha: yes

PR #1791, opened 2019-04-10
Fix dt::barrier(), add tests

oleksiyskononenko:
Fix `dt::barrier()`, when called multiple times, and add corresponding C++ tests.

### DIFF
--- a/c/parallel/ztest_barrier.cc
+++ b/c/parallel/ztest_barrier.cc
@@ -32,18 +32,16 @@ static void test_barrier_1() {
       size_t ithread = dt::this_thread_index();
       data_old[ithread] = ithread + 1;
       dt::barrier();
-      std::vector<size_t>::iterator max = max_element(std::begin(data_old),
-                                                      std::end(data_old));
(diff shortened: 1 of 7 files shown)


c/parallel/ztest_barrier.cc
```diff
@@ -0,0 +1,111 @@
...
+  std::vector<size_t> data_old(nthreads, 0);
+  std::vector<size_t> data_new(nthreads, 0);
+
+  dt::parallel_region(
+    [&]() {
+      size_t ithread = dt::this_thread_index();
+      data_old[ithread] = ithread + 1;
+      dt::barrier();
+      size_t min = *min_element(std::begin(data_old), std::end(data_old));
+      size_t max = *max_element(std::begin(data_old), std::end(data_old));
```
st-pasha: Please change to `std::min_element` / `std::max_element`, otherwise it may fail to compile on some compilers

oleksiyskononenko: Done

PR #1795, opened 2019-04-12
Options functionality revamp

### LINKED ISSUE #1792: Improve options interface
Currently there are multiple problems with `dt.options`:
- the options "live" in C++, but are registered from python, which is very inconvenient. Moreover, most options logically have to be declared & validated in C++.
- each option is returned as a class derived from primitive `int`, `bool`, etc. This is done only for the purpose of being able to call `help(dt.options.option)`. However, a more straightforward mechanism would be better.
- handling of options is done via huge chained-ifs statement in options.cc, which is far from ideal.

The proposal is to switch to a system that is closer to the one used in pandas. The following API is envisioned:
- `dt.options` lists all available options and their values (same as now);
- `dt.options.<opt>` returns the value of option `<opt>`;
- `dt.options.get(opt)` does the same;
- `dt.options.<opt> = x` sets the valud of option `<opt>`;
- `dt.options.set(opt, x)` does the same;
- `del dt.options.<opt>` changes the value of option `<opt>` to default;
- `dt.options.reset(opt)` the same;
- `dt.options.describe(opt)` prints the description of the option: its docstring, and current value;
- `dt.options.context(opt, x)` creates a context in which the value of `<opt>` becomes `x`, and then switches back to its previous value in the end.
- `dt.options.register(o)` will register a new option object. This object should have the following API: `.name` returns the option's name, `.get()` returns the value, `.set(x)` sets the value, `.defaultpd.` returns the default value, `.doc` returns the docstring

st-pasha:
- The declaration of C++-related options moved entirely into C++;

- Added method `dt.options.describe()`, which will print the available options together with their values and descriptions.

 - Added `dt.options.context(option=value)`, which can be used in a with-statement to temporarily change the value of one or more options, and then go back to their original values at the end of the with-block.

 - Added options `fread.log.escape_unicode` (controls treatment of unicode characters in fread's verbose log); and `display.use_colors` (allows to turn on/off colored output in the console).

 - `dt.options` now helps the user when he/she makes a typo: if an option with a certain name does not exist, the error message will suggest the correct spelling.

Closes #1792

### DIFF
--- a/c/parallel/thread_pool.cc
+++ b/c/parallel/thread_pool.cc
@@ -19,7 +19,9 @@
 #include "parallel/thread_pool.h"
 #include "parallel/thread_team.h"
 #include "parallel/thread_worker.h"
+#include "python/_all.h"
 #include "utils/assert.h"
+#include "options.h"
 namespace dt {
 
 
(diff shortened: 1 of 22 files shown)


c/parallel/thread_pool.cc
```diff
@@ -173,5 +175,42 @@ size_t get_hardware_concurrency() noexcept {
...
+    "`std::thread::hardware_concurrency()`. This is usually equal to the\n"
+    "number of available cores.\n"
+    "\n"
+    "You can set `nthreads` to a value greater or smaller than the\n"
+    "initial setting. For example, setting `nthreads = 1` will force the\n"
+    "library into a single-threaded mode. Setting `nthreads` to 0 will\n"
+    "restore the initial value equal to the number of processor cores.\n"
+    "Setting `nthreads` to a value less than 0 is equivalent to\n"
+    "requesting that fewer threads than the maximum.\n");
+}
```
oleksiyskononenko: Are we going to follow the same logic in `dt::parallel_region()`, when number of threads requested is larger than `thpool->size()`? Right now I assume `dt::parallel_region()` won't change number of threads in the pool, and will fall back to the maximum number of threads available. Would prefer that it stays like this, then no need for additional checks in the code.

st-pasha: `dt::parallel_region` will never change the number of threads in the pool. If its parameter `nthreads` is too large, it will be truncated to pool's size.
However if the user is explicitly setting `dt.options.nthreads`, he is free to request as many threads as he'd like. Even 1000000. Though the OS might have some objections to that.

PR #1797, opened 2019-04-17
Add nthreads parameter to dt::parallel_for_static and dt::parallel_for_dynamic

oleksiyskononenko:
Add `nthreads` parameter to `parallel_for_static()` and `parallel_for_dynamic()`, so that when these functions are invoked not from `parallel_region()`, there is a way to adjust number of threads as needed.

Also, add a `dt::num_threads_available()` function, that returns number of threads in the pool, if called from the master thread, and number of threads in a team, otherwise.

### DIFF
--- a/c/parallel/parallel_for_dynamic.cc
+++ b/c/parallel/parallel_for_dynamic.cc
@@ -101,7 +101,7 @@ void dynamic_scheduler::abort_execution() {
 // parallel_for_dynamic
 //------------------------------------------------------------------------------
 
-void _parallel_for_dynamic(size_t nrows, size_t nthreads,
+void parallel_for_dynamic(size_t nrows, size_t nthreads,
                            function<void(size_t)> fn)
 {
   size_t ith = dt::this_thread_index();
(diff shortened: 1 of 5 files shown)


c/parallel/parallel_for_dynamic.cc
```diff
@@ -101,27 +101,44 @@ void dynamic_scheduler::abort_execution() {
...
+void parallel_for_dynamic(size_t nrows, function<void(size_t)> fn) {
+  _parallel_for_dynamic(nrows, dt::num_threads_available(), fn);
+}
+
+void parallel_for_dynamic(size_t nrows,
+                          size_t nthreads,
+                          function<void(size_t)> fn)
+{
+  _parallel_for_dynamic(nrows, nthreads, fn);
+}
```
st-pasha: It seems like there is no difference between this overload and `_parallel_for_dynamic`, then why have both?

oleksiyskononenko: I thought we need `_functions` for some reason, but now I see we really don't. Fixed.

oleksiyskononenko: Btw, when you say

```
// Private
void _parallel_for_static(size_t, size_t, size_t,
                          function<void(size_t, size_t)>);
```
it is kind of "private", because there is nothing that prevents one from calling it. May be we just call it a "driver" or something like that?

st-pasha: yeah, it's python-style "private", i.e. by gentleman's agreement

Issue #1643, opened 2019-02-07
Universal progress bar 


st-pasha:
There should be a progress bar for all long-running tasks

PR #1798, opened 2019-04-18
Added universal progress bar

st-pasha:
This PR adds generic progress bar functionality, available to any C++ executor. The API is still subject to change, in particular the details about nested jobs.

Currently the progress bar will be displayed for any parallel ordered job, including fread, write_csv, string conversions, etc. We'll be adding progress bar support to other functions in future PRs.

Fread's own progress bar was removed; instead it uses the common progress bar now. Consequently, fread's parameters `show_progress` and `progress_fn` are now disregarded (will be removed in the future). 

Added options `progress.enabled`, `progress.callback`, `progress.min_duration` and `progress.updates_per_second`. The "callback" allows the user to supply a custom function that will be called on every progress event.

If the progress function raises an exception, including a KeyboardInterrupt exception, the current operation is stopped. This allows the user to cancel any long-running computation by pressing Ctrl+C; or add an equivalent GUI cancelling mechanism (say, pressing [x] would raise an exception, which would cause the operation to stop).

WIP for #1643


oleksiyskononenko:
Just tried `dt.fread()` on 10GB airlines dataset on Linux with 40 threads, and it takes about 25-30 seconds for progress bar to appear (out of about 1 minute of the total read time). Sometimes even more. What is interesting is that when I do `Ctrl+C` the progress bar appears at non-zero progress:

```
93%  |█████████████████████████████████████▏  | (cancelled)
``` 

It looks like there may be multiple long operations, but not long enough to be considered to display the progress bar.

st-pasha:
@oleksiyskononenko 
There is a logic that prevents progress bar from appearing too early, though it looks like it doesn't work quite properly... Do you see the same behavior if you set `dt.options.progress.min_duration = 0` ? Perhaps I somehow messed up the unit of measurement of the system clock?

oleksiyskononenko:
I did `dt.options.progress.min_duration = 0`, but it doesn't seem to help. Could it be that `fread` does several operations, each of those is not considered long enough to show the progress bar, however, overall they're time consuming on big datasets?

st-pasha:
Hmm, I can confirm this on my Mac too. 
Notably, it happens only on the first read -- all subsequent freads show the progress bar properly from the beginning. Needs more investigating...

oleksiyskononenko:
For me the delay for the largest on the first run, on subsequent runs it just dropped from ~30s to ~10s or something like that. Checked out your latest commits, and now there is no delay. Progress bar comes out at the very beginning of read operation. Great!

PR #1809, opened 2019-04-22
Include debugging information in produced binaries

st-pasha:


ci/setup_utils.py
```diff
@@ -436,26 +436,25 @@ def get_extra_compile_flags():
...
                       "-shared-libasan"]
         elif "DTCOVERAGE" in os.environ:
-            flags += ["-g", "--coverage", "-O0"]
-            flags += ["-DDTTEST"]
+            flags += ["-g2", "--coverage", "-O0"]
+            flags += ["-DDTTEST", "-DDTDEBUG"]
         else:
-            flags += ["-O3", "-g0"]
+            # Optimize at best level, but still include some debug information
+            flags += ["-g2", "-O3"]
```
oleksiyskononenko: May we also need `-ggdb` here?

PR #1810, opened 2019-04-22
removed llvm env setup in dockerfiles

abal5:
Making changes to build and CI files so LLVM is no longer needed.


abal5:
Few things we may want to look into after this is merged:

1. Do we want to also remove the need for LLVM on Mac OSX
2. We'll need to update the Makefile line to reflect an newly updated master image which includes the current changes Line 248 in Makefile https://github.com/h2oai/datatable/blob/7a17aa5f0eeda945a94b43f91d37a73ba94cfe5b/Makefile#L248 but only after these changes appear in master. I can help with this.

PR #1812, opened 2019-04-22
Fix nthreads usage in parallel_for_dynamic, add C++ tests

oleksiyskononenko:
- with @st-pasha's help fix `dt::parallel_for_dynamic()` that didn't account for `nthreads` being less than `num_threads_in_pool()`;
- add C++ tests for `parallel_for_static()` and `parallel_for_dynamic()` that test the above fix.

### DIFF
--- a/tests/test_dt.py
+++ b/tests/test_dt.py
@@ -40,7 +40,8 @@ from tests import same_iterables, list_equals, noop, isview, assert_equals
 # Check if we need to run C++ tests
 #-------------------------------------------------------------------------------
 
-cpp_tests_enabled = hasattr(core, "test_coverage")
+cpp_test = pytest.mark.skipif(not hasattr(core, "test_coverage"),
+                              reason="C++ tests were not compiled")
 
 #-------------------------------------------------------------------------------
 # Prepare fixtures & helper functions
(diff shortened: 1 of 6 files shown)


tests/test_dt.py
```diff
@@ -32,9 +32,15 @@
...
 from datatable.internal import frame_column_rowindex, frame_integrity_check
+from datatable.lib import core
 from tests import same_iterables, list_equals, noop, isview, assert_equals
 
 
+#-------------------------------------------------------------------------------
+# Check if we need to run C++ tests
+#-------------------------------------------------------------------------------
+
+cpp_tests_enabled = hasattr(core, "test_coverage")
```
st-pasha: There is a pytest fixture for this kind of functionality: https://docs.pytest.org/en/latest/skipping.html
```
cpp_test = pytest.mark.skipif(not hasattr(core, "test_coverage"), reason="C++ tests were not compiled")
```
and then mark each test with `@cpp_test`.

oleksiyskononenko: Sure, let me change it in this way.

oleksiyskononenko: Done.

PR #1820, opened 2019-04-26
Replace progress bar in the Aggregator

st-pasha:
Aggregator will now use the standard progress bar functionality.
The previous parameter `progress_fn` is now defunct: if present, it will be ignored with a warning.

WIP for #1643


oleksiyskononenko:
It looks like progress bar doesn't replace existing message properly. Here "Samplinging" probably means "Aggregating" + "Sampling".

```
100% |██████████████████████████████████████████████████| Samplinging
```

PR #1831, opened 2019-04-28
Change the calling convention dt.options.progress.callback function

### LINKED ISSUE #1823: Use named tuple in progress callback function
Current format of the callback function is `callback(progress, status, message)`. It gets difficult to remember which parameter is which, much easier to have `callback(p)`, where `p` has fields `p.progress`, `p.status`, and `p.message`. In addition, when we add new parameters (such as `p.tentative`), user code will not have to change.

st-pasha:
Now the callback function accepts a single argument, `p`, which is a named tuple with 3 fields (more in the future). The fields are `p.progress`, `p.status` and `p.message`, having the same meaning as the three parameters `(progerss, status, message)` before. 

Closes #1823

### DIFF
--- a/tests/models/test_aggregate.py
+++ b/tests/models/test_aggregate.py
@@ -33,15 +33,6 @@ from datatable.internal import frame_column_rowindex, frame_integrity_check
 
 
 #-------------------------------------------------------------------------------
-# Test progress callback function
-#-------------------------------------------------------------------------------
-
-def progress_fn(progress, status, message):
-    assert progress >= 0 and progress <= 1
-    assert status in ("running", "finished", "cancelled", "error")
-
-
-#-------------------------------------------------------------------------------
 # Aggregate 0D
 #-------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


tests/models/test_aggregate.py
```diff
@@ -512,11 +503,21 @@ def aggregate_nd(nd):
     d_in = dt.Frame(matrix)
     d_in_copy = dt.Frame(d_in)
 
+    messages = []
+    def progress_fn(p):
+        assert 0 <= p.progress <= 1
+        assert p.status in ("running", "finished", "cancelled", "error")
```
oleksiyskononenko: Guess we also should add this check here
```
assert p.message in ("Preparing", "Aggregating", "Sampling", "Finalizing")
```

and this one in the bottom
```
assert messages[0].message == "Preparing"
assert messages[-1].message == "Finalizing"
```
to make sure the messages are passed correctly to the `progress_fn`.

oleksiyskononenko: Btw, is `.message` is the same thing that is showing up right next to the progress bar?

st-pasha: Yes, it's the same thing

oleksiyskononenko: Ok, I will update this test shortly then.

oleksiyskononenko: When I print it, It looks like quite a lot of initial messages and the final one are empty strings:

```
[]
...
[]
[Aggregating]
...
[Finalizing]
[]
```

st-pasha: Yes, there is no message at "finished" state, and also we did not setup any message at the initial stages of the job.

PR #1835, opened 2019-04-30
Fix crash when dt.colindex() was used without arguments

### LINKED ISSUE #1834: colindex producing Segmentation Fault core dump
Hi, 

I'm getting a core dump Segmentation Fault when I build a Frame, and then try to use the colindex method.  I did this twice just to make sure.  I'm using Datatable 0.8.0 on Python 3.5.2, on Ubuntu 16.

Here is what I did:
```
import datatable as dt
import numpy as np
a = np.zeros((4,5))
b = dt.Frame(a)
a
>>> array([[0., 0., 0., 0., 0.],
>>>        [0., 0., 0., 0., 0.],
>>>        [0., 0., 0., 0., 0.],
>>>        [0., 0., 0., 0., 0.]])

a[1,1] = 1
a
>>> array([[0., 0., 0., 0., 0.],
>>>        [0., 1., 0., 0., 0.],
>>>        [0., 0., 0., 0., 0.],
>>>        [0., 0., 0., 0., 0.]])

b
>>> C0  C1  C2  C3  C4
>>> ---  --  --  --  --  --
>>>  0    0   0   0   0   0
>>>  1    0   0   0   0   0
>>>  2    0   0   0   0   0
>>>  3    0   0   0   0   0
>>> 
>>> [4 rows x 5 columns]

b[1,2]=2
b
>>> C0  C1  C2  C3  C4
>>> ---  --  --  --  --  --
>>>  0    0   0   0   0   0
>>>  1    0   0   2   0   0
>>>  2    0   0   0   0   0
>>>  3    0   0   0   0   0
>>> 
>>> [4 rows x 5 columns]

b.to_dict()
>>> {'C3': [0.0, 0.0, 0.0, 0.0], 'C0': [0.0, 0.0, 0.0, 0.0], 'C1': [0.0, 0.0, 0.0, 0.0], 'C4': [0.0, 0.0, 0.0, 0.0], >>> 'C2': [0.0, 2.0, 0.0, 0.0]}

b.colindex()
>>> Segmentation fault (core dumped)
```
Thanks for the help.  We're building a long-term dependency on datatable in both R and Python for a multi-language, high-performance implementation of the random forest algorithm called the Arborist.  Thanks again,  
~Hawk Berry

st-pasha:
Closes #1834

### DIFF
--- a/c/frame/names.cc
+++ b/c/frame/names.cc
@@ -139,6 +139,10 @@ name: str
 
 oobj Frame::colindex(const PKArgs& args) {
   auto col = args[0];
+  if (!col) {
+    throw TypeError() << "Frame.colindex() is missing the required "
+                         "positional argument `name`";
+  }
 
   if (col.is_string()) {
     int64_t index = dt->colindex(col.to_pyobj());
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   unicode characters in Jupyter notebook. The error only manifested for
   strings that were longer than 50 bytes in length (#1825).
 
+- Fixed crash when `Frame.colindex()` was used without any arguments, now this
+  raises an exception instead (#1834).
+
 
 ### Changed
 
@@ -178,6 +181,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
   - [Arno Candel][] (#1619, #1730, #1738, #1800, #1803),
   - [Antorsae][] (#1639),
+  - [Hawk Berry][] (#1834),
   - [Jonathan McKinney][] (#1816),
   - [NachiGithub][] (#1789, #1793),
   - [Pasha Stetsenko][] (#1672, #1694, #1695, #1697, #1703, #1705)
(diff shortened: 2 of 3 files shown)


c/frame/names.cc
```diff
@@ -139,6 +139,10 @@ name: str
 
 oobj Frame::colindex(const PKArgs& args) {
   auto col = args[0];
+  if (!col) {
```
oleksiyskononenko: Do you think there is a way to have this kind of check being automatic with some default message for all the positional parameters?

st-pasha: Positional arguments may still be optional, for example consider `range(a) / range(a, b) / range(a, b, c)`. Since we do not "bake" the defaults into the function's signature, it is the application that must inspect whether the arguments where there or not.

That said, we have `args.get<T>(pos)` and `args.get<T>(pos, default)` helper functions to deal with simple cases.

oleksiyskononenko: Yes, sure. May be it is not related to positional/keyword parameters at all, but just to all the parameters. When we specify a list of parameters in `PKArgs`, we can mark a parameter as a required or as an optional one. Then for required parameters the check is done automatically.

st-pasha: How would the API look like?

oleksiyskononenko: May be using `std::map` for defining a list of parameters and specifying if they're required or optional:

```
static PKArgs args_range(
    3, 0, 0, false, false,
    {{"a", PKArgs::REQUIRED}, {"b", PKArgs::OPTIONAL}, {"c", PKArgs::OPTIONAL}}, "range",
    ...
```

then, before the `range` method is called, we walk through this map and do checks for all the `PKArgs::REQUIRED` parameters. Is it possible to be done with the way how `ADD_METHOD` and `ExtType<T>::Methods` work?

st-pasha: Well, in that case you could simply pass the number of required arguments (owing to the fact that the required argument are always in front).

oleksiyskononenko: Yep, one can also re-arrange them and pass the number of required parameters. However, if there is some other types, like PKArgs::REQUIRED_NOT_NONE, or something like that, having a map just makes it more flexible.

PR #1836, opened 2019-04-30
Minor improvement of Aggregator tests

oleksiyskononenko:
Add message checks to progress reporting Aggregator test.

### DIFF
--- a/tests/models/test_aggregate.py
+++ b/tests/models/test_aggregate.py
@@ -518,7 +518,7 @@ def aggregate_nd(nd):
         assert messages[0].progress == 0
         assert messages[0].status == "running"
         assert messages[0].message == "Preparing"
-        assert messages[-2].progress < 1.0
+        assert messages[-2].progress <= 1.0
         assert messages[-2].status == "running"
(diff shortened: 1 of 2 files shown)


tests/models/test_aggregate.py
```diff
@@ -516,8 +517,13 @@ def progress_fn(p):
                                              nd_max_bins=div, seed=1)
         assert messages[0].progress == 0
         assert messages[0].status == "running"
+        assert messages[0].message == "Preparing"
+        assert messages[-2].progress < 1.0
```
oleksiyskononenko: @st-pasha for a running task, can progress be equal to 1.0 or will always be < 1.0?

st-pasha: It can. In fact, it must always reach 1.0 before switching to "finished", so you're quite likely to see progress message (1.0, "running") right before (1.0, "finished")

oleksiyskononenko: Yeah, ok, let me make it `<= 1.0` then.

PR #1852, opened 2019-05-13
Re-factor FTRL options handling and Python setters 

oleksiyskononenko:
- as FTRL model now getting more and more options, we re-factor their handling in FTRL. Namely, now all options are handled (validated and set) in the corresponding setters only. This makes it much easier to support;
- added an `interactions` parameter to FTRL constructor, before we only had `.interactions` property for that;
- with @st-pasha's help start using `py::Arg` instead of `py::robj` in Python setters. This improves flexibility of error messages and makes them more user friendly;
- improve FTRL test coverage, and adjust error messages according to the above changes.


c/progress/_options.cc
```diff
@@ -67,7 +67,7 @@ static py::oobj get_ups() {
 
 static void set_ups(py::oobj value) {
   double x = value.to_double();
-  py::Validator::check_positive(x, value);
+  py::Validator::check_positive(x, {value, "updates_per_second"});
```
st-pasha: If you look in the definition of `config_option::set()` (file "options.cc"), you'll see that it is defined as
```
void config_option::set(const PKArgs& args) { 
    setter(args[0].to_oobj()); 
}
```
Here `setter` is the option-setter function specified when the option was declared. For example, `set_ups()` is one such setter function; `set_min_duration()` is another. 
Inside `config_option::set()`, the name of the option is already known: it is `config_option::name` (unfortunately, it is not the same as `args[0]`). Hence, there should be no need to pass it along explicitly: the signature of the setter function can be updated so that it takes a `const Arg&` instead of an `oobj`.

oleksiyskononenko: Thanks, this has been fixed.

PR #1863, opened 2019-05-25
Fix split_into_nhot(), so that it encodes NA's into NA's

oleksiyskononenko:
- `split_into_nhot()` function encoded NA's into zeros, i.e. negatives. In this PR we fix it, so that NA's are encoded into NA's;
- minor change in FTRL to make it consistent with the new `split_into_nhot` output;
- fixed segfault when `split_into_nhot()` was called on `None`;
- add and adjust corresponding tests.

### DIFF
--- a/c/str/py_str.cc
+++ b/c/str/py_str.cc
@@ -29,6 +29,8 @@ static PKArgs args_split_into_nhot(
 
 static oobj split_into_nhot(const PKArgs& args) {
   DataTable* dt = args[0].to_datatable();
+  if (dt == nullptr) return py::None();
+
   std::string sep = args[1]? args[1].to_string() : ",";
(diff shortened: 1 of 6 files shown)


c/str/py_str.cc
```diff
@@ -29,6 +29,8 @@ static PKArgs args_split_into_nhot(
 
 static oobj split_into_nhot(const PKArgs& args) {
   DataTable* dt = args[0].to_datatable();
+  if (dt == nullptr) return py::None();
```
st-pasha: what about the case when the function is called without any arguments? we should probably have such test too

oleksiyskononenko: Sure, fixed this one too

PR #1873, opened 2019-06-06
Use semaphores to handle thread sleep cycle

st-pasha:
Right now we're using condition variables to notify the thread workers when they need to wake up and do some job. The condition variables are standard in C++. However, this process has some problems. The standard condition-variable wait cycle looks like this:
```
// In thread workers:
std::unique_lock<std::mutex> lock(mutex);
while (!next_work) {
  cond_var.wait(lock);
}

// In master thread:
{ 
  std::lock_guard<std::mutex> lock(mutex);
  next_work = ...;
}
cond_var.notify_all();
```
The semantic of the condition variable is that it takes a reference to a `lock` variable; then the `.wait()` call temporarily unlocks it while waiting, but then must **reacquire** the lock after it stops waiting. This means that after `.notify_all()` all worker threads will be scheduled by the OS for execution almost simultaneously, and all of them will attempt to reacquire the mutex at that point. To make matters worse, the condition variable only works with standard mutexes, so we cannot use a spin-lock mutex there. The end result is that all threads are fighting over the mutex which was not even designed for such high-contention use case: threads that fail to acquire the mutex may get de-prioritized/temporarily put to sleep.

The new approach avoid condition variables altogether, and instead uses semaphores, which are absent from the standard library, but can be implemented relatively easily using OS-native methods. Adding the "light sleeping" stage, the median wake-up time can be reduced to just few microseconds (although it could still be up to 10-15ms in the worst case)

### DIFF
--- a/c/parallel/semaphore.h
+++ b/c/parallel/semaphore.h
@@ -202,10 +202,15 @@ class LightweightSemaphore {
     }
 
     void wait() {
-      int spin_count = 10000000;
+      int spin_count = 1000000;
       do {
         if (try_wait()) return;
-        std::atomic_signal_fence(std::memory_order_acquire);
+        // This line significantly boosts the performance.
+        // When a thread busy-waits, it consumes its CPU quota, so that when
+        // the "real" work arrives, the thread my receive a priority penalty
+        // from the OS (especially when the total number of threads is equal
+        // to the number of cores in the system).
+        std::this_thread::yield();
       } while (spin_count--);
 
       int old_count = m_count.fetch_sub(1, std::memory_order_acquire);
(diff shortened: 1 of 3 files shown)


c/parallel/semaphore.h
```diff
@@ -202,10 +202,15 @@ class LightweightSemaphore {
...
 
     void wait() {
-      int spin_count = 10000000;
+      int spin_count = 1000000;
       do {
         if (try_wait()) return;
-        std::atomic_signal_fence(std::memory_order_acquire);
+        // This line significantly boosts the performance.
+        // When a thread busy-waits, it consumes its CPU quota, so that when
+        // the "real" work arrives, the thread my receive a priority penalty
```
oleksiyskononenko: Probably a typo: "my" -> "may"

st-pasha: yep

PR #1879, opened 2019-06-15
Fix segfault in progress reporting

oleksiyskononenko:
In some cases `progress_manager` tries to access progress bar that is already deleted by `progress_manager::finish_work()` method, and this results in a segfault. In particular, some of the aggregator tests, that also tested progress reporting, segfaulted on Linux and PPC.

To fix it, we use `std::mutex` to protect progress bar access.

### DIFF
--- a/c/progress/manager.cc
+++ b/c/progress/manager.cc
@@ -34,6 +34,7 @@ progress_manager::progress_manager()
 void progress_manager::start_work(work* task) {
   if (tasks.empty()) {
     xassert(pbar == nullptr);
+    std::lock_guard<std::mutex> lock(mutex);
     pbar = new progress_bar;
     task->init(pbar, nullptr);
   } else {
--- a/c/progress/manager.cc
+++ b/c/progress/manager.cc
@@ -57,17 +57,17 @@ void progress_manager::finish_work(work* task, bool successfully) {
 }
 
 
-void progress_manager::update_view() {
+void progress_manager::update_view() const {
   xassert(dt::this_thread_index() == size_t(-1));
-  if (!pbar) return;
   std::lock_guard<std::mutex> lock(mutex);
+  if (!pbar) return;
   pbar->refresh();
 }
 
 
 void progress_manager::set_error_status(bool cancelled) noexcept {
-  if (!pbar) return;
   std::lock_guard<std::mutex> lock(mutex);
+  if (!pbar) return;
   try {
     pbar->set_status_error(cancelled);
   } catch (...) {}
--- a/c/progress/manager.cc
+++ b/c/progress/manager.cc
@@ -59,10 +59,9 @@ void progress_manager::finish_work(work* task, bool successfully) {
 
 void progress_manager::update_view() {
   xassert(dt::this_thread_index() == size_t(-1));
-  if (pbar) {
-    std::lock_guard<std::mutex> lock(mutex);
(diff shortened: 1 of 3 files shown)


c/progress/manager.cc
```diff
@@ -49,23 +49,25 @@ void progress_manager::finish_work(work* task, bool successfully) {
...
 
-void progress_manager::update_view() const {
+void progress_manager::update_view() {
   xassert(dt::this_thread_index() == size_t(-1));
-  if (pbar) {
-    pbar->refresh();
-  }
+  if (!pbar) return;
+  std::lock_guard<std::mutex> lock(mutex);
+  pbar->refresh();
```
st-pasha: `pbar` may still become null while the mutex waits to be locked

c/progress/manager.h
```diff
@@ -43,9 +44,10 @@ class progress_manager {
     // while more and more top-level tasks are received.
     progress_bar* pbar;
     std::stack<work*> tasks;
+    std::mutex mutex;
 
   public:
-    void update_view() const;
+    void update_view();
```
st-pasha: `update_view()` is still conceptually a `const` method: it isn't supposed to change the state of the `progress_manager` instance. So you could either declare `mutex` as `mutable` (I intend all my puns), or even make it a static file-level variable.

c/progress/manager.h
```diff
@@ -43,9 +44,10 @@ class progress_manager {
     // while more and more top-level tasks are received.
     progress_bar* pbar;
     std::stack<work*> tasks;
+    std::mutex mutex;
```
st-pasha: Could you please add a comment about the purpose of this mutex: what it's supposed to protect, otherwise it's rather hard to verify that it is used in all places where its supposed to and not anywhere else.

c/progress/manager.cc
oleksiyskononenko: Right, the same if true for `set_error_status()`, I will just swap mutex line with the nullptr check in these two methods.

c/progress/manager.h
oleksiyskononenko: If you want to keep this method `const`, I guess it is better to declare `mutex` as mutable and keep it as a part of the class for the purpose of encapsulation.

c/progress/manager.h
oleksiyskononenko: Sure.

oleksiyskononenko: Done.

c/progress/manager.cc
st-pasha: I'm not so sure, which is why I added a comment about the purpose of `mutex`. 
As far as I can see, `update_view()` is called from the monitor thread, and only from the monitor thread. All the other functions are called from the master thread only, presumably while the thread pool is sleeping (though the monitor thread is not in a thread pool, so its sleep cycle may be slightly out of sync).
The set_error_status() is interesting because it is called from `exception_to_python()`, but presumably we only enter python when the thread pool is already asleep.

These kind of assumptions tend to prove brittle over time though. Who knows, maybe in the next iteration we'll have the threadpool still working while the control is yielded to python -- in order to remember to verify these kind of assumptions, they should be encoded in form of asserts.

st-pasha:
Also, we need to think hard here whether we are creating a potential for a deadlock with this mutex...

PR #1881, opened 2019-06-19
Attempt to prevent crashes connected with monitor thread and GIL

st-pasha:
This may or may not resolve current problems with PPC tests.


oleksiyskononenko:
Thanks, let me test it. Though there is no guarantee I can catch a segfault, for some reason the behavior is kind of different between interactive test in a docker and what CI does.

oleksiyskononenko:
Looks good, let's try to merge it and see if it helps getting master build green.

PR #1913, opened 2019-07-12
Single-column selector DT[col]

### LINKED ISSUE #1910: Consider restoring DT[col] api
In #1395 this API was deprecated and removed. However, I'm wondering whether it was a good judgment. In fact, we have 3 choices for `DT[a]`: treat `a` as a column selector ("option 1"), treat it as a row selector ("option 2"), or make it an error ("option 3"). Currently this form raises an exception (so option 3), but the message says it will be treated as row-selector in 0.9 (option 2).

The best way to think about this problem is in terms of examples, so let's start:

- `DT[0]`: this could mean either the first row, or the first column. Generally, I think, selecting columns is much more useful (and faster) than selecting individual rows, giving more weight to option 1. However when the index value is large `DT[1000000]`, people tend to think that it refers to the row number. Obviously, we can't make the interpretation dependent on the magnitude of the value. 

- When the integer index is in a variable with a neutral name: `DT[i]`, the expression more naturally looks like a column selector than a row selector.

- Since iteration over a Frame produces columns, it seems natural for `DT[i]` to mean the same as `list(DT)[i]`, meaning option 1.

- `DT[:10]` -- does this mean the first 10 rows, or the first 10 columns? This better be consistent with the meaning of an integer selector. However, in this case the row-interpretation (option 2) comes to mind easier, especially when the numbers are large: `DT[:1000000]` looks like the first 1M rows rather than columns. 

- `DT["id"]` -- under option 1 this selects column with the name "id", whereas under option 2 this is not valid (unless maybe if DT is keyed by a string column, in which case this could mean selecting a row with index "id"). This form strongly favors option 1, especially since this syntax is natural to people who have experience with pandas.

- `DT[:]` means "select everything" both under option 1 and option 2, so this is unambiguous.

- `DT[None]` could also mean "select everything", or it could be an error, or it could mean "select nothing and return None". The latter option may be useful when trying to select a particular column (or row?) dynamically, and in some cases you want to not select anything.

- `DT[f.A > 0]` is not valid under option 1, but becomes a row filter under option 2. Obviously, under option 1 one could write `DT[f.A>0, :]`.

- Allowing column selector `DT["id"]` reinforces the view of a Frame as an ordered dictionary of columns; in particular it allows syntax `dict(DT)` and `**DT` (similar to `*DT` which is already allowed). In particular, cbinding two frames could be written as `dt.Frame(**DT1, **DT2)`.

In summary, it seems to me that interpreting a single-element selector as _column_ is more useful than interpreting it as a _row_. 

Thoughts / feedback welcome.

st-pasha:
The expression `DT[a]` is now valid when `a` is either an integer or a string, and it is interpreted as a column selector from `DT`. This is not a breaking API change, however previously expression `DT[a]` produced an error message saying that it will eventually be interpreted as row-selection.

Few other changes:
- added C++ method `Frame::oframe(DataTable*)->oobj` to simplify creation of Frame objects;
- A single-column Frame can now be used as a source for constructing a frame, e.g.: `dt.Frame(A=DT)` where DT is single-column;
- Added "hidden" Frame method `.keys()`, which returns just the tuple of column names. This method is needed to satisfy the requirements of python mapping interface;
- `**DT` can now be used -- expands into a sequence of columnname=column named arguments. For example, cbinding can be equivalently written as `dt.Frame(**DT1, **DT2)` (however if the frames have duplicate names, an exception will be raised).

Closes #1910

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -200,6 +200,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Removed deprecated Frame methods `.topython()`, `.topandas()`, `.tonumpy()`,
   and `Frame.__call__()`.
 
+- Syntax `DT[col]` has been restored (was previously deprecated in 0.7.0),
+  however it works only for `col` an integer or a string. Support for slices
+  may be added in the future, or not: there is a potential to confuse
+  `DT[a:b]` for a row selection. A column slice may still be selected via
+  the i-j selector `DT[:, a:b]`.
+
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Frame can now be treated as an iterable over the columns. Thus, a Frame
   object can now be used in a for-loop, producing its individual columns.
 
+- A Frame can now be treated as a mapping; in particular both `dict(frame)`
+  and `**frame` are now valid.
+
+- Single-column frames can be be used as sources for Frame construction.
+
 
 ### Fixed
 
--- a/c/csv/py_csv.cc
+++ b/c/csv/py_csv.cc
@@ -36,8 +36,7 @@ static oobj read_csv(const PKArgs& args)
   robj pyreader = args[0];
   GenericReader rdr(pyreader);
   std::unique_ptr<DataTable> dtptr = rdr.read_all();
-  return oobj::from_new_reference(
-          Frame::from_datatable(dtptr.release()));
+  return Frame::oframe(dtptr.release());
 }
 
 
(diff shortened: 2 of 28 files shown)


CHANGELOG.md
```diff
@@ -92,6 +92,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Frame can now be treated as an iterable over the columns. Thus, a Frame
   object can now be used in a for-loop, producing its individual columns.
 
+- A Frame can now be treated as a mapping; in particular both `dict(frame)`
+  and `**frame` are now valid.
+
+- Single-column frames can be be used as sources for Frame construction.
```
oleksiyskononenko: A typo: "can be be" -> "can be".

PR #1927, opened 2019-07-23
Gzip compression when writing csvs

st-pasha:
Csv writer method (`Frame.to_csv()`) now supports parameter `compression="gzip"`, producing an output compressed using the gzip format. For a large file, the output will contain multiple flush points, allowing to be decompressed in parallel (provided the reader is smart enough).

The compression method can also be detected automatically, for example the following will work:
```
frame.to_csv("out.csv.gz")  # creates a gzip file
frame.to_csv("out.csv")  # creates an uncompressed file
```


st-pasha:
Looks like the build process became slightly more complicated, in view of this I think we should postpone merging this after the release 0.9 is out.

Issue #1396, opened 2018-11-08
Column should be converted to a value class


st-pasha:
Currently, "Column" objects are passed as C-style pointers: `Column*`.  This is because Column is abstract, and the actual class that is instantiated is one of `BoolColumn`, `IntColumn`, `StringColumn`, etc.

However, this presents a big problem: the ownership of this pointer is unclear, and the semantics of copying a Column object makes its use error-prone.

This issue suggests replacing current `Column` class with the hierarchy similar to `RowIndex`: the base `Column` class contains only an `impl` pointer, where that pointer has a shared-ptr semantics, and is equivalent to the current "Column" class.

This will have the following benefits:
- The lifetime of `Column` class will be automatically managed (for example, `std::vector<Column>` will be able to destruct safely), no possibility for resource leaks;
- In rare cases, the `impl` within the `Column` can be replaced with another `impl` transparently. For example, sometimes `StringColumn<uint32_t>` may need to be replaced with `StringColumn<uint64_t>`, as in #574;
- The internal structure of the `Column` class will be encapsulated better.

-------

- [x] Create basic `OColumn` ("owner" column) class;
- [x] Convert `DataTable.columns` into `std::vector<OColumn>`;
- [x] Implement `get_element()` API in `OColumn`;
- [x] Upgrade `Stats` class, so that individual stats can be computed without casting;
- [x] Eliminate all instances where external classes are expecting `Column*` as input;
- [x] Use refcount-based copying for `OColumn` (faster than shallow copy) (#2010);
- [x] Eliminate all `OColumn::get()` calls;
- [x] Eliminate all `Column::operator->()` calls, create a well-defined API for `OColumn` (#2012);
- [x] Remove `RowIndex` from `Column`, instead a column with rowindex should become virtual (#2011);
- [x] Merge `expr::virtual_column` class with `OColumn`;
- [x] Rename `Column -> ColumnImpl` and `OColumn -> Column`;
- [x] Remove `Column::mbuf` (#2052);

st-pasha:
This is a prerequisite for many other feature requests, so assigning a high priority.

st-pasha:
Pull request #2057 finished the last task within this epic. The epic is now complete!   🎉

PR #1945, opened 2019-08-01
Hashers now use OColumn's ::get_element() interface

st-pasha:
The number of classes is reduced and API simplified. This class is now future-proof with respect to new virtual column classes that will be added soon.

WIP for #1396

### DIFF
--- a/c/models/column_hasher.cc
+++ b/c/models/column_hasher.cc
@@ -20,46 +20,25 @@
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
 #include "models/column_hasher.h"
+#include "utils/assert.h"
 
 
 /**
  *  Abstract Hasher class constructor and destructor.
  */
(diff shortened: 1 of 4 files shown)


c/models/column_hasher.cc
```diff
@@ -20,46 +20,25 @@
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
 #include "models/column_hasher.h"
+#include "utils/assert.h"
```
oleksiyskononenko: Any reason to include `assert.h`?

st-pasha: I added an `xassert()` at some point to test the changes, but then forgot to remove the header after the assert was no longer necessary.
So, not really needed, can be removed anytime (but at least it doesn't hurt).

oleksiyskononenko: Yeah, that's what I thought.

PR #1947, opened 2019-08-02
Add progress reporting to Ftrl model

oleksiyskononenko:
- add progress reporting to FTRL `fit()` and `predict()` methods;
- add `clear_on_success` option to `dt.options.progress` to control if progress bar is kept on a screen when job finished or not.

WIP for #1643


c/models/dt_ftrl.cc
```diff
@@ -585,6 +586,7 @@ FtrlFitOutput Ftrl<T>::fit(T(*linkfn)(T), U(*targetfn)(U, size_t), T(*lossfn)(T,
 
   double epoch_stopped = static_cast<double>(iteration_end) / dt_X_train->nrows;
   FtrlFitOutput res = {epoch_stopped, static_cast<double>(loss)};
+  job.set_done_amount(total_work_amount);
```
st-pasha: The documentation for `work` class says
> At the end of `work`'s lifetime the user is expected to call
> `work.done()`. This will check whether the amount of work reported
> is the same as was planned, and update the progress bar. The `work`
> instance must not be used after this call.
By setting the `total_work_amount` right in the end you are defeating the check, so that if the workers have under-reported the amount of work they did, this fact would be hidden from us.

PR #1952, opened 2019-08-07
Restructure Stats class to use the new OColumn API

st-pasha:
Refactoring work on `OColumn` class, including:
- clean-up of `DataTable::verify_integrity()`;
- `Column::set_elem()` removed;
- Use new OColumn API in repr functions;
- Use new OColumn API in cast functions.

The largest part of this PR consists of refactoring the `Stats` class. The highlights of this effort are:
- `OColumn` class gets new methods `.stats()` and `.get_stats_if_exist()`, which return a pointer to the internal `Stats` object. This object can then be queried to obtain different column stats without the need to cast either Stats or OColumn into any subclass (which will be ill-defined in the future);
- The `Stats` class now carries a back-reference to its parent `Column` class, allowing the stats to be computed on demand without the need to pass `Column` pointer;
- Individual stats can be queried on the `Stats` object in 5 different ways, accommodating a variety of use-cases.
- Each stat now has 2 state indicators: computed, and valid. The "computed" flag has existed before, whereas "valid" is new. When a stat is "invalid", its value should be treated as NA.
- When a stat is queried that is not supported for the underlying column type (e.g. mean for a string column), then an "invalid" indicator will be returned without raising an error.

External API change:
- `Frame.sum()` now returns a value of type `double`, even for integer columns. This is done to guard against possible overflows when summing a large number of values.

WIP for #1396


oleksiyskononenko:
When we save frame to an external .jay file, what would happen to stats?

st-pasha:
Some of them (nacount/min/max) are saved, and then later restored when the file is opened. The rest of the stats perish.

oleksiyskononenko:
So If I 
- calculate min/max for a frame
- save this frame to a jay file
- open the jay file
- do min/max on this frame again
the last step won't trigger any calculations and should be almost immediate?

st-pasha:
yes

PR #1956, opened 2019-08-09
Adjust the way FTRL handles infinities

oleksiyskononenko:
- make FTRL model to ignore infinite targets during training and validation, infinite values in a dataset itself are still ok as they can be hashed based on their float representation;
- minor re-factoring of py::Validator methods: separate infinity check and simplfy model validation for pickling/unpickling;
- adjust several FTRL tests to check what happens with inf values/targets.

Could help in fixing https://github.com/h2oai/h2oai/issues/10413#issuecomment-519641680


st-pasha:
I suggest rebasing on top of master, to avoid errors with zlib.h

PR #1968, opened 2019-08-20
Export_names() method

### LINKED ISSUE #1866: export_names() function
This is a proposal to add the following new `Frame` method:
```
DT.export_names(include=None, exclude=None, prefix=None)
```
This would create a set of local variables, referencing the columns of the frame `DT`. For example, if `DT` has columns `A`, `B` and `C`, then after calling `DT.export_names()` 3 corresponding local variables will be created, so that instead of writing `DT[f.A < f.B, :]` we could write simply `DT[A < B, :]`.

The options `include`/`exclude` allow better control over which column names are exported, and optional `prefix` allows appending a certain prefix to each column name. We could also allow this to be used as a context manager, so that the new variables would get automatically deleted in the end.

There could also be an option for what to do if a variable with the same name as one of the columns already exists in the local scope (error? warning? skip? overwrite? mangle?).

Please let me know whether this sounds useful.

st-pasha:
It appears there is no way to modify the set of local variables in a function: the python makes determination whether a variable is local or global at compile time. As a consequence, the `export_names()` method exports into the global namespace.
 
Closes #1866


c/frame/py_frame.cc
```diff
@@ -100,6 +100,51 @@ oobj Frame::copy(const PKArgs&) {
...
+Create local variables for each column of this frame.
+
+This method inserts into `globals()` as many new variables as there are
+columns in this frame. Each variable is an f-expression referring to
+its column. For example, if the frame has columns A, B, and C, then
+three new variables will be created A = f.A, B = f.B and C = f.C.
+After calling this method you will be able to write column expressions
+using the column names directly, without using the f symbol:
+
+    DT[A + B > C, :]
```
oleksiyskononenko: What if `A` is a valid column name, but not valid as a Python variable name? Something like `1abc`.

c/frame/py_frame.cc
```diff
@@ -100,6 +100,51 @@ oobj Frame::copy(const PKArgs&) {
...
+        globals()[name] = f[name]
+)");
+
+void Frame::export_names(const PKArgs&) {
+  py::oobj f = py::oobj::import("datatable", "f");
+  py::odict globals = py::robj(PyEval_GetGlobals()).to_pydict();
+  py::otuple names = dt->get_pynames();
+  for (size_t i = 0; i < dt->ncols; ++i) {
+    py::robj name = names[i];
+    globals.set(name, f.get_item(name));
```
oleksiyskononenko: I suppose this will overwrite the variable with the same name that already exists in the global scope? May be we should mention this in the documentation.

c/frame/py_frame.cc
st-pasha: You'll have a variable `1abc` in your global scope, though good luck trying to actually use that variable :)

c/frame/py_frame.cc
st-pasha: It is mentioned in line 125 above:
> If a variable with the same name's as one of the columns already
exists in the globals, it will be overwritten. If the same variable
exists locally, it will remain as-is.

c/frame/py_frame.cc
oleksiyskononenko: Yes, I noticed that. May be worth to mention it in the docs...

st-pasha: Nah, those variables are perfectly harmless. 
What could be problematic is when you accidentally overwrite variable `print`, or `str`, or `open`, etc. 
However, this function expects you to be somewhat aware of what columns you have in your frame, and use it only when it's safe to do so.

oleksiyskononenko: What I mean is writing something like that:

```
After calling this method you will be able to write column expressions
using the column names directly (of course, if the column names are also valid python identifiers):
```

PR #1969, opened 2019-08-21
Further refactoring into the new OColumn API

st-pasha:
Fixed handling of string columns in :
- sort
- aggregator
- expr_str
- expr_binaryop

WIP for #1396


c/models/aggregator.cc
```diff
@@ -644,26 +620,14 @@ void Aggregator<T>::group_2d_categorical_str() {
...
-void Aggregator<T>::group_2d_mixed() {
-  switch (dt_cat->get_ocolumn(0).stype()) {
-    case SType::STR32:  group_2d_mixed_str<uint32_t>(); break;
-    case SType::STR64:  group_2d_mixed_str<uint64_t>(); break;
-    default:            throw ValueError() << "For 2D mixed aggretation, the categorical column "
-                                           << "type should be either STR32 or STR64";
+void Aggregator<T>::group_2d_mixed()
+{
+  const OColumn& col0 = dt_cat->get_ocolumn(0);
+  const auto& col1 = *contconvs[0];
```
oleksiyskononenko: Do you have any plans to do cast on the fly with `get_element()` that we discussed? I'm looking at this code of the column convertor

```
template<typename T1, typename T2>
T2 ColumnConvertorReal<T1, T2>::operator[](size_t row) const {
  // size_t i = row;
  size_t i = this->ri[row];
  if (i == RowIndex::NA  || ISNA<T1>(values[i]) ) {
    return GETNA<T2>();
  } else {
    // return values[i];
    return static_cast<T2>(values[i]);
  }
}
```

and it seems that everything, except for the casting, is now handled by `get_element()`.

st-pasha: Generally, this is a typical "type-cast" column.

We could, of course, implement multiple `get_element()` interfaces for a single column, effectively merging type-casting into the column itself, but this will necessarily cause a lot of duplication in functionality. Still might be worth it from the efficiency standpoint, if type-casts are sufficiently frequent.

st-pasha: Currently, each `OColumn` class only needs to implement a single `get_element()` call: the one most suitable for its stype.
If we want to expand this, there are 2 possibilities:
- require that all `OColumn`s implement casts into all types. This may be too much work though, considering that there will be multiple classes implementing each stype.
- implement only some of the casts (depending on what we feel is most useful), and for the other casts automatically fall back to the "standard" type-cast OColumn. This would require some mechanism for signaling which get-elements are implemented natively in each class.

oleksiyskononenko: Yes, I definitely like the second option.

PR #1985, opened 2019-09-06
Progress fix

### LINKED ISSUE #1937: progress bar does not stop when Ctrl+C is pressed
Emulating Ctrl+C sent to a process is not exactly trivial, but this presumably does the trick:
```
import signal
import subprocess
import time
def test():
    proc = subprocess.Popen(["python", "-c", "import datatable as dt; DT = dt.fread('/very/large/file')"],
                            stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    time.sleep(0.1)
    proc.send_signal(signal.Signals.SIGINT)
    ret = proc.wait()
    print("Return code = %r" % ret)
    print("STDOUT = %r" % proc.stdout.read().decode())
    print("STDERR = %r" % proc.stderr.read().decode())

test()
```

oleksiyskononenko:
- fix progress bar not stopping when Ctrl+C is pressed;
- make parallel infrastructure more interrupt friendly;
- improve progress bar reporting;
- add tests, progress bar may need some additional tests though;
- adjust progress reporting for FTRL model.

Closes #1937


c/parallel/parallel_for_ordered.cc
```diff
@@ -300,23 +300,29 @@ void ordered::parallel(function<void(size_t)> pre_ordered,
                        function<void(size_t)> post_ordered)
 {
   if (sch->n_threads <= 1) {
+    enable_monitor(true);
```
st-pasha: What if there is an exception during the ordered loop execution -- will the monitor remain enabled, potentially later causing a crash in python?

c/parallel/api.h
```diff
@@ -101,7 +112,7 @@ void barrier();
  * Run parallel loop `for i in range(nrows): f(i)`, with dynamic scheduling.
  */
 void parallel_for_dynamic(size_t nrows, std::function<void(size_t)> fn);
-void parallel_for_dynamic(size_t nrows, size_t nthreads,
+void parallel_for_dynamic(size_t nrows, NThreads _NThreads,
```
st-pasha: I don't think that's a valid variable name in C++, see https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

c/parallel/ztest_parallel_for.cc
```diff
@@ -102,7 +102,7 @@ void test_parallel_for_ordered(size_t n) {
 
   dt::parallel_for_ordered(
     /* n_iters = */ n,
-    /* n_threads = */ size_t(-1),
+    /* n_threads = */ dt::NThreads(size_t(-1)),
```
st-pasha: this is same as `dt::NThreads()`

c/progress/progress_bar.cc
```diff
@@ -32,65 +32,71 @@ static void init_status_strings() {
...
 
+progress_bar::~progress_bar() {}
 
-progress_bar::progress_bar() {
+
+void progress_bar_disabled::set_progress(double, double) noexcept {};
+void progress_bar_disabled::set_status_finished() {};
+void progress_bar_disabled::set_status_error(bool) {};
+void progress_bar_disabled::set_message(std::string&&) {};
+void progress_bar_disabled::refresh() {};
```
st-pasha: semicolons not needed after function bodies

c/parallel/api.h
oleksiyskononenko: Changed it everywhere to `NThreads_` to be safe.

c/parallel/ztest_parallel_for.cc
oleksiyskononenko: Yeah, didn't notice this was just a way to get all the available threads. Expected to see something like `dt::num_threads_in_pool()` here. Anyways, simplified this to `dt::NThreads()`.

c/progress/progress_bar.cc
oleksiyskononenko: Not sure how they got there :) Fixed.

c/parallel/parallel_for_ordered.cc
oleksiyskononenko: Yes, could be a problem. Also for `parallel_for_static()` when we don't start a parallel region. Can add `try`/`catch` around the `func(i)` call and stop monitor thread if there is an exception, but not sure what to do with the exception itself in this case.

st-pasha: well, you can always just re-throw it

oleksiyskononenko: Sounds like a plan :)

st-pasha: I moved try-catch blocks out of the loop: seem more robust that way.

c/progress/progress_manager.h
```diff
@@ -53,6 +56,20 @@ class progress_manager {
...
+    // the interrupt, it also resets `caught_interrupt` back to `false`.
+    mutable bool caught_interrupt;
+
+    // This flag is kind of similar to the above, but is atomic and not
+    // cleared by the interrupt handler. It is only cleared when threads are
+    // joining, and used in the cases when there is no special scheduler
+    // (like in the cases of `parallel_for_static`, `nested_for_static`
+    // or a single threaded `parallel_for_ordered`) but we still want
+    // to abort execution of all threads when SIGINT is received.
+    mutable std::atomic<bool> abort_execution;
```
st-pasha: I'm still not sure I understand the purpose of this flag fully.
I see that it is set together with `caught_interrupt`, but cleared at a later stage. But then, why do we need to clear `caught_interrupt` so early? Would it work if we removed `caught_interrupt` altogether, and left just this flag?
What about other, non-interrupt exceptions?

c/parallel/parallel_for_ordered.cc
oleksiyskononenko: Yes, though I'm not sure if anything besides the `*_ordered()` functions can throw. May still be a little bit better from the performance point of view.

c/progress/progress_manager.h
oleksiyskononenko: Replaced these two flags with just one, hopefully its purpose is clear now.

Issue #1979, opened 2019-08-26
Implement row-wise functions


st-pasha:
Now that we have columnsets (#1962), we can build the functionality for functions such as

- [x] `rowall()`
- [x] `rowany()`
- [x] `rowcount()`
- [x] `rowsum()`
- [x] `rowmean()`
- [x] `rowsd()`
- [x] `rowmin()`
- [x] `rowmax()`
- [x] `rowfirst()`
- [x] `rowlast()`

PR #1992, opened 2019-09-10
[WIP] New expression evaluation pipeline

### LINKED ISSUE #1980: Evaluation pipeline should support multiple inputs/outputs
This is a prerequisite for #1979 .
Will also help with implementing some of the missing functions from math module where two outputs are produced.

st-pasha:
This PR adds new `Expr` class that will replace previous `pexpr`. The main features of this class are:
- Each `Expr` object is used by-value (encapsulates the pimpl idiom);
- Each `Expr` contains the "head" part, and the "args". The "head" part is similar to a function pointer, whereas "args" are bound arguments to that function;
- Whereas previously each `pexpr` evaluated to a single column, now `Expr`s are allowed to evaluate to an arbitrary column list (possibly named). Thus, there is no longer a distinction between an "expr" and a "collist";
- Most functions will automatically map over their argument if the argument represents a list of column. For example, `sum(f[:])` is equivalent to `[sum(f[i]) for i in range(ncols)]`, except that the number of columns will be resolved dynamically;
- Exprs integrate tightly with virtual-column framework: all `Column` objects returned are expected to be virtual. This allows us to not have any separate "resolve" step during evaluation.
- This work is a prerequisite for #1979.

This PR adds a lot of code; however this code will eventually replace other expr code. This may or may not happen in this PR.
Currently this is still work-in-progress (some tests were marked as xfail).

Closes #1980


st-pasha:
This is a huge update, and it is not fully finished yet.
Still, it finally reached a more-or-less stable point where it can be merged. I could work on it more, but at this point it will only make the PR even more unmanageable. I believe, it is better to merge it into master, and then finish any remaining tasks as separate PRs. Or, if there are any new bugs (there probably are) that were not caught by tests, we should fix them as they are discovered.

PR #2013, opened 2019-09-19
Fix a bug in infinity check when fitting FTRL model

oleksiyskononenko:
We don't train FTRL model when target is infinite, and do an infinity check for both training and validation targets. In this PR we fix a bug in the infinity check for validation targets. This could cause an incorrect behavior of FTRL algo, and in some cases could even lead to a crash.


oleksiyskononenko:
This PR is not needed anymore due to changes introduced in #2014

PR #2016, opened 2019-09-19
Don't explicitly add system includes

siddhesh:
Adding /usr/include into the include path confuses gcc and breaks its
include_next mechanism.

### DIFF
--- a/ci/setup_utils.py
+++ b/ci/setup_utils.py
@@ -381,14 +381,14 @@ def get_compile_includes():
         includes.add("c")
         log.info("`c` is the main C++ source directory")
 
+        # Adding the include directory in sys.prefix as an -isystem flag breaks
+        # gcc because it messes with its include_next mechanism on Fedora where
+        # sys.prefix resolves to /usr.  It doesn't seem necessary either because
+        # with non-standard sys.prefix, CONFINCLUDEPY is good enough.
         confincludepy = sysconfig.get_config_var("CONFINCLUDEPY")
         if confincludepy:
             includes.add(confincludepy)
             log.info("`%s` added from CONFINCLUDEPY" % confincludepy)
-
-        sysprefixinclude = os.path.join(sys.prefix, "include")
-        includes.add(sysprefixinclude)
-        log.info("`%s` added from sys.prefix" % sysprefixinclude)
 
         # Include path to C++ header files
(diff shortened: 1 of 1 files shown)


ci/setup_utils.py
```diff
@@ -386,10 +386,6 @@ def get_compile_includes():
             includes.add(confincludepy)
             log.info("`%s` added from CONFINCLUDEPY" % confincludepy)
 
-        sysprefixinclude = os.path.join(sys.prefix, "include")
-        includes.add(sysprefixinclude)
-        log.info("`%s` added from sys.prefix" % sysprefixinclude)
-
```
st-pasha: Hmm, on my machine `sys.prefix` contains the path to the virtual env: 
```
>>> sys.prefix
'/Users/pasha/py36/bin/..'
```
So, `sysprefixinclude` is actually the path to virtual env's include directory, which sym-links to `/Library/Frameworks/Python.framework/Versions/3.6/include/python3.6m`. Admittedly, that's the same path as  `sysconfig.get_config_var("CONFINCLUDEPY")`, but it could be potentially something different with a different setup.

Does `sysprefixinclude` resolve to `/usr/include` on your system though?

siddhesh: Yeah, `sys.prefix` is `/usr` for a default python installation on Fedora and sysprefixinclude as a result resolves to `/usr/include`.  Could the path be not the same as `CONFINCLUDEPY` and be a non-standard include path at the same time?  If yes then a hack like excluding a hard coded '/usr' may be necessary.

The other alternative is to never use `-isystem` to avoid tinkering with the standard headers management.  The idea is really to just avoid using isystem for `/usr/include`.

st-pasha: Ok, so we know of at least 1 system (Fedora) where using `sys.prefix` is harmful, and at the same time we don't know of any system where `sys.prefix` is actually needed once `CONFINCLUDEPY` is used. Seems like it should be ok to remove it (and if it breaks for some other user, they will open an issue and we'll deal with it then).

Could you please add a comment into the code as to why `sys.prefix` should not be used -- otherwise someone else might add it back later ?

siddhesh: Done, pushed updated patch with the comment.

st-pasha: Thanks

PR #2045, opened 2019-09-27
Add a sort method to random attacker

oleksiyskononenko:
- add a sort method to random attacker;
- add random attack to testing pipeline (for Python 3.6+);
- several minor fixes.

### DIFF
--- a/tests/random_attack.py
+++ b/tests/random_attack.py
@@ -226,9 +226,8 @@ class Attacker:
         frame.replace_nas_in_column(icol, replacement_value)
 
     def sort_column(self, frame):
-        if frame.nrows == 0 or frame.ncols == 0:
+        if frame.ncols == 0:
             return
-
         icol = random.randint(0, frame.ncols - 1)
         print("[10] Sorting column %d ASC" % icol)
         if python_output:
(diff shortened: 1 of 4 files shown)


tests/random_attack.py
```diff
@@ -224,6 +224,15 @@ def replace_nas_in_column(self, frame):
                                 % (icol, icol, replacement_value))
         frame.replace_nas_in_column(icol, replacement_value)
 
+    def sort_column(self, frame):
+        if frame.nrows == 0 or frame.ncols == 0:
```
st-pasha: Sorting a 0-row frame is a valid operation, so why prohibit this case? 
Especially since it has a higher chance of uncovering an error...

oleksiyskononenko: This case won't work with the existing python part. Need to adjust it.

st-pasha: well you can do an early return in the python part

oleksiyskononenko: Yeah, adding it now.

oleksiyskononenko: Actually, even sorting on a zero-column frame will work. If one does `DT.sort()`. Not sure if this could be a useful use case though.

PR #2056, opened 2019-10-03
Fixed incorrect cloning of resized virtual columns

### LINKED ISSUE #2055: Internal error when resizing frame constructed from a masked numpy array
```
import datatable as dt
import numpy as np

DT = dt.cbind(
	dt.Frame(A=[1, 2]),
	dt.Frame(np.ma.array([True, True], mask=[False, False]))
)
DT.nrows = 1
DT = DT.copy()
```

st-pasha:
Certain virtual ColumnImpl classes were not propagating the `nrows_` property correctly during cloning after being resized.  `NpMasked_ColumnImpl` was one of such classes.

Closes #2055


oleksiyskononenko:
It looks like this PR fixed quite a lot of numpy tests that were failing in random attacker. Didn't encounter any new ones for the moment.

PR #2060, opened 2019-10-04
Cleanup of DataTable class

st-pasha:
- Do not materialize DataTable during join operation (this will result in a speedup);
- In some cases when `Column&` was requested changed the code to work with `const Column&`;
- DataTable properties renamed to have `_` at the end and made private;
- `DataTable::nrows()` and `DataTable::ncols()` are now getter functions;
- In certain places `DataTable` is now used by-reference rather than by pointer. Eventually we should transition the code to use by-value DataTable objects everywhere.

### DIFF
--- a/c/datatable.cc
+++ b/c/datatable.cc
@@ -3,7 +3,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 //
-// © H2O.ai 2018
+// © H2O.ai 2018-2019
 //------------------------------------------------------------------------------
 #include <algorithm>
 #include <limits>
@@ -19,7 +19,7 @@
 //------------------------------------------------------------------------------
 
 DataTable::DataTable()
-  : nrows(0), ncols(0), nkeys(0)
+  : nrows_(0), ncols_(0), nkeys(0)
 {
   TRACK(this, sizeof(*this), "DataTable");
 }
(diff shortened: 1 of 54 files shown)


c/datatable.cc
```diff
@@ -89,17 +106,17 @@ size_t DataTable::xcolindex(int64_t index) const {
 /**
  * Make a shallow copy of the current DataTable.
  */
-DataTable* DataTable::copy() const {
-  colvec newcols = columns;  // copy the vector
-  DataTable* res = new DataTable(std::move(newcols), this);
-  res->nkeys = nkeys;
+DataTable DataTable::copy() const {
```
oleksiyskononenko: What about renaming this to `clone()` for consistency with the corresponding `ColumnImpl` method? Btw, why do you think `clone` is a better name than `shallowcopy`?

st-pasha: Some time long ago we had `shallowcopy()` (which copied a reference to Column's buffer), and `deepcopy()`, which copied the buffer's content.
Today, `deepcopy()` no longer exists -- the buffer is copied automatically and only when necessary. But we added a "standard" copy too: `Column(const Column&)` which is even "shallower" than `shallowcopy()`. Thus, in order to avoid possible misunderstandings, I renamed `shallowcopy()` into `clone()`.

The `DataTable::copy()` method could be renamed into `clone()` as well, I don't see any contradiction there. Right now, however, its name mimics python's `Frame.copy()` (which in turns follows standard python naming conventions).

st-pasha: On a second thought, we don't have to name `DataTable::copy()` at all: this could be just a regular copy-constructor `DataTable(const DataTable&)`.

PR #2064, opened 2019-10-05
Minor cleaning of FTRL and Aggregator

oleksiyskononenko:
Minor cleaning and improvement of FTRL `update()` method


oleksiyskononenko:
Ok, let's use the convention you described from now on. I've just made the corresponding changes.

PR #2066, opened 2019-10-07
Close all files opened with open() in python

st-pasha:
Occasionally python emits ResourceWarnings in our test suit, such as
```
Exception ignored in: <_io.FileIO name='/tmp/tmp21vyfziy' mode='rb' closefd=True>
ResourceWarning: unclosed file <_io.BufferedReader name='/tmp/tmp21vyfziy'>
```
These, however, are not regular warnings, and can neither be detected nor captured by `pytest`. They do not reproduce reliably, and we can't write any test to check for their presence/absence.

After a bit of googling, it appears that these warnings are caused by file handlers that are not closed explicitly. Thus, in this PR I'm fixing several instances where we used to have un-closed file handles, in hope that the warnings will go away.

### DIFF
--- a/datatable/fread.py
+++ b/datatable/fread.py
@@ -380,21 +380,21 @@ class GenericReader(object):
 
         elif ext == ".bz2":
             import bz2
-            zf = bz2.open(filename, mode="rb")
-            if self._verbose:
-                self.logger.debug("Extracting %s into memory" % filename)
-            self._text = zf.read()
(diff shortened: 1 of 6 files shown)


st-pasha:
There are still 10 warnings in the logs, though I can't reproduce them locally...
Merging this PR, since it does eliminate the warnings at least partially, and it is not clear how to pinpoint the problem any further ...

PR #2073, opened 2019-10-09
Adjust NA handling in FTRL

oleksiyskononenko:
- to avoid confusions and to increase compatibility with DAI, FTRL hashers will hash `NA`'s to  `NA_S8`, that is `uint64_t(1) << 63`, for all the supported column types; 
- minor adjustment to frame rendering and tests.


thirdwing:
@oleksiyskononenko Can I know from which version of datatable includes this change?

oleksiyskononenko:
@thirdwing datatable includes this change as of 0.10.0a0.dev76

thirdwing:
Thanks. I will add more tests once we upgrade datatable in DAI.

PR #2076, opened 2019-10-10
Add a function to set key columns to random attacker

### LINKED ISSUE #2075: Stats is wrong after unsuccessful attempt to set a key
```
>>> import datatable as dt
>>> from datatable.internal import frame_integrity_check
>>> 
>>> DT = dt.Frame([1, 2, None, None])
>>> 
>>> try:
...     DT.key = ["C0"]
... except ValueError:
...     pass
... 
>>> DT.nrows = 3
>>> frame_integrity_check(DT)
AssertionError: in column 0 'C0': Stat NaCount's value is 2, but it was 1 upon recalculation
```

The following code doesn't produce any errors though:
```
>>> import datatable as dt
>>> from datatable.internal import frame_integrity_check
>>> 
>>> DT = dt.Frame([1, 2, None, None])
>>> DT.nrows = 3
>>> frame_integrity_check(DT)
```

oleksiyskononenko:
- set key columns in a random attacker;
- fix numpy python output;
- invalidate stats when mutable column implementation is requested;
- minor cleaning.

Closes #2075.

### DIFF
--- a/tests/random_attack.py
+++ b/tests/random_attack.py
@@ -138,7 +138,12 @@ class Attacker:
         new_nrows = int(curr_nrows * 10 / (19 * t + 1) + 1)
         print("[01] Setting nrows to %d" % (new_nrows, ))
         if python_output:
-            python_output.write("DT.nrows = %d\n" % new_nrows)
+            python_output.write("\ntry:\n"
+                                "   DT.nrows = %d\n"
+                                "except ValueError as e:\n"
+                                "   assert str(e) == 'Cannot increase the number of rows in a keyed frame'\n\n"
+                                % new_nrows)
+
         frame.resize_rows(new_nrows)
 
     def slice_rows(self, frame):
@@ -264,14 +269,21 @@ class Attacker:
     def set_key_columns(self, frame):
         if frame.ncols == 0:
             return
-        ncols_key = random.randint(1, frame.ncols)
-        a = random.sample(range(0, frame.ncols), ncols_key)
+        nkeys = random.randint(1, frame.ncols)
+        colids = random.sample(range(0, frame.ncols), nkeys)
+        colnames = [frame.names[i] for i in colids]
 
         print("[13] Setting %s: %r"
-              % (plural(len(a), "key column"), a))
+              % (plural(nkeys, "key column"), colids))
+
         if python_output:
-            python_output.write("DT = DT.key(%r)\n" % a)
-        frame.set_key_columns(a)
+            python_output.write("\ntry:\n"
+                                "   DT.key = %r\n"
+                                "except ValueError as e:\n"
+                                "   assert str(e) == 'Cannot set a key: the values are not unique'\n"
+                                % colnames)
+
+        frame.set_key_columns(colids, colnames)
 
 
 
@@ -296,10 +308,9 @@ class Attacker:
             if newn > 0 or n == 0:
                 return res
 
-    def random_array(self, n, newn=0, positive=False):
+    def random_array(self, n, positive=False):
         assert n > 0
-        if newn == 0:
-            newn = max(5, random.randint(n // 2, 3 * n // 2))
+        newn = max(5, random.randint(n // 2, 3 * n // 2))
         lb = 0 if positive else -n
         ub = n - 1
         return [random.randint(lb, ub) for i in range(newn)]
@@ -318,7 +329,7 @@ class Attacker:
         sort_columns: 1,
         cbind_numpy_column: 1,
         add_range_column: 1,
-        set_key_columns: 0,
+        set_key_columns: 1,
         # cbind_unique_columns: 0,
     }
     ATTACK_WEIGHTS = list(itertools.accumulate(ATTACK_METHODS.values()))
@@ -586,7 +597,13 @@ class Frame0:
 
     def resize_rows(self, nrows):
         curr_nrows = self.nrows
-        self.df.nrows = nrows
+        try:
+            self.df.nrows = nrows
+        except ValueError as e:
+            pass
+            assert(str(e) == "Cannot increase the number of rows in a keyed frame")
+            return
+
         if curr_nrows < nrows:
             append = [None] * (nrows - curr_nrows)
             for i, elem in enumerate(self.data):
@@ -695,12 +712,14 @@ class Frame0:
         names = self.random_names(1)
         df = dt.Frame(np_data.T, names=names)
         if python_output:
-            python_output.write("DTNP = dt.Frame(np.ma.array(%s, "
-                                "mask=%s, "
-                                "dtype=np.dtype(%s)).T)\n"
+            python_output.write("DTNP = dt.Frame(np.ma.array(%s,\n"
+                                "                mask=%s,\n"
+                                "                dtype=np.dtype(%s)).T,\n"
+                                "                names=%r)\n\n"
                                 % (repr_data([data], 14),
                                    repr_data([mmask], 14),
-                                   coltype.__name__))
+                                   coltype.__name__,
+                                   names))
             python_output.write("assert DTNP.shape == (%d, %d)\n\n"
                                 % (self.nrows, 1))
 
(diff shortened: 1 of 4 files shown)


tests/random_attack.py
```diff
@@ -4,7 +4,7 @@
 #   License, v. 2.0. If a copy of the MPL was not distributed with this
 #   file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #-------------------------------------------------------------------------------
-import sys; sys.path.insert(0, '.'); sys.path.insert(0, '..')
+import sys; sys.path += ('.', '..')
```
st-pasha: I don't think this will work: if `.` and `..` are added at the end of the path, then `datatable` module will be found among the installed `site-packages` first, not your working copy.
And btw I don't think `.` is needed either: the lookup is performed in the current directory first by default.

tests/random_attack.py
```diff
@@ -230,14 +236,13 @@ def replace_nas_in_column(self, frame):
     def sort_columns(self, frame):
         if frame.ncols == 0:
             return
-        ncols_sort = random.randint(1,5)
-        a = self.random_array(frame.ncols, newn=ncols_sort, positive=True)
+        ncols_sort = random.randint(1, frame.ncols)
```
st-pasha: This is problematic: a frame may have 100s or even 1000s columns. When you select a random integer in the range [1, ncols], you are usually selecting a very large number of columns, which will be very slow.
The most typical use-case is to sort by a single column, occasionally by few columns, and only very rarely we want to sort by many columns.

Same for creation of key down below.

tests/random_attack.py
```diff
@@ -262,6 +267,26 @@ def add_range_column(self, frame):
...
+        print("[13] Setting %s: %r"
+              % (plural(nkeys, "key column"), keys))
+
+        if python_output:
+            python_output.write("\ntry:\n"
+                                "   DT.key = %r\n"
+                                "except ValueError as e:\n"
+                                "   assert str(e) == 'Cannot set a key: the "
+                                "values are not unique'\n"
+                                % names)
```
st-pasha: Randomly chosen columns may or may not be valid for setting a key. However, at this point it is definitely knowable which case it is. Thus, the python output should be either
```
DT.key = %r
```
or 
```
with check_raises(ValueError('Cannot set a key: the values are not unique')):
    DT.key = %r
``` 
(this context manager can be added trivially).

tests/random_attack.py
```diff
@@ -573,7 +598,12 @@ def check_np_data(self):
 
     def resize_rows(self, nrows):
         curr_nrows = self.nrows
-        self.df.nrows = nrows
+        try:
+            self.df.nrows = nrows
+        except ValueError as e:
+            assert str(e) == "Cannot increase the number of rows in a keyed frame"
+            return
+
```
st-pasha: Same here: you should know in advance whether resizing should or should not throw an error.
It should throw iff (a) frame.nkeys > 0, and (b) new_nrows > nrows.

tests/random_attack.py
```diff
@@ -707,6 +739,23 @@ def add_range_column(self, name, rangeobj):
...
         self.df.cbind(dt.Frame(rangeobj, names=[name]))
 
+    def set_key_columns(self, keys, names):
+        try:
+            self.df.key = names
+        except ValueError as e:
+            assert str(e) == "Cannot set a key: the values are not unique"
+            return
+
+        nonkeys = sorted(list(set(range(len(self.data))) - set(keys)))
```
st-pasha:
```suggestion
        nonkeys = set(range(self.ncols)) - set(keys)
        new_col_order = list(keys) + sorted(nonkeys)
```

tests/random_attack.py
```diff
@@ -707,6 +739,23 @@ def add_range_column(self, name, rangeobj):
         self.dedup_names()
         self.df.cbind(dt.Frame(rangeobj, names=[name]))
 
+    def set_key_columns(self, keys, names):
+        try:
+            self.df.key = names
+        except ValueError as e:
```
st-pasha: Same here: you should know in advance whether setting the key is valid or not,
and then either set the key (when it's valid),
or check that a proper exception is thrown (if it's invalid).

tests/random_attack.py
```diff
@@ -707,6 +739,23 @@ def add_range_column(self, name, rangeobj):
...
+            assert str(e) == "Cannot set a key: the values are not unique"
+            return
+
+        nonkeys = sorted(list(set(range(len(self.data))) - set(keys)))
+
+        self.types = [self.types[i] for i in keys] + [self.types[i] for i in nonkeys]
+        self.names = [self.names[i] for i in keys] + [self.names[i] for i in nonkeys]
```
st-pasha:
```suggestion
        self.types = [self.types[i] for i in new_col_order]
        self.names = [self.names[i] for i in new_col_order]
```

tests/random_attack.py
```diff
@@ -707,6 +739,23 @@ def add_range_column(self, name, rangeobj):
...
+
+        if (len(self.data[0])):
```
st-pasha:
```suggestion
        if self.nrows > 0:
```

tests/random_attack.py
```diff
@@ -757,14 +806,16 @@ def dedup_names(self):
                                str(args.seed) + ".py")
         python_output = open(outfile, "wt")
         python_output.write("#!/usr/bin/env python\n")
+        python_output.write("import sys; sys.path += ('.', '..')\n")
         python_output.write("import datatable as dt\n")
         python_output.write("import numpy as np\n")
         python_output.write("from datatable import f\n")
-        python_output.write("from datatable.internal import frame_integrity_check\n\n")
+        python_output.write("from datatable.internal import frame_integrity_check\n")
```
st-pasha: The extra `\n` was needed to visually separate imports from the rest of the code.

tests/random_attack.py
oleksiyskononenko: I have uninstalled `datatable` prior to start attacks, so that to be sure only a single instance of datatable is tested. However, there is no problem to add `.` and `..` to the beginning of path. Btw, without adding `.` some imports like `from datatable.utils.terminal` won't work, so I will leave it.

tests/random_attack.py
oleksiyskononenko: I tested both approaches, and found that even larger number of columns is ok, but may be testing smaller number of columns is better, so let's do something like `int(random.expovariate(1.0)) + 1` here.

tests/random_attack.py
oleksiyskononenko: `self.ncols` is calculated on a frame as `self.df.ncols`, while here we're dealing with python data, so it would be nice to stay within the python data context and keep using `len(self.data)` or something like that. The rest, yes, could be simplified.

tests/random_attack.py
oleksiyskononenko: Not sure we should use frame data here either.

tests/random_attack.py
oleksiyskononenko: Yeah, I was cleaning my debug code and this one got affected, will revert it back.

oleksiyskononenko:
Ok, let's do this in `_get_mutable_impl()`, but I will add an early return if `new_nrows == curr_nrows` to preserve stats in this case.

tests/random_attack.py
oleksiyskononenko: Sure, we can improve this part easily.

tests/random_attack.py
st-pasha: 👍

tests/random_attack.py
st-pasha: You may adjust the calculation of `.ncols` property, but the code clarity is generally improved when you use `self.ncols` instead of `self.shape[1]` or `len(self.data)`.

oleksiyskononenko: Yes, I agree, but for some reason `self.ncols` and `self.nrows` are functions of a frame, though `self.types`, `self.names` and `self.data` are functions of python data. 

I propose we keep all datatable stuff under `self.df`, i.e. whenever we do some operations on the frame we use `self.df.ncols` and `self.df.nrows`. When we do operations on the python data along with `self.names`, `self.types`, we also use `self.ncols` defined as `len(self.data)` and `self.nrows` defined as `len(self.data[0])`.

Issue #2008, opened 2019-09-18
Handle zip-files with multiple files.


mmalohlava:
### Describe what feature you would like to see implemented

`fread` option to allow optimistic parse of zip-files with multiple files. The fread will try to parse the file individually and rbind the resulting frame.

I.e, avoid limitations in the code mentioned here: https://github.com/h2oai/datatable/blob/master/datatable/fread.py#L359-L362

pseudotensor:
related https://github.com/h2oai/datatable/issues/488

pseudotensor:
Also related https://github.com/h2oai/h2oai/issues/10526. Dt already supports multiple tables, DAI doesn't use it.

pseudotensor:
basically it should be smarter, ignore unrelated file , try to rbind, and if fails return as individual files like done for excel sheets.

st-pasha:
This is something that is relatively easy to implement, but not that easy to figure a good API.

From the API standpoint, having a function that sometimes returns a Frame, while other times returning a dictionary of Frames -- it just doesn't look very reliable. Unfortunately, we're already doing this, returning multiple frames for certain types of inputs. However, extending that functionality will only exacerbate the problem.

So I wonder what a proper solution would be here, API-wise. A separate function that returns multi-frame output? A flag that forces the output to have multiple frames? How would this all look from the perspective of a type annotation system?

pseudotensor:
Sure, you can have different function arguments to control eventual result, but at least needs some smarts to avoid errors, e.g. errors='ignore' if some files are "bad" like MAC os files that often float around together with zips.

st-pasha:
So, I was thinking about the following solution:

1. New function `mread(...)`, which works similarly to `fread()` but always returns a dictionary, even for a single source (or, more generally, a `FrameSet` object which is derived from `dict`).

2. The current `fread()` acquires new option `multiple_sources = "warn" | "error" | "ignore" | "rbind" | "rbind_force"`. This option controls the behavior of what fread should do when the input resolves into multiple sources. The option "error" will raise an error immediately. The option "warn" would return the first input only, with a warning. The "ignore" will do the same, but without the warning. The "rbind" will attempt to rbind() the sources, or raise an error if it cannot. The "rbind_force" will perform `rbind(force=True)`, which will unlikely to throw an exception but if the frames were mismatched there will be many holes in the returned Frame.

[...]

Issue #1832, opened 2019-04-28
fread python code should be converted to C++


st-pasha:
Having this code in python prevents from fully integrating C++ functionality (such as progress bars) with functionality that is currently in python (such as loading content from URL or un-archiving a file). In addition, we are prevented from introspecting file's content when detecting its format.

PR #2077, opened 2019-10-10
Fread partial refactoring

st-pasha:
- Code for checking fread inputs was repeated in Python and C++, so removed python part;
- No longer pass property `._columns` from C++ to python and then back again;
- Class FreadReader significantly simplified (to be removed in the future);
- Start working on a new (prototype) fread function, though it's not doing anything yet.

WIP for #2008 
WIP for #1832

### DIFF
--- a/c/csv/py_csv.cc
+++ b/c/csv/py_csv.cc
@@ -31,12 +31,10 @@ file types, not just csv.
 )");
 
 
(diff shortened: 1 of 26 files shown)


st-pasha:
(build is stuck completely, merging)

PR #2081, opened 2019-10-11
Import dt from datatable

### LINKED ISSUE #2080: Allow importing datatable from itself
Currently we do the following imports at the top of a script:
```
import datatable as dt
from datatable import f, by, join, ...
```
It would be nice if we could do the following instead:
```
from datatable import dt, f, by, join, ...
```

st-pasha:
Symbol `dt` (which is the `datatable` module itself) is now exported from `datatable` module.

Closes #2080

### DIFF
--- a/datatable/__init__.py
+++ b/datatable/__init__.py
@@ -47,6 +47,7 @@ except ImportError:
 __all__ = (
     "__git_revision__",
     "__version__",
+    "dt",
     "Frame",
     "mean",
     "median",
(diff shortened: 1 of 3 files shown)


st-pasha:
So, you can now even do this:
```
>>> import datatable as dt
>>> dt.dt.dt.dt.dt.dt.dt.dt.dt.dt.dt.dt.dt.dt.dt.Frame([1])
    C0
--  --
 0   1

[1 row x 1 column]
```

PR #2082, opened 2019-10-11
Fix grouping of an empty frame

### LINKED ISSUE #2078: Internal error when groupping an empty frame
```
>>> import datatable as dt
>>> from datatable import f, by, count
>>> 
>>> DT = dt.Frame([[]])
>>> DT[:, count(f[0]), by(0)]
AssertionError: Assertion 'max < rii->length || max == RowIndex::NA' failed in c/rowindex_array.cc, line 365
```

st-pasha:
Closes #2078

### DIFF
--- a/tests/test-groups.py
+++ b/tests/test-groups.py
@@ -117,9 +117,17 @@ def test_group_empty_frame1():
 def test_group_empty_frame2():
     DT = dt.Frame(A=[])
     D1 = DT[:, count(), by(f.A)]
-    D2 = DT[:, count(f.A), by(f.A)]
     frame_integrity_check(D1)
+    assert D1.shape == (1, 0)
+    assert D1.stype == dt.int64
+
+
+def test_group_empty_frame3():
+    DT = dt.Frame(A=[])
+    D2 = DT[:, count(f.A), by(f.A)]
     frame_integrity_check(D2)
+    assert D2.shape == (1, 0)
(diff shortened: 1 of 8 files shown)


tests/test-groups.py
```diff
@@ -30,89 +30,115 @@
...
-def test_groups_internal2():
-    d0 = dt.Frame([[1,   5,   3,   2,   1,    3,   1,   1,   None],
+def test_groups1():
+    DT = dt.Frame([[1,   5,   3,   2,   1,    3,   1,   1,   None],
                    ["a", "b", "c", "a", None, "f", "b", "h", "d"]],
                   names=["A", "B"])
-    d1 = d0[:, :, by("A")]
-    # gb = d1.internal.groupby
-    # assert gb.ngroups == 5
-    # assert gb.group_sizes == [1, 4, 1, 2, 1]
```
oleksiyskononenko: btw, how to get access to number of groups in python since `.internal` is gone?

st-pasha: well, that property no longer exists, so there is no way to access it.

oleksiyskononenko: Yeah, but sometimes it is nice to know number of groups.

st-pasha: It sure is! So maybe we can add some kind of helper function to access this and similar properties. Though I'm not sure how it's supposed to work API-wise

PR #2084, opened 2019-10-12
Implement creation of Frames from numpy scalars

### LINKED ISSUE #2079: [FR] datatable should recognize numpy scalar values
Original Title: **ValueError: Cannot read values of stype obj64**

I am not sure this is a problem of `datatable`. 

The `ltypes` is changed into `ltype.obj` and `stypes` is changed into `stype.obj64`. When I do `df.Frame()` after `np.append()`.
So I can not write data into csv file and `ValueError: Cannot read values of stype obj64` is shown 

This is example code

```python
import numpy as np
import datatable as dt

a = [(i, i+1) for i in range(10)]
a_np = np.array(a)

a_list = []
for i in a_np:
    for x in np.array(range(i[-1]+2, 10)):
        merge = tuple(np.append(i, x))
        a_list.append(merge)

a_list_dt = dt.Frame(a_list) # ltypes, stypes is changed into object
```

I don't make sense why `stypes` of column becomes `obj64` when creating Frame using `datatable`

st-pasha:
- Add ability to create a Frame from a list of numpy integers, or numpy floats;
- Added internal methods `obj::is_numpy_int()` and `obj::is_numpy_float()`;

Closes #2079

### DIFF
--- a/c/column_from_pylist.cc
+++ b/c/column_from_pylist.cc
@@ -204,6 +204,11 @@ static bool parse_as_int(const iterable* list, Buffer& membuf, size_t& from)
         outdata[i] = litem.ovalue<T>(&overflow);
         if (!overflow) continue;
       }
+      int r = item.is_numpy_int();
+      if (r && r <= int(sizeof(T))) {
+        outdata[i] = static_cast<T>(item.to_int64());
+        continue;
+      }
       from = i;
       return false;
     }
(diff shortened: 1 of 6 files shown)


oleksiyskononenko:
Re-triggering the build made it all green, thanks to @abal5.

PR #2085, opened 2019-10-12
Add delete columns operation to random attacker

oleksiyskononenko:
- add delete column operation to random attacker;
- don't delete boolean column in `select_rows_with_boolean_column()`;
- minor other adjustments to the unit tests.

Helped to reveal #2083.

PR #2086, opened 2019-10-12
Fix crash when key columns are deleted from a Frame

### LINKED ISSUE #2083: Segfault after removing a key column
```
import datatable as dt
DT = dt.Frame([[1], [2]])
DT.key = ["C0", "C1"]
del DT[0]
print(DT)
```

Produces `Segmentation fault: 11` / `Bus error: 10`.

st-pasha:
Closes #2083

### DIFF
--- a/c/datatable.cc
+++ b/c/datatable.cc
@@ -103,10 +103,21 @@ DataTable DataTable::extract_column(size_t i) const {
 }
 
 
+// Remove columns at the specified indices
 void DataTable::delete_columns(intvec& cols_to_remove) {
   if (cols_to_remove.empty()) return;
   std::sort(cols_to_remove.begin(), cols_to_remove.end());
   cols_to_remove.push_back(size_t(-1));  // guardian value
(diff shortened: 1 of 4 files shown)


oleksiyskononenko:
Merging this to master as it passes all the tests, and this branch was already merged to #2085.

oleksiyskononenko:
Btw, what about at least issuing a warning when we un-key frame deleting column(s)? We already do this when we de-duplicate column names during cbinding, in this case we also kind of doing operation that was not directly requested by the user.

st-pasha:
Well, I was thinking that warnings ought to be reserved for the situations when something really unexpected happens. In this case a user deletes a column that is in the key -- pretty clear the key would no longer exist afterwards.

We could discuss further how to best judge the situations of when to issue warnings or not.

oleksiyskononenko:
I guess it could be clear for us, but user didn't request this operation explicitly.

PR #2093, opened 2019-10-15
Add join operation to random attacker

### LINKED ISSUE #2090: rbind operation on a keyed frame results in duplicated data in the key column
With this code one gets a key column with the duplicated data:
```
import datatable as dt
DT = dt.Frame([1])
DT.key = "C0"
DT.rbind(DT)
print(DT)
```

Output:
```
C0 | 
---+ 
 1 | 
 1 | 

[2 rows x 1 column]
```
Looks like `rbind` doesn't do any checks related to keys.

oleksiyskononenko:
- add join-frame-with-itself operation to random attacker;
- add key check to make sure number of keys is the same between python and datatable;
- restore frame sorting operation that was accidentally deleted in one of the previous PRs;
- drop keys after `rbind()` operation.

Adding key checks revealed a problem that we don't preserve key columns in many cases when we could: row slicing, join, etc. 

Closes #2090

### DIFF
--- a/c/frame/rbind.cc
+++ b/c/frame/rbind.cc
@@ -295,6 +295,7 @@ void DataTable::rbind(
   }
   ncols_ = new_ncols;
   nrows_ = new_nrows;
+  nkeys_ = 0; // this will drop key columns, if they existed
 }
 
 
--- a/c/frame/rbind.cc
+++ b/c/frame/rbind.cc
@@ -295,7 +295,7 @@ void DataTable::rbind(
   }
   ncols_ = new_ncols;
   nrows_ = new_nrows;
-  nkeys_ = 0; // this will drop keys, if they was previously set
+  nkeys_ = 0; // this will drop all the keys, if there were any
 }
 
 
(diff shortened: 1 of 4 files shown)


c/frame/rbind.cc
```diff
@@ -295,6 +295,7 @@ void DataTable::rbind(
   }
   ncols_ = new_ncols;
   nrows_ = new_nrows;
+  nkeys_ = 0; // this will drop all the keys, if there were any
```
st-pasha: We should prefer to throw an error when a user tries to rbind a keyed frame, similarly to how we do not allow resizing a keyed frame:
```
>>> import datatable as dt
>>> DT = dt.Frame(A=range(5))
>>> DT.key = 'A'
>>> DT.nrows = 10
ValueError: Cannot increase the number of rows in a keyed frame
```
However, this can be deferred to a later PR

oleksiyskononenko: Yes, this is a temporary fix. I will create an issue and list cases where operations on the keyed frame should be adjusted.

Issue #1677, opened 2019-02-21
Eliminate dependency on blessed


st-pasha:
We use `blessed` to handle widget rendering in a python terminal. That library itself has 2 dependencies: `wcwidth` and `six`.

Currently, the amount of functionality that we use from that library is not too big; plus we add a few hacks on top of that to achieve what we actually need. Overall, I believe removing blessed will have a net positive effect.

The following functions to be reimplemented:
- [x] various colors
- [x] `term.is_a_tty`
- [x] `term.width` / `term.height`
- [x] `term._encoding`
- [x] `term.length()`
- [x] `term.move_x`
- [x] `term.move_up`
- [x] `term.clear_eol`
- [ ] `term.cbreak` / `term.inkey`

jquast:
author of blessed and wcwidth here, its all MIT licensed, you're welcome to copy any code that you wish directly :) enjoy

st-pasha:
Thanks, Jeff.

PR #2096, opened 2019-10-16
Display

### LINKED ISSUE #2087: print() could output top+bottom rows of a Frame
Long time user and advocate for R's data.table. Would love if datatable's `print()` method behaved like data.table's `print()` method, whereby

1. By default, the top 5 and bottom 5 rows are printed (assuming the row count is > 100)
2. The argument `topn` can specify how many rows should be printed

Keep up the great work. This package is awesome!

st-pasha:
- created C++ `Terminal` class, which can replace the python `Terminal` at some point. This class allows producing proper colored output from C++;
- Re-implemented text widget rendering in C++ (to replace `DataFrameWidget` in "widget.py"). This  ensures that Frame rendering into HTML and text can  share common functionality;
- Added display options `display.max_nrows`, `display.head_nrows` and `display.tail_nrows` to control the  appearance of the  widget in the console/HTML.
- Slight adjustments to styling of the output frame, in particular all NA values are now displayed as "NA" (previously they were rendered  as empty strings).

Closes #2087 
WIP for #1677

### DIFF
--- a/c/frame/__repr__.cc
+++ b/c/frame/__repr__.cc
@@ -446,11 +446,23 @@ static PKArgs args_view(
   0, 1, 0, false, false, {"interactive"}, "view", nullptr);
 
 void Frame::view(const PKArgs& args) {
-  oobj interactive = args[0].is_undefined()? obool(true) : args[0].to_oobj();
-  oobj DFWidget = oobj::import("datatable")
-                  .get_attr("widget")
-                  .get_attr("DataFrameWidget");
-  DFWidget.call({oobj(this), interactive}).invoke("render");
+  bool interactive = true;  // default when `interactive` is omitted entirely
+  if (args[0].is_none()) interactive = dt::display_interactive;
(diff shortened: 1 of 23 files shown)


c/frame/__repr__.cc
```diff
@@ -75,135 +82,135 @@ static bool in_jupyter() {
...
-        if (j == cols0) {
-          j = ncols - cols1;
+      if (render_row_indices_) {
+        html << "<td class='row_index'>";
+        _render_comma_separated(i);
+        html << "</td>";
+      }
+      for (size_t j : colindices_) {
+        if (j == dt::NA_index) {
           html << "<td class=vellipsis>&hellip;</td>";
```
oleksiyskononenko: Adding single-quotes will make it consistent with the rest of the code: `class='vellipsis'`

st-pasha: sure, I can do it with the next PR

PR #2097, opened 2019-10-16
Fix setting a key on a copied frame

### LINKED ISSUE #2095: Setting a key  may occasionally fail?
MRE:
```
import datatable as dt
medium = dt.open('medium.jay')
y = medium.copy()
y.key = 'id2'  # works
y = medium.copy()
y.key = 'id2'  # fails
```
Setting the key the second time fails with `ValueError: Cannot set a key: the values are not unique`.

(Thanks to @jangorecki for discovering this bug).

[medium.jay.zip](https://github.com/h2oai/datatable/files/3735404/medium.jay.zip)

st-pasha:
The issue was that when the frame was copied, the `py_inames` dictionary was copied using python's default copy-by-reference semantics. Subsequently, changes to that dictionary in one frame were affecting the other frame too, causing them to become inconsistent.  Now we explicitly deep-copy the `py_inames` dictionary.

Closes #2095

### DIFF
--- a/c/datatable.cc
+++ b/c/datatable.cc
@@ -22,6 +22,20 @@ DataTable::DataTable()
   : nrows_(0), ncols_(0), nkeys_(0) {}
 
 
+// Default copy is not suitable because `py_inames` has no CoW semantics, and
+// its default copy-by-reference will cause shared state between Frames.
+DataTable::DataTable(const DataTable& other)
+  : nrows_(other.nrows_),
+    ncols_(other.ncols_),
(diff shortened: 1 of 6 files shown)


st-pasha:
The issue was that when the frame was copied, the `py_inames` dictionary was copied using python's default copy-by-reference semantics. Subsequently, changes to that dictionary in one frame were affecting the other frame too, causing them to become inconsistent.  Now we explicitly deep-copy the `py_inames` dictionary.

Closes #2095

PR #2099, opened 2019-10-17
Add shallow copy operation to random attacker

oleksiyskononenko:
This is in response to #2095, that we missed due to the absence of copy operations in the random attacker. Testing it on a commit before PR #2097 was pushed, revealed the problem right away. With the latest master 10K seeds passed — no failures.

### DIFF
--- a/tests/random_attack.py
+++ b/tests/random_attack.py
@@ -925,6 +925,7 @@ if __name__ == "__main__":
         python_output.write("#!/usr/bin/env python\n")
         python_output.write("#seed: %s\n" % args.seed)
         python_output.write("import sys; sys.path = ['.', '..'] + sys.path\n")
+        python_output.write("import copy\n")
         python_output.write("import datatable as dt\n")
         python_output.write("import numpy as np\n")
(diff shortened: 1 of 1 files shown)


tests/random_attack.py
```diff
@@ -851,6 +864,12 @@ def join_self(self):
         self.dedup_names()
         return True
 
+    # This is a noop operation for the python data
+    def shallow_copy(self):
+        self.df_shallow_copy = self.df.copy()
+        self.df_deep_copy = copy.deepcopy(self.df)
```
st-pasha: Does `copy.deepcopy` even works on a Frame?

st-pasha: Looks like both `copy.copy()` and `copy.deepcopy()` invoke Frame's pickling protocol... We should probably adjust this behavior: https://docs.python.org/3/library/copy.html

oleksiyskononenko: Yes, it works, I was like it is just doing pickling/unpickling, no? I've been using it in some FTRL tests and it seems to work fine.

PR #2100, opened 2019-10-17
Allow a rowindex from int column with NAs

### LINKED ISSUE #1496: Support for i selector an integer column with value -1

st-pasha:
- The "array" rowindex now keeps its data in a `Buffer` object (previously it was a plain C `void*` pointer);
- ArrayRowIndex can now be created with 0-copying from a material column of type int32/int64;
- ArrayRowIndex can now be constructed from an integer columns containing NA values;
- The ArrayRowIndex now uses the MININT value to indicate missing rowindices (previously `-1` was used);
- The `min` field of the RowIndex is removed (not needed anymore);
- Added boolean field `max_valid`, which indicates whether the value of `max` field is NA or not. The flag is false iff either the length of rowindex is 0, or if all elements of the  rowindex are missing.

The RowIndex is now conceptually not much different from a `Column`, which means in the future we may just represent the RowIndex as a Column internally.

Closes #1496


oleksiyskononenko:
I understand a reason for this change, but won't it be a source of confusion?

```
In [1]: import datatable as dt                                                                                                                         

In [2]: DT = dt.Frame([0,1,2,3,4])                                                                                                                     

In [3]: x = dt.Frame([-1,0,1])                                                                                                                         

In [4]: DT[x, :]                                                                                                                                       
Out[4]:    | C0
---+ --
 0 | NA
 1 |  0
 2 |  1

[3 rows x 1 column]

In [5]: DT[[-1,0,1],:]                                                                                                                                 
Out[5]:    | C0
---+ --
 0 |  4
 1 |  0
 2 |  1

[3 rows x 1 column]

```

st-pasha:
You're right, it might be confusing at first. But I don't see any way out of this dilemma:
- `DT[-1, :]` should be the last row of a Frame, because this is python's standard  way of addressing arrays;
- this extends to a simple list: `DT[[-1, -2], :]` are the two last rows;
- on the other hand, internally we store RowIndex as an array of `int32` indices that can be used directly for element lookup, which means no wrap-around semantics and no negatives (except for `-1` which indicates  a "NA" index);
- this is why in `DT[col, :]`, the `col`  is not allowed to contain any negative numbers or NAs: the data in the `col` will be directly used  as a RowIndex. With this PR we allow the value `-1` as well, to indicate the missing index.

Note that at some point I'd like to expose  the RowIndex of a Frame as a  property that will return that RowIndex in the  form of an int column. And I'd like to be able to apply that  column to  some other frame.

oleksiyskononenko:
>Note that at some point I'd like to expose the RowIndex of a Frame as a property that will return that RowIndex in the form of an int column. And I'd like to be able to apply that column to some other frame.

What if we just apply it not through the `[i, j]` syntax, but through a rowindex setter?

st-pasha:
But `[i, j]` is THE syntax to apply filters  / column transformations to a Frame. How would  you explain it that in one case that syntax doesn't work and another syntax is needed? And btw `DT[col, :]` is already supported  -- why would  we stop supporting it?

oleksiyskononenko:
It looks natural that if one gets RowIndex as `ri = DT.rowindex`,  then applying it one should do `DT.rowindex = ri`. I don't see why we should stop supporting `DT[col, :]` in this case, we continue of course and just say that RowIndex is not directly applicable as a row selector, because it comes from internals.

st-pasha:
Yes, but the hypothetical `DT.rowindex` would  still return a column, of stype `int32` / `int64`. And we want the user to be able to manipulate that column. So when it arrives into `DT[i, j]`, it would be indistinguishable from any other int column.

Though I concede that encoding "missing" indices with `-1` is an implementation detail that might not be of interest to the user. And perhaps they would have found it more natural for missings to be represented with `NA` values...

So, let's suppose we do as you suggest, and treat negative values in an integer column as wrap-around modulo nrows. Then in order to convert an int column into a `RowIndex` object we'd have to:
- check the min() stat of a column, and if it is negative - modify the values in the buffer by converting them into positive ones;
- check the column's NA count, and if there are any NAs replace them with `-1`s.

In the opposite direction it would just involve creating a column where `-1` values are mapped into NAs. It could be a virtual column.

oleksiyskononenko:
I understand drawbacks of both approaches, but still think that `DT[some_list, :]` and `DT[dt.Frame(some_list), :]` should return the same data, otherwise it is very confusing. To me it seems that the source of this problem is that we encode RowIndex `NA`s with `-1`s. Is there any particular reason to do so?

I imagine we can do something like that instead:
1) encode RowIndex `NA`s with int32/int64 `NA`s;
2) make `DT[[None], :]` result in a row filled with `NA` s.

Then when one does `ri = DT.rowindex`, `ri` will have `NA`s instead of `-1`s (will look more natural), and when we apply it to a frame as a row selector, `DT[ri, :]` will contain the same data as DT[ri.to_list(), :]`.

Right now when you do `DT[[None], :]`, you simply get no rows at all. Do you foresee any problems if we change this behavior and just do (2)?

st-pasha:
Yes, this is something we might want to straighten out.

First, the reason for `-1` being used as a "missing" indicator is because this number works equally well for both `int32` and `int64` arrays. This allows us to  write `if (indices[i]  == RowIndex::NA) ...` without having to worry about 2 different kinds of NAs. In addition, conversion from int32 into int64 becomes straightforward too:  `static_cast<int64_t>(int32_t(-1)) == int64_t(-1)`. The  same is not true for `GETNA<T>()` constants.

Thus, the primary reason is convenience, which means it is not particularly compelling. Good thing we don't have anymore that many places that explicitly check rowindex values, so it might not be that hard to change  the convention.

oleksiyskononenko:
Yeah, especially that this is something we can just change internally, but then we will expose API that will be much more self-consistent.

Issue #2098, opened 2019-10-17
Improve datatable behavior when working with the keyed frames


oleksiyskononenko:
Improving random attacker, it was found that we need to adjust datatable behavior in terms of the operations on keyed frames. Since keying a frame is an expensive operation, as it requires sorting one or several columns, we need to preserve keys as much as possible. In particular, key columns can be easily kept after the following operations:

- selecting any row subset;
- selecting a column subset that contains all the keyed columns;
- joining two frames;
- sorting, only in the case when it is done over all the key columns asc (not sure if we want to implement this, because it is probably a very rare case as the result is identical to the original frame). 

The following operations on keyed frames should throw an exception:
- rbinding it to anything besides the zero-row frame;
- deleting a column that is a part of the multi-column key, unless frame contains zero rows.

PR #2111, opened 2019-10-21
Do not allow particular operations on keyed frames

oleksiyskononenko:
Throw an exception when the following operations are performed on keyed frames:
- rbinding it to anything besides the zero-row frame;
- deleting a column that is a part of the multi-column key, unless frame contains zero rows;
- replacing values in a key column;
- increasing number of rows.

Unit tests and random attacker have been adjusted according to these changes.

WIP for #2098


c/datatable.cc
```diff
@@ -123,17 +123,33 @@ void DataTable::delete_columns(intvec& cols_to_remove) {
...
+  // property or to remove a key column if there is no other keys.
+  if (cols_to_remove[0] < nkeys_ && nkeys_ > 1 && nrows_ > 0) {
+    throw ValueError() << "Cannot delete a column that is a part of a "
+        "multi-column key";
   }
 
+  // Calculate how many key columns we need to remove and adjust `nkeys_`
+  size_t nkeys_remove = 0;
   size_t j = 0;
+  while (cols_to_remove[j] < nkeys_) {
```
st-pasha: This can be easily merged with the loop below.

c/frame/replace.cc
```diff
@@ -252,7 +255,7 @@ void ReplaceAgent::parse_x_y(const Arg& x, const Arg& y) {
 //
 // Here we analyze the input lists `vx`, `vy` and split them into 4 sublists
 // according to their elements types. We also do further verification that the
-// types of elements in vectors `vx`, `vy` match, and that there no duplicates.
+// types of elements in vectors `vx`, `vy` match, and that there is no duplicates.
```
st-pasha: *are

c/datatable.cc
```diff
@@ -172,7 +188,7 @@ void DataTable::delete_all() {
 void DataTable::resize_rows(size_t new_nrows) {
   if (new_nrows == nrows_) return;
   if (new_nrows > nrows_ && nkeys_ > 0) {
-    throw ValueError() << "Cannot increase the number of rows in a keyed frame";
+    throw ValueError() << "Cannot increase number of rows in a keyed frame";
```
st-pasha: ["cannot increase the number"](https://www.google.com/search?sxsrf=ACYBGNQzQdr2rtmjvarkZfW6_zTmjgLvaA%3A1571677765426&ei=ReatXc_TGc-QtQWAsZBw&q=%22cannot+increase+the+number%22&oq=%22cannot+increase+the+number%22&gs_l=psy-ab.3..0i8i30.5173.5554..5812...0.0..0.98.372.4......0....1..gws-wiz.......0i5i30j0i8i7i30.-_KsUQPg1X0&ved=0ahUKEwiPiqP3663lAhVPSK0KHYAYBA4Q4dUDCAs&uact=5) - 261000 results
["cannot increase number"](https://www.google.com/search?sxsrf=ACYBGNR4Kb8FBbFw8_-w_xa88Pg8vRl2DA%3A1571677772224&ei=TOatXa6bDc_0tAWAjYugDQ&q=%22cannot+increase+number%22&oq=%22cannot+increase+number%22&gs_l=psy-ab.3..0i5i30l2j0i8i30.62745.62858..63189...0.0..0.96.178.2......0....1..gws-wiz.DqYUKJdGBJ4&ved=0ahUKEwiu8cH6663lAhVPOq0KHYDGAtQQ4dUDCAs&uact=5) - 1880 results

c/datatable.cc
```diff
@@ -123,17 +123,33 @@ void DataTable::delete_columns(intvec& cols_to_remove) {
...
   // out of multi-column key is removed, the remaining columns may no longer
   // provide unique key values.
-  // On the other hand, if the DataTable is keyed, it is still safe to remove
-  // any non-key columns without affecting the DataTable "keyed" property.
-  if (cols_to_remove[0] < nkeys_) {
-    nkeys_ = 0;
+  // On the other hand, if the DataTable is keyed, it is still safe
+  // to remove any non-key columns without affecting the DataTable "keyed"
+  // property or to remove a key column if there is no other keys.
+  if (cols_to_remove[0] < nkeys_ && nkeys_ > 1 && nrows_ > 0) {
```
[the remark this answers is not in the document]
oleksiyskononenko: Sure, let me also add this case.

c/frame/replace.cc
oleksiyskononenko: I guess both forms can be used:
"[there is no duplicates](https://www.google.com/search?q=%22there+is+no+duplicates%22)" About 146,000 results 
"[there are no duplicates](https://www.google.com/search?q=%22there+are+no+duplicates%22)" About 155,000 results

c/datatable.cc
oleksiyskononenko: I will check with a native speaker, because not sure about the context. To me "cannot increase the number of rows" sounds like we cannot increase "this particular number of rows", though we can't increase any number in this case.

c/datatable.cc
oleksiyskononenko: Yes, but I'm now starting to think that there could be just two possibilities: we don't remove any key columns or we remove all of them. In all other cases we should throw an error. Will follow-up on this part.

oleksiyskononenko: After adding a possibility to remove all the key columns at once, these two loops cannot be merged anymore, because the second loop affects the column order.

c/frame/replace.cc
oleksiyskononenko: After comparing the sources reliability, changed this to "are".

c/datatable.cc
oleksiyskononenko: Looks like "the" in this case is kind of arbitrary, so just leave it as it was before.

c/datatable.cc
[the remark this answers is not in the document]
oleksiyskononenko: Done.

PR #2114, opened 2019-10-22
dt.options.display.allow_unicode = False

st-pasha:
Under the option `dt.options.display.allow_unicode = False` we will now display in Frame view unicode strings escaped. For example:
```
>>> DT = dt.Frame(uni=["møøse", "𝔘𝔫𝔦𝔠𝔬𝔡𝔢", "J̲o̲s̲é̲", "🚑💥✅"])
>>> DT
   | uni                                                                   
-- + ----------------------------------------------------------------------
 0 | m\xF8\xF8se                                                           
 1 | \U0001D518\U0001D52B\U0001D526\U0001D520\U0001D52C\U0001D521\U0001D522
 2 | J\u0332o\u0332s\u0332\xE9\u0332                                       
 3 | \U0001F691\U0001F4A5\u2705                                            

[4 rows x 1 column]
```


c/frame/repr/text_column.cc
```diff
@@ -183,30 +184,61 @@ static std::string _escaped_char(uint8_t a) {
...
 std::string Data_TextColumn::_escape_string(const CString& str) {
   ostringstream out;
+  bool allow_unicode = term_->unicode_allowed();
+
   size_t n = static_cast<size_t>(str.size);
-  for (size_t i = 0; i < n; ++i) {
-    auto c = static_cast<unsigned char>(str.ch[i]);
+  auto ch = reinterpret_cast<const unsigned char*>(str.ch);
+  auto end = ch + n;
+  for (;ch < end;) {
```
oleksiyskononenko: Why not just `while (ch < end)`?

st-pasha: evolution :)
You're right though, it should be `while`

st-pasha:
The build timed out again, so merging, no time to waster

PR #2127, opened 2019-10-25
Convert python integers into int32/int64 only during frame creation

### LINKED ISSUE #1702: Negative caclulations during join
```
from datatable import f, by, sum,g, join
DT = dt.Frame(product_id=[1,1,1,2,2,2,3,3,3], quantity=[11,22,16,45,65,60,33,37,39])
DT1 = dt.Frame(product_id=[1,2,3], price=[1,2,3])
DT1.key = "product_id"
DT[:, f.quantity * g.price,by(f.product_id), join(DT1)]
```

returns unexpected negative value at row 4

```
product_id	C0
▪	▪
0	1	11
1	1	22
2	1	16
3	2	90
4	2	−126
5	2	120
6	3	99
7	3	111
8	3	117
```

st-pasha:
(originally from @Viktor-Demin's PR https://github.com/h2oai/datatable/pull/1780, which was not  marked as editable, so cannot push updates directly into there).

Before this PR, `dt.Frame([1, 2, 3])` would create an `int8` column, and `dt.Frame([100, 200, 300])` had stype `int16`. After this PR both examples will create an `int32` column instead.

The reason for this change is that it is too error-prone: the user may not notice that int8/int16 columns are created, and cause an integer overflow during simple mathematical operations. With int32 the likelihood of an overflow is much less.

Closes #1702
Closes #1780


c/column_from_pylist.cc
```diff
@@ -199,7 +199,7 @@ static bool parse_as_int(const iterable* list, Buffer& membuf, size_t& from)
         outdata[i] = GETNA<T>();
         continue;
       }
-      if (item.is_int()) {
+      if (item.is_int() && sizeof(T) >= sizeof(int32_t)) {
```
oleksiyskononenko: It seems like with this change we will still do several iterations in `ocolumn_from_iterable()`, namely trying `SType::INT8` and `SType::INT16` as the `stype`, but we know in advance that these will not pass.

May be we just remove these two lines? https://github.com/h2oai/datatable/blob/master/c/column_from_pylist.cc#L542-L543

st-pasha: That's what I tried at first too, but turns out you need these cases for creating from  numpy integers.

st-pasha: there is a test "tests/test_dt_create.py::test_create_from_numpy_ints2"  which fails otherwise

oleksiyskononenko: ic, and we don't want to have the same kind of change for numpy?

st-pasha: If you have a list of `np.int8` scalars, then I think it's pretty unambiguous that you expect an `int8` column as a result. At the same time python integers don't have a fixed  storage size, so we're free to choose there.

st-pasha: It's integer overflow, not stack overflow :) 
For example, `int8_t(10) * int8_t(20) == int8_t(-56)`.

Note that in C any arithmetic operation involving `int8` or `int16` automatically promotes those into `int32`. Thus, adding or multiplying `int8`s can never overflow in C. Perhaps we could use a similar convention in datatable?

oleksiyskononenko: Yes, that's what I thought we're doing, but it seems that we're not. In such a case we can store data as `int8` or `int16` until we do any arithmetics on these columns.

PR #2128, opened 2019-10-25
[DO NOT MERGE] Fortify test_progress_interrupt

oleksiyskononenko:
Fortify `test_progress_interrupt()` by retrying several times with an increasing `sleep_time` in the main process.

### DIFF
--- a/tests/test_parallel.py
+++ b/tests/test_parallel.py
@@ -124,6 +124,12 @@ def test_progress(parallel_type, nthreads):
                             [1, dt.options.nthreads//2, dt.options.nthreads]
                          )
                         )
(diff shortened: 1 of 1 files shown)


oleksiyskononenko:
Closing this PR, because the problem appears to be not this test, but the way how we detect unicode support on some systems:

```
>>> import datatable as dt
>>> DT = dt.Frame(range(100))
>>> DT
  File "/root/datatable/datatable/utils/terminal.py", line 169, in _new_displayhook
    value.view(None)
UnicodeEncodeError: 'ascii' codec can't encode character '\u2026' in position 416: ordinal not in range(128)
```

PR #2131, opened 2019-10-28
Auto-detect terminal's encoding

st-pasha:
We will now detect the encoding of `sys.stdout` at datatable startup, and set the option `display.allow_unicode` only if that encoding is UTF-8. The user still has the possibility to change the option.

In addition, the option `display.allow_unicode` will no longer affect the behavior of `str(DT)`: when stringifying we allow unicode always.


oleksiyskononenko:
I've enabled `test_progress_interrupt()`, it should start working again with these changes.

PR #2133, opened 2019-10-29
Improve key handling for single column selectors

oleksiyskononenko:
A single column selector will now return a keyed one-column frame iff the column was the only key  in the original frame. 

WIP for #2098


tests/test_keys.py
```diff
@@ -164,3 +164,28 @@ def test_key_after_group():
...
+    assert DT_A.names == ("A",)
+    assert DT_A.to_list() == [list(range(100))]
+
+
+def test_key_kept_after_single_column_selector():
+    DT = dt.Frame([range(100), list(range(50))*2, list(range(25))*4],
+                  names = ["A", "B", "C"])
+    DT.key = ["A"]
+    DT_A = DT["A"]
+    frame_integrity_check(DT_A)
```
st-pasha: Can also check that `DT_B =  DT["B"]` is not keyed (i.e. nkeys==1 but the column selected is not first)

st-pasha:
Probably need a note in the Changelog as well, since this is a user-visible change

tests/test_keys.py
oleksiyskononenko: Yeah, wanted to add this test, but forgot.

PR #2162, opened 2019-11-08
Improve FTRL interactions handling

### LINKED ISSUE #2154: FTRL interactions property should accept various data structures
Right now we only allow `interactions` to be a list of lists of features. We should also allow other formats like list of tuples, tuple of tuples, etc. There is no reason to not allow it, this will also make it directly compatible with the output of `itertools.combinations()` that creates each combination in the form of a tuple.

oleksiyskononenko:
- allow various formats to be used for FTRL `.interactions`, i.e. a list or a tuple of lists or tuples of features;
- fix a bug when a `replace()` method was used on tuples and namedtuples, but the reference count for these objects was more than one;
- improve FTRL help and tests;
- minor cleaning.

Closes #2154


c/python/namedtuple.cc
```diff
@@ -132,4 +132,14 @@ onamedtuple::onamedtuple(const onamedtupletype& type) {
 }
 
 
+void onamedtuple::copy_v() {
```
st-pasha: `copy_v()` is quite a mysterious name, how about `prepare_for_editing()` or `make_editable()`? And also add a comment why do we need to jump through these hoops: because in python a tuple can only be modified if its ref-count is 1.

c/python/namedtuple.cc
```diff
@@ -132,4 +132,14 @@ onamedtuple::onamedtuple(const onamedtupletype& type) {
...
+  PyObject* v1 = PyTuple_GetSlice(v, 0, PyTuple_Size(v));
+  PyTypeObject* type = Py_TYPE(v);
+  Py_TYPE(v1) = type;
+  Py_INCREF(type);
+  Py_DECREF(v);
+  v = v1;
```
st-pasha: `Py_SETREF` macro is a safer alternative

c/python/tuple.cc
```diff
@@ -103,15 +111,26 @@ void otuple::set(size_t i, oobj&& value) {
...
+  if (v->ob_refcnt > 1) copy_v();
+  xassert(v->ob_refcnt == 1);
   // PyTuple_SetItem "steals" a reference to the last argument
   PyTuple_SetItem(v, static_cast<Py_ssize_t>(i), value.to_pyobject_newref());
 }
 
+
 void otuple::replace(size_t i, oobj&& value) {
+  if (v->ob_refcnt > 1) copy_v();
+  xassert(v->ob_refcnt == 1);
```
st-pasha: The refcount-check can be folded into the function too.

c/python/tuple.h
```diff
@@ -71,6 +71,10 @@ class otuple : public oobj {
      */
     void replace(size_t i,  const _obj& value);
     void replace(size_t i,  oobj&& value);
+    virtual ~otuple();
+
+    // Replaces `v` with a copy of `v`, that has `ob_refcnt == 1`.
+    virtual void copy_v();
```
st-pasha: This should probably be private? We don't anticipate this function to be used externally

tests/models/test_ftrl.py
```diff
@@ -122,13 +122,25 @@ def test_ftrl_construct_wrong_double_precision_type():
 
 def test_ftrl_construct_wrong_interactions_type():
     with pytest.raises(TypeError) as e:
-        noop(Ftrl(interactions = 2))
-    assert ("Argument `interactions` in Ftrl() constructor should be an iterable, "
-            "instead got <class 'int'>" == str(e.value))
+        noop(Ftrl(interactions = "C0"))
+    assert ("Argument `interactions` in Ftrl() constructor should be a list or a tuple, "
```
st-pasha: Why not use the new `match=` argument to `pytest.raises`?

tests/models/test_ftrl.py
```diff
@@ -287,7 +299,17 @@ def test_ftrl_set_individual():
...
+
+def test_ftrl_set_individual_after_params():
+    ft = Ftrl()
+    params = ft.params
+    ft.alpha = tparams.alpha
+    assert params == Ftrl().params
+    assert (ft.alpha, params.beta, params.lambda1, params.lambda2, params.nbins,
+           params.mantissa_nbits, params.nepochs, params.double_precision,
+           params.negative_class, params.interactions,
+           params.model_type) == ft.params
```
st-pasha: I think you missing a check that `ft.params` still remains a namedtuple

c/python/tuple.cc
```diff
@@ -20,10 +20,18 @@
...
+#include "utils/assert.h"
 
 namespace py {
 
 
+//------------------------------------------------------------------------------
+// Destructor
+//------------------------------------------------------------------------------
+
+otuple::~otuple() {}
```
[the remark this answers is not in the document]
oleksiyskononenko: tuple doesn't become a virtual class, it only gets one virtual function that is defined in the same class. Actually, I was not even going to use virtual functions, but I was like this is what you suggested when we discussed that yesterday?

c/python/namedtuple.cc
oleksiyskononenko: Sure, we can have a better name for it.

c/python/tuple.h
oleksiyskononenko: Yeah, I've been having some problems making this function virtual and was moving things back and forth, then forgot to make it protected. If we don't want to use virtual functions, we can make it private for sure.

tests/models/test_ftrl.py
oleksiyskononenko: For this I guess we need to make a separate PR by converting all the existing tests to the `match` functionality.

tests/models/test_ftrl.py
oleksiyskononenko: I'm doing that in `test_ftrl_create_params` and `test_ftrl_get_parameters`, but surely we can also do it here.

c/python/namedtuple.cc
oleksiyskononenko: Nice, can't see this macro anywhere in the docs, and it seems that we're also not using it in dt? The only place where there is some info is here https://github.com/python/cpython/blob/master/Include/cpython/object.h#L353-L370 Looks like a much better approach.

tests/models/test_ftrl.py
st-pasha: `test_ftrl_get_params` checks for a simple tuple, not a named tuple. And so does `test_ftrl_create_params`. This is because a namedtuple compares equals to a regular tuple if their components match.

Here I was merely suggesting to add an extra check that after assignment, the `.params` remain a named tuple (this could have failed if we'd forgotten to re-assign tuple's type in `copy_v()`)

c/python/namedtuple.cc
st-pasha: Yeah, I can't find it in the docs either... Weird.

c/python/tuple.cc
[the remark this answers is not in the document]
st-pasha: A class becomes virtual whenever it gets even a single virtual method.
I agree that this is what we discussed yesterday, but I didn't realize then that it would force the class to become virtual. Still, it's quite simple to avoid virtual functions here: see my suggestion in a comment below.

tests/models/test_ftrl.py
st-pasha: As you wish. I was merely thinking that since you're updating the tests anyways, might as well upgrade their syntax too.

c/python/tuple.cc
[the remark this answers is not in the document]
oleksiyskononenko: Sorry, I confused it with the abstract class. There is no problem for me to avoid virtual classes/functions.

tests/models/test_ftrl.py
oleksiyskononenko: Yes, I noticed that at some point but looks like never addressed in `test_ftrl_get_params` and `test_ftrl_create_params`. This check should also be there. It looks like the easiest way to do this kind of check is just to add component-wise asserts.

PR #2167, opened 2019-11-11
Fix creating a Frame from a numpy array with negative stride

### LINKED ISSUE #2163: Creating Frame from a reversed numpy array fails
```
>>> import numpy as np
>>> import datatable as dt
>>> dt.Frame(np.array(range(10))[::-1])
  File "/Users/pasha/github/datatable/datatable/utils/terminal.py", line 169, in _new_displayhook
    value.view(None)
AssertionError: Assertion 'i < nrows()' failed in c/column.cc, line 188
```

st-pasha:
When a stride is negative, the calculation of buffer size as `nrows * stride * elemsize` was producing incorrect results, and creating a buffer object from this was failing internal consistency checks. Now the case of negative stride is handled properly: we use `|stride|` to compute the buffer's size, but then change the `start` from 0 to the end of the buffer when applying a SliceView to that buffer.

The case `nrows == 0` is now also handled correctly (cannot have a non-null pointer to a memory region of size 0).

Closes #2163

### DIFF
--- a/tests/test_dt_create.py
+++ b/tests/test_dt_create.py
@@ -28,6 +28,7 @@
 import math
 import os
 import pytest
+import random
 import datatable as dt
 from datatable import ltype, stype, DatatableWarning
 from datatable.internal import frame_integrity_check
@@ -1109,6 +1110,20 @@ def test_create_from_numpy_floats_mixed(numpy):
 def test_create_from_numpy_reversed(numpy):
     DT = dt.Frame(numpy.array(range(10))[::-1])
     assert_equals(DT, dt.Frame(range(9, -1, -1), stype=dt.int64))
+
+
+@pytest.mark.parametrize("seed", [random.getrandbits(32) for _ in range(100)])
+def test_create_from_numpy_sliced(numpy, seed):
+    random.seed(seed)
+    start = random.randint(-20, 20)
+    end = random.randint(-20, 20)
+    step = 0
+    while step == 0:
+        step = random.randint(-5, 5)
+    arr = numpy.arange(20)[start:end:step]
+    DT = dt.Frame(arr)
+    assert_equals(DT, dt.Frame(arr.tolist(), stype=dt.int64))
+
 
 
 #-------------------------------------------------------------------------------
--- a/tests/test_dt_create.py
+++ b/tests/test_dt_create.py
@@ -1106,6 +1106,10 @@ def test_create_from_numpy_floats_mixed(numpy):
     assert DT.to_list() == [[4.0, 3.5, None, 9.33]]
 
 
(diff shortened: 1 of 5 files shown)


tests/test_dt_create.py
```diff
@@ -1106,6 +1107,24 @@ def test_create_from_numpy_floats_mixed(numpy):
...
+def test_create_from_numpy_sliced(numpy, seed):
+    random.seed(seed)
+    start = random.randint(-20, 20)
+    end = random.randint(-20, 20)
+    step = 0
+    while step == 0:
+        step = random.randint(-5, 5)
+    arr = numpy.arange(20)[start:end:step]
+    DT = dt.Frame(arr)
+    assert_equals(DT, dt.Frame(C0=arr.tolist(), stype=dt.int64))
```
oleksiyskononenko: Do we need to add randomness to our unit tests? I've been thinking that the only source of randomness would be a random attacker, and in unit tests we just try all the reasonable use-cases. Like here we just do a simple loop over the reasonable `start`, `end` and `step` values. Not all the 41x41x10 options you have now, but may be 10x10x5, or something like this to make it fast.

st-pasha: I don't see how this could hurt. If a test fails, we'll have the seed to reproduce the problem. This is still a unit test too: it's focused on creating a frame from a slice numpy array, with various kinds of slices. The random attacker would have a broader task: not only to check that the frame is created successfully, but also that this frame can later be used in various operations without problems.

oleksiyskononenko: Let's say there is a bug with the negative slice step, you run this test and, by chance, it doesn't fail. Or, there is more rare bug when number of elements in the slice is one — who knows how many times one needs to run this test to catch it. To me, it seems more reasonable just to cover all these options in several loops rather than picking up one particular profile randomly.

st-pasha: The test has `seed` argument, which already implies that it's random.
And it's ok if the test doesn't catch the rare bug on the first try. This is actually a desirable outcome: after all, Jenkins will run the same test over and over again, and eventually the bug will be found. The logs will even show which seed caused the problem, allowing us to reproduce and fix the issue quickly.

oleksiyskononenko: I agree on the rare bugs that are complicated to be caught. However, here most of the important use-cases can be covered with [making the step positive or negative] times [having zero, one or some arbitrary number of elements in the slice], so in total just six basic profiles that can be tested pretty fast. If I'm developing something related to numpy and doing `make test`, I wouldn't want to miss these profiles due to randomness. The rest can come as a bonus from Jenkins, of course.

PR #2182, opened 2019-11-14
Fix ellipsis column when displaying frame in a terminal

### LINKED ISSUE #2153: Displaying a frame sometimes lacks elipsis to indicate omitted columns
Terminal size: `151x45`
Host: `mr-dl11`
Dataset: `airlines.csv`

image

Note, that there should be `...` in between the columns `CRSArrTime` and `SecurityDelay`.

st-pasha:
Closes #2153


oleksiyskononenko:
I confirm that it now works as expected:
image

PR #2183, opened 2019-11-14
fread() can read Jay files, open() deprecated

### LINKED ISSUE #2138: fread could open jay files
`fread` could recognise `jay` extension and use `open` method instead. Now it raise exception.

st-pasha:
This PR introduces two changes:

1. The function `dt.fread()` is now able to read Jay files (as detected by their extension). By extension, it means `dt.Frame(jay_file)` will now work too.

2. The `fread()` function is promoted as the **primary** method for reading jay files. Consequently, the `dt.open()` function is now deprecated, and will emit a warning when used. It is also scheduled for future removal in version 0.12.

The reason to remove `dt.open()` is that the name clashes with standard python `open()` function, which may create problems if `dt.open` is imported in user's namespace.

Closes #2138


st-pasha:
The last test timed out after 2 hours because it waited in queue for too long, so I'm going to merge regardless

PR #2191, opened 2019-11-18
Binary operators implemented using virtual columns

### LINKED ISSUE #2002: Binary operators need to use virtual columns
Currently, binary operators generate "material" columns, which is inefficient.

st-pasha:
The primary achievement of this PR is that all binary operators (such as `+`, `-`, `%`, `&`, etc) are now implemented with the modern virtual columns mechanism, which avoids (a) materialization of lhs and rhs columns, (b) materialization of the resulting column.

In addition,

- the number of template instantiations drastically improved: we no longer create templates for all possible combinations of lhs/rhs stypes, instead we implement them for common stypes only and rely on upcasts for all other cases;

- most arithmetic operations will now upcast their arguments to at `int32` if the arguments are `bool`, `int8`, or `int16`. For example, if X and Y are both int8, then the result of `X+Y` will now be
int32 instead of int8;

- operator `&` (AND) for boolean inputs will now produce `false` if either argument is `false` (previously `false & NA` was `NA`). In addition, this operator now short-circuits: `x & y` will evaluate `x`, and if it is `false` then the result will be also `false` without evaluating `y`;

- operator `|` (OR) for boolean inputs will now produce `true` if either argument is `true` (previously `true | NA` was `NA`). In addition, this operator now short-circuits: `x | y` will evaluate `x`, and if it is `true` then the result will be also `true` without evaluating `y`;

- operators `&`, `|` and `^` (XOR) now work on integer arguments too;

- implemented operator `**` (power), which wasn't implemented previously.

Closes #2002


st-pasha:
(ignoring the Travis error for now)

PR #2193, opened 2019-11-19
Simplify the way Aggregator and FTRL calculate number of threads for parallelization

oleksiyskononenko:
- add `nthreads_from_niters()` function that returns an `NThreads` object with the number of threads calculated from the number of iterations and the minimum number of iterations a thread should process to be efficient;
- make both the Aggregator and FTRL to use this function, and get rid of some duplicated code;
- make use of this function in `parallel/string_utils` fixing minor bugs when we didn't handle a single-threaded mode correctly;
- since now there is a `dt.options.nthread` option in Python that controls number of threads dt-wise, `nthreads` parameter in the `aggregate()` function became useless and was removed;
- minor code clean-up.

### DIFF
--- a/c/parallel/api_primitives.h
+++ b/c/parallel/api_primitives.h
@@ -49,16 +49,6 @@ class NThreads {
       size_t n0 = num_threads_in_pool();
       value = (nth == 0 || nth >= n0) ? n0 : nth;
     }
-    // Calculate number of threads from the number of iterations
-    // and the minimum number of iterations per thread.
-    explicit NThreads(size_t niters, size_t min_iters_per_thread) {
-      if (niters > min_iters_per_thread) {
-        size_t n0 = num_threads_in_pool();
-        value = std::min(niters / min_iters_per_thread, n0);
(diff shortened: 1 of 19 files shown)


c/parallel/api_primitives.h
```diff
@@ -49,6 +49,17 @@ class NThreads {
       size_t n0 = num_threads_in_pool();
       value = (nth == 0 || nth >= n0) ? n0 : nth;
     }
+    // Calculate number of threads from the number of iterations
+    // and the minimum number of iterations per thread.
+    explicit NThreads(size_t niters, size_t min_iters_per_thread) {
```
st-pasha: I don't think that's a good API. This will lead to code such as `NThreads(n, 10)`, and that will totally confuse the reader... Besides, I'm not even sure that the NThreads class should know anything about any iterations

oleksiyskononenko: For my purpose it doesn't really matter, so I've converted the constructor into a separate function.

PR #2196, opened 2019-11-23
Prevent fread emitting message about BufferedIO file not closed

st-pasha:
`proc.stdout.read()` was not closing the stdout/stderr parts of the Popen object, causing the warnings.


oleksiyskononenko:
Just checked, yes, the fix helps.

PR #2197, opened 2019-11-23
Fix segfault when C++ tests were called with missing arguments

oleksiyskononenko:
Even though we do not expect users to call internal `datatable.lib.core` C++ tests, they should not segfault when called with missing arguments. In this PR we add a `PKArgs::check_required_args()` method that checks if the required arguments are missing and improve`test_parallel.py` to make sure C++ tests do not segfault.

### DIFF
--- a/c/ztest.cc
+++ b/c/ztest.cc
@@ -34,7 +34,7 @@ static PKArgs arg_test_shmutex(3, 0, 0, false, false,
   "test_shmutex");
 
 static void test_shmutex(const PKArgs& args) {
-  args.check_required_args();
+  args.check_posonly_args();
   size_t n_iters = args[0].to_size_t();
   size_t n_threads = args[1].to_size_t();
   int impl = args[2].to_int32_strict();
@@ -53,7 +53,7 @@ static PKArgs arg_test_barrier(
   1, 0, 0, false, false, {"n"}, "test_barrier");
 
 static void test_barrier(const PKArgs& args) {
-  args.check_required_args();
+  args.check_posonly_args();
   size_t n = args[0].to_size_t();
   dttest::test_barrier(n);
 }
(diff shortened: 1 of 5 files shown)


c/ztest.cc
```diff
@@ -27,6 +46,7 @@ static PKArgs arg_test_shmutex(3, 0, 0, false, false,
   "test_shmutex");
 
 static void test_shmutex(const PKArgs& args) {
+  check_posonly_args(arg_test_shmutex, args);
```
st-pasha: `args` and `arg_test_shmutex` are actually exactly one and the same object here

c/ztest.cc
```diff
@@ -1,18 +1,37 @@
...
-#include "python/args.h"
 #include "utils/exceptions.h"
 #include "ztest.h"
 namespace py {
 
 
+/**
+ * Helper function to check for missing positional arguments
+ */
+static void check_posonly_args(const PKArgs& args_spec, const PKArgs& args) {
```
st-pasha: This could be a method on `PKArgs`. In fact, there could even be a property such as `num_required_args`, which would check those arguments automatically, though I'm afraid it'll make the constructor even more complicated

c/ztest.cc
oleksiyskononenko: Oh, didn't notice that :)

c/ztest.cc
oleksiyskononenko: Are there any cases when number of arguments received is less than `n_posonly_args`?

st-pasha: technically any arguments (including positional) can be optional

oleksiyskononenko: Btw, what about also adding a check to `PKArgs::operator[]`?
```
const Arg& PKArgs::operator[](size_t i) const {
  //check here
  return bound_args[i];
}
```
It seems that this is a source of many segfaults we've seen in the past and also the issue with these cpp tests.

st-pasha: Can add an `xassert()` there

PR #2198, opened 2019-11-23
Fix test_horizontal_elision on Travis

st-pasha:
The test fails on Travis because their terminal width is 80 instead of 120.
This PR fixes the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,17 @@ before_install:
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


oleksiyskononenko:
Title says "Disable test_horizontal_elision on Travis", but it doesn't seem to be disabled, more like fixed. Should we change "Disable" to "Fix","Adjust" or "Improve"?

PR #2203, opened 2019-11-25
Move monitor thread from idle_job to thread_pool

oleksiyskononenko:
Creating monitor thread on the datatable import causes a deadlock on Windows, see this [discussion](https://stackoverflow.com/questions/32252143/stdthread-cause-deadlock-in-dllmain) for more details. In fact, we don't really need the monitor thread being available right after the `import datatable`, so its creation can be easily postponed. 

In this PR we 
- move the monitor handler from `idle_job` to `thread_pool` class, and create the monitor thread only when it is needed by datatable;
- disable fork related code on Windows, because there is no fork mechanism there.

WIP for #1369

### DIFF
--- a/c/parallel/thread_pool.cc
+++ b/c/parallel/thread_pool.cc
@@ -160,8 +160,7 @@ void thread_pool::enable_monitor(bool a) noexcept {
 
 
 bool thread_pool::is_monitor_enabled() noexcept {
-  init_monitor_thread();
-  return monitor->get_active();
+  return monitor? monitor->get_active() : false;
 }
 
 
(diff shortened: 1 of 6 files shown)


c/parallel/thread_pool.cc
```diff
@@ -132,28 +153,33 @@ thread_team* thread_pool::get_team_unchecked() noexcept {
...
+  init_monitor_thread();
+  monitor->set_active(a);
 }
 
 
-bool thread_pool::is_monitor_enabled() const noexcept {
-  return controller.is_monitor_enabled();
+bool thread_pool::is_monitor_enabled() noexcept {
+  init_monitor_thread();
+  return monitor->get_active();
```
st-pasha: don't need to init monitor here: if there is no monitor then it is definitely "not enabled"

oleksiyskononenko: Yeah, forgot to add just a simple check here.

PR #2232, opened 2019-12-11
New Xbuild module

st-pasha:
This module offers an alternative way for building the `_datatable` extension, instead of `setuptools`. It has several advantages over the traditional way:
  - There is no automatic adding of compile/link flags: the consumer of the module is in full control over how the compilation is done;
  - The build process is performed in parallel, reducing the compile time by a factor of 10 or so (similar to make);
  - Only files that need to be recompiled are recompiled (like make);
  - The implementation is pure-python (unlike make), which makes it easy to modify the code, and to reuse it from other python processes;
  - Allow the code to be automatically recompiled during python import;
  - Better log of the build process (and fully customizable);
  - Support fast recompiling in different modes (without clearing the previous build);
  - Automatic recompilation when compile flags or python version change.

### DIFF
--- a/ci/xbuild/extension.py
+++ b/ci/xbuild/extension.py
@@ -1,0 +1,678 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#-------------------------------------------------------------------------------
+# Copyright 2019 H2O.ai
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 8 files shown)


ci/xbuild/extension.py
```diff
@@ -0,0 +1,679 @@
...
+            # Given a queue of workers (`subprocess.Popen` processes),
+            # wait until at least one of them finishes and return that
+            # worker, and also remove it from the `queue`.
+            def await_one(queue):
+                while queue:
+                    for i, worker in enumerate(queue):
+                        if worker.poll() is not None:
+                            del queue[i]
+                            return worker
+                    time.sleep(0.1)
```
oleksiyskononenko: So it seems that when we have a lot of output from the subprocess, this loop becomes infinite. `worker.poll()` doesn't deadlock, it just never becomes equal to`None`. For example, on Windows I see the following 7 subprocesses that never finish according to `.poll()`:

```
c\sort.cc <subprocess.Popen object at 0x00000210CE6F1F98>
c\column\sentinel_fw.cc <subprocess.Popen object at 0x00000210CE6F1E10>
c\column\sentinel_str.cc <subprocess.Popen object at 0x00000210CE6F1C18>
c\frame\rbind.cc <subprocess.Popen object at 0x00000210CE6F1F28>
c\frame\__sizeof__.cc <subprocess.Popen object at 0x00000210CE6F1D30>
c\models\aggregator.cc <subprocess.Popen object at 0x00000210CE6F1BE0>
c\models\dt_ftrl.cc <subprocess.Popen object at 0x00000210CE6F13C8>
```

In reality these subprocesses did produce the `.obj` files (and I guess stopped).
What is also true is that these files produce a lot of warnings, so the pipe could cause some problems.

st-pasha: Yep, that's exactly what I was talking about yesterday -- the pipe fills up and the process dead-locks. However, that's only if a single file emits voluminous amount of warnings, which normally doesn't happen. Still, we should fix this in pr `xbuild2`

oleksiyskononenko: Yes, I thought you've been using `.wait()` instead of `.communicate()` somewhere, but it seems that even `.poll()` doesn't work as expected in such cases.

oleksiyskononenko: May be we also add a variable that controls the maximum time allowed for compiling a single file. Even when we solve the pipe problem, the loop can potentially become infinite if compilation hangs at some stage.

st-pasha: Never heard of a compiler hanging

oleksiyskononenko: May not be a direct compiler hang, but if the `while` loop can go infinite, it will at some point :)

PR #2236, opened 2019-12-12
Fix min/max properties of bool8 stype

### LINKED ISSUE #2231: stype.bool8.min/max are not boolean
`stype.bool8.min` returns 0, which is not a suitable value to be put into a boolean column. This prevents some generic code that creates a column from "min" value of each stype from working.

st-pasha:
(We may want to include this into 0.10.1 release as well)

Closes #2231

### DIFF
--- a/docs/changelog/v-0-11-0.rst
+++ b/docs/changelog/v-0-11-0.rst
@@ -9,5 +9,20 @@ Frame
 -----
 
 - |fix| A list of frames now displays properly in a Jupyter lab (#2222).
+
+
+
+General
+-------
+
+- |api| Properties ``dt.stype.min`` and ``dt.stype.max`` are now equal to
+  ``False`` and ``True`` respectively, instead of integers 0 and 1 (#2231).
+
+
+
+FTRL model
+----------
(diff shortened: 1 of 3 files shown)


docs/changelog/v-0-11-0.rst
```diff
@@ -9,5 +9,20 @@ Frame
...
 
 - |fix| A list of frames now displays properly in a Jupyter lab (#2222).
+
+
+
+General
+-------
+
+- |api| Properties ``dt.stype.min`` and ``dt.stype.max`` are now equal to
+  ``False`` and ``True`` respectively, instead of integers 0 and 1 (#2231).
```
oleksiyskononenko: Should we add "For boolean columns..."?

st-pasha: You right, it's a typo

st-pasha: *You're

what's up with the typos today? 😉

st-pasha:
No point in waiting for PPC, going to merge

PR #2237, opened 2019-12-13
updated version of pandas

abal5:


st-pasha:
Does the fact that "Build image for ppc64le" stage failed again mean this fix didn't quite work?

oleksiyskononenko:
@abal5 It still fails to build datatable on ppc...

PR #2238, opened 2019-12-13
Add a code to detect terminal size on Windows

oleksiyskononenko:
- add a code to detect terminal size on Windows;
- there is no simple way to catch the terminal re-size event on Windows, because there is no `SIGWINCH` signal there. For such a reason, on WIndows we just re-check the terminal size everytime the `Terminal::get_width()` and `Terminal::get_height()` are called.

WIP for #1369

### DIFF
--- a/c/utils/terminal/terminal.cc
+++ b/c/utils/terminal/terminal.cc
@@ -187,18 +187,18 @@ void Terminal::_detect_window_size() {
   CONSOLE_SCREEN_BUFFER_INFO csbi;
   HANDLE h = GetStdHandle(STD_OUTPUT_HANDLE);
   int ret = GetConsoleScreenBufferInfo(h, &csbi);
-  bool is_size_detected = ret > 0;
+  bool is_size_not_detected = (ret == 0);
   width_ = csbi.srWindow.Right - csbi.srWindow.Left + 1;
   height_ = csbi.srWindow.Bottom - csbi.srWindow.Top + 1;
 #else
   struct winsize w;
   int ret = ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
-  bool is_size_detected = ret != -1;
+  bool is_size_not_detected = (ret == -1);
   width_ = w.ws_col;
   height_ = w.ws_row;
 #endif
 
-  if (!is_size_detected || width_ == 0) {
+  if (is_size_not_detected || width_ == 0) {
     width_ = 120;
     height_ = 45;
   }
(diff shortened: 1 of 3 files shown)


c/utils/terminal/terminal.cc
```diff
@@ -163,11 +183,22 @@ void Terminal::forget_window_size() {
...
 
 void Terminal::_detect_window_size() {
+#if DT_OS_WINDOWS
+  CONSOLE_SCREEN_BUFFER_INFO csbi;
+  HANDLE h = GetStdHandle(STD_OUTPUT_HANDLE);
+  int ret = GetConsoleScreenBufferInfo(h, &csbi);
+  bool is_size_not_detected = (ret == 0);
+  width_ = csbi.srWindow.Right - csbi.srWindow.Left + 1;
+  height_ = csbi.srWindow.Bottom - csbi.srWindow.Top + 1;
+#else
```
st-pasha: macros should be indented the same way as if they were regular C++ `if` statements, in order to allow proper code folding.

c/utils/terminal/terminal.cc
```diff
@@ -143,12 +155,20 @@ bool Terminal::unicode_allowed() const noexcept {
 }
 
 int Terminal::get_width() {
+#if DT_OS_WINDOWS
+  _detect_window_size();
+#else
```
st-pasha: One problem here is that if i needed both width&height, then `_detect_window_size()` would be called twice.

Possible solution is to reset the `width_` to 0 after returning it, then `height_` will remain with a proper value which can be queried (and autoreset) too. Something like
```C++
int Terminal::get_width() {
  if (!width_) _detect_window_size();
  int w = width_;
  #if DT_OS_WINDOWS
    width_ = 0;
  #endif
  return w;
}
```

c/utils/terminal/terminal.cc
oleksiyskononenko: Sure, I've adjust the macros indentation. Don't use the folding much...

c/utils/terminal/terminal.cc
oleksiyskononenko: Calling `_detect_window_size()` is pretty cheap, but if we want to make it optimal I would just propose to create a single method called `get_size()`: it will return a structure with `.width` and `.height` fields. 

Btw, I don't see we're using `get_heigh()` anywhere in the code.

st-pasha: Yeah, I guess `get_height()` isn't nearly as useful as it looks...

oleksiyskononenko:
ppc build still fails, so merging...

PR #2244, opened 2019-12-17
Fix path in zlib headers

oleksiyskononenko:
Add `lib/zlib/` path to zlib includes, otherwise the code doesn't compile on some systems.

WIP for #1369


st-pasha:
Won't be needed once we use `xbuild` exclusively, but ok for now

PR #2255, opened 2019-12-19
Frame from list

### LINKED ISSUE #2250: Avoid exceptions when constructing a Frame from python list
Currently we're using exceptions in order to signal a type-bump, which means at least several exceptions will be thrown (and caught) when a python list is converted into a frame. We found that on some architectures (ppc64le) and with older compilers (pre- gcc-5) this could be problematic: too many exceptions may cause a deadlock, for example such as in this backtrace:
```
#0  0x00003fffa69d2da8 in __lll_lock_wait () from target:/usr/lib64/libpthread.so.0
#1  0x00003fffa69cb750 in pthread_mutex_lock () from target:/usr/lib64/libpthread.so.0
#2  0x00003fffa6801e4c in dl_iterate_phdr () from target:/usr/lib64/libc.so.6
#3  0x00003fffa47808d4 in _Unwind_Find_FDE () from target:/usr/lib64/libgcc_s.so.1
#4  0x00003fffa477b3ec in ?? () from target:/usr/lib64/libgcc_s.so.1
#5  0x00003fffa477bf28 in ?? () from target:/usr/lib64/libgcc_s.so.1
#6  0x00003fffa477c920 in _Unwind_RaiseException () from target:/usr/lib64/libgcc_s.so.1
#7  0x00003fff6053ba94 in __cxa_throw () from target:/usr/lib64/libstdc++.so.6
#8  0x00003fff5f37f86c in py::_obj::to_bool (this=0x3fffddbf5b70, em=...) at c/python/obj.cc:249
#9  0x00003fff5f0fbec8 in parse_as_bool (from=<synthetic pointer>, membuf=..., list=0x3fffddbf60f0) at c/column_from_pylist.cc:131
#10 Column::from_py_iterable (il=0x3fffddbf60f0, stype0=<optimized out>) at c/column_from_pylist.cc:527
#11 0x00003fff5f0fd744 in Column::from_pylist (list=..., stype0=<optimized out>) at c/column_from_pylist.cc:557
#12 0x00003fff5f3002b8 in dt::expr::expr_literal::expr_literal (this=0x10041650390, v=...) at c/expr/expr_literal.cc:30
#13 0x00003fff5f2fe4a4 in py::_obj::to_dtexpr (this=<optimized out>) at c/expr/expr.cc:169
#14 0x00003fff5f2fecf0 in dt::expr::make_binop (op=dt::expr::Op::LT, args=...) at c/expr/expr.cc:104
#15 0x00003fff5f2fe664 in py::_obj::to_dtexpr (this=<optimized out>) at c/expr/expr.cc:177
#16 0x00003fff5f1ff9ac in dt::expr_in::expr_in (this=0x1003e04d280, src=...) at c/expr/i_node.cc:273
#17 0x00003fff5f201730 in dt::_make (src=...) at c/expr/i_node.cc:612
#18 0x00003fff5f202590 in dt::i_node::make (src=..., wf=...) at c/expr/i_node.cc:648
#19 0x00003fff5f2f8bc0 in dt::workframe::add_i (this=0x3fffddbf6bb0, oi=...) at c/expr/workframe.cc:75
#20 0x00003fff5f335034 in py::Frame::_main_getset (this=0x3fff512b1db8, item=..., value=...) at c/frame/__getitem__.cc:206
#21 0x00003fff5f335bc8 in py::Frame::m__setitem__ (this=0x3fff512b1db8, item=..., value=...) at c/frame/__getitem__.cc:83
#22 0x00003fff5f33ff60 in py::_safe_setitem<py::Frame, &py::Frame::m__setitem__> (self=0x3fff512b1db8, key=<optimized out>, val=0x3ffd69e28d68) at c/python/xobject.h:369
#23 0x00003fffa6a926b8 in PyObject_SetItem (o=<optimized out>, key=<optimized out>, value=<optimized out>) at Objects/abstract.c:180
```

### LINKED ISSUE #2251: A list of integers should always construct an integer column
Currently when a list contains only 0s and 1s we create a boolean column, which is counterintuitive to the point that some people assume it must be a bug.

We propose to change this behavior, and create an int8 column instead. Thus:
```python
# Current behavior
assert dt.Frame([0, 1]).stype == dt.bool8

# New behavior
assert dt.Frame([0, 1]).stype == dt.int8
```

This is a breaking API change; however the support for it seems to be universal.

st-pasha:
- Reorganize code in "column_from-python.cc" so that it no longer relies on exceptions for type-bumping.

- Lists containing integers 0 and 1 will now produce columns of type `int8` (instead of `bool8`)

- Make code more robust in what types are accepted: all stypes should be able to accept values for lower stypes. For example, `bool` is considered lower than `int8`, and therefore an `int8` list will now accept `bool`s mixed with `int8` values.

Closes #2250 
Closes #2251

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,16 +92,16 @@ test:
 		tests
 
 xasan:
-	$(PYTHON) ext.py asan
+	@$(PYTHON) ext.py asan
 
 xbuild:
-	$(PYTHON) ext.py build
+	@$(PYTHON) ext.py build
 
 xcoverage:
-	$(PYTHON) ext.py coverage
+	@$(PYTHON) ext.py coverage
 
 xdebug:
-	$(PYTHON) ext.py debug
+	@$(PYTHON) ext.py debug
 
 
 benchmark:
--- a/c/column/pysources.cc
+++ b/c/column/pysources.cc
@@ -1,0 +1,102 @@
+//------------------------------------------------------------------------------
+// Copyright 2019 H2O.ai
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#include "column/pysources.h"
+#include "python/dict.h"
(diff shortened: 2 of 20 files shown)


surenH2oai:
how do we know this fixes https://github.com/h2oai/h2oai/issues/12645 @st-pasha Thanks.

st-pasha:
@surenH2oai  The reason for the stall was because of a deadlock in `__cxa_throw()` standard library function, i.e. the function which is used when an exception is thrown.
In this PR we changed the code so that the exception is no longer thrown, and therefore the problematic call to `__cxa_throw()` is avoided.

PR #2259, opened 2019-12-20
Improve handling of .labels in FTRL

oleksiyskononenko:
- use `SType::INT32` for `id` column in `.labels` for all the model types as this makes the code more consistent and much cleaner. Since for numeric and binomial regressions we only have one or two labels this change has a negligible effect on the memory usage;
- improve tests.

### DIFF
--- a/c/models/dt_ftrl.cc
+++ b/c/models/dt_ftrl.cc
@@ -759,36 +759,10 @@ void Ftrl<T>::update(const uint64ptr& x,
 
 
 /**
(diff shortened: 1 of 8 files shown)


st-pasha:
Do we want this for DAI 1.8.2, or wait with the merge?

oleksiyskononenko:
@st-pasha just checked with @arnocandel and we are good to go. So let's merge it and DAI will pick it up.

PR #2264, opened 2019-12-23
Allow FTRL training for a fractional number of epochs

### LINKED ISSUE #2263: FTRL model should allow training for a fractional number of epochs
The type of the FTRL `nepochs` parameter should be changed from `int` to `float` allowing training for a fractional number of epochs. 

Since early stopping in `.fit()` can be triggered in multiples of  `nepochs_validation` and this parameter has a `float` type — there is no reason to prevent training with no validation for explicitly specified fractional number of epochs .

oleksiyskononenko:
Closes #2263


c/models/dt_ftrl.cc
```diff
@@ -556,16 +558,20 @@ FtrlFitOutput Ftrl<T>::fit(T(*linkfn)(T),
 
   // Training settings. By default each training iteration consists of
   // `dt_X_train->nrows` rows.
-  size_t niterations = params.nepochs;
+  size_t niterations = static_cast<size_t>(std::ceil(nepochs));
+  T frac_nepochs = nepochs - niterations + 1;
```
st-pasha: if `nepochs` is integer, say `3`, then `niterations = 3` too, and then `frac_nepochs = 1`. Is this what you intended?

c/models/py_ftrl.cc
```diff
@@ -1200,7 +1202,7 @@ mantissa_nbits : int
     Number of bits from mantissa to be used for hashing floats,
     defaults to `10`.
 
-nepochs : int
+nepochs : float
     Number of training epochs, defaults to `1`.
```
st-pasha: Perhaps add an explanation here what does it mean for `nepochs` to be fractional

c/models/dt_ftrl.cc
oleksiyskononenko: Yeah, but the variable names I've used here could really cause some confusion. I've streamlined this part and added more comments.

c/models/py_ftrl.cc
oleksiyskononenko: Sure, done.

Issue #290, opened 2017-08-30
Roadmap


st-pasha:
- [x] Rollup stats for each column (#276)
- [x] Write a CSV file
- [x] Generalized computational graph (#1477)
- [x] Grouping by a single column (integer type)
- [x] Group-by reduction operators: sum, min, max, mean, etc.
- [x] Grouping by a single column (other types)
- [x] Modify data in a datatable in-place
- [x] Sorting by multiple columns
- [x] Grouping by multiple columns
- [ ] Joining (inner, left outer)
- [x] Insert / delete rows
- [x] Computation on variable-width columns (strings)
- [ ] String operations
- [ ] Rolling join
- [ ] Group-by operators that involve sorting: median, quantile, rank (median also in https://github.com/h2oai/datatable/issues/1530)
- [x] rbind disk-based columns
- [x] Keys to optimize performance
- [ ] Non-equi joins
- [ ] Compute rollup stats during fread
- [ ] Categorical (enum) type
- [ ] Datetime type
- [ ] Time window functions
- [ ] Full join
- [ ] fread auto-detected parse format save into FReader object
- [ ] skipna parameter for reduction operators
- [ ] Repro of NYCTaxi analytics using datatable
- [ ] Benchmark Szilard's test (grouping vs joining)
- [ ] Benchmark against fst & paratext
- [x] Benchmark parity with R data.table 5 grouping tests - https://h2oai.github.io/db-benchmark/
- [ ] Benchmark against TPC-H
- [ ] Finish fread outstanding bug master list ([#2247](https://github.com/Rdatatable/data.table/issues/2247))
- [ ] Comparison of fread capabilities against other read csv frameworks (esp. smart capabilities)
- [ ] Demo of interoperability with other frameworks
- [ ] Log operations with datatable (data provenance)
- [ ] Read/write into feather (#1461)

**Out of scope**
- Multi-user access, including permissions
- Sharded data files, growing file size

jangorecki:
Marked _Benchmark parity with R data.table 5 grouping tests_ as resolved.
What about `cor`, `cov`, `var` functions, are they going to be in scope? Asking because `cor` or `cov+var` is required to answer q9 in groupby benchmark.

st-pasha:
@jangorecki  Of course, `cor`, `cov`, `var` are all quite straightforward, no reason not to add them.

st-pasha:
Replaced with #2281

mdancho84:
[...]

PR #2274, opened 2019-12-30
Add comparison operators for strings (<, >, <=, >=)

bboe:
This pull request addresses some of the `String operations` tasks in #290.


bboe:
@st-pasha any thoughts here?

st-pasha:
@bboe Looks solid, however, let me test it locally too.

bboe:
Thanks for the review and merging. And thanks for the awesome tool. I am thoroughly impressed by how quickly it parses CSV files (it was the fastest tool I've tried), and now I can quickly perform simple date-string filtering (assuming timezones are normalized in the file).

st-pasha:
@bboe And thank **you** for contributing as well! 
Note that for the next version we've planned further improvements to CSV parser, string functions, and even dedicated date-time column type: see the roadmap at #2281. So I hope that you will have even more reasons to like `datatable` in the future!

Issue #2258, opened 2019-12-19
Use PEP-517/518-compliant build process


st-pasha:
[PEP-518](https://www.python.org/dev/peps/pep-0518/) introduces new format for distributing python packages: based on `pyproject.toml` configuration file. In addition, [PEP-517](https://www.python.org/dev/peps/pep-0517/) describes the API for a "build backend" capable of creating source distributions and binary wheels.

This epic therefore consists of:
- [x] Create a new system (`xbuild`) for building the binary extensions (#2235, #2246);
- [x] Implement wheel-building functionality (#2247);
- [x] Implement creating source distributions in xbuild (#2257);
- [x] Use auditwheel to verify wheel integrity on Linux distribution (#2266);
- [x] Deploy `xbuild` as the primary tool for CI pipelines (#2276);
- [x] Generate version string and version files in ext.py (#2287);
- [x] Remove old code used for building datatable.

st-pasha:
So it appears we need `pip >= 20.0` for this, which will be released in Jan 2020. 
For now, you'd need to do `pip install git+https://github.com/pypa/pip` to get the bleeding edge version of `pip`.

PR #2287, opened 2020-01-07
[WIP] Clean up datatable version generation

### LINKED ISSUE #2265: [CI] Switch to manylinux official docker images
Instead of using internally defined Docker image we should use official docker image for manylinux https://github.com/pypa/manylinux.

st-pasha:
The following changes are introduced:

- Files `datatable/__version__.py` and `datatable/__git__.py` are removed, replaced with a single auto-generated file `datatable/_build_info.py`;

- The file `datatable/_build_info.py` can be generated manually with `make geninfo`, but is also generated automatically when building wheels, sdist, and even the `_datatable` binaries;

- Added file `VERSION.txt` which contains the simplified datatable version: the release segment + possibly suffix "a" for non-release branches;

- **In order to build the release version of datatable, env. variable `DT_RELEASE` must be set**;

- If env.variable `DT_BUILD_SUFFIX` is set, then it will be used as a local version suffix, producing version strings of the form `{releaseN}a0+{DT_BUILD_SUFFIX}`. The build suffix may contain only alphanumeric characters and dots, but can neither start nor end with a dot;

- In order to build a master/dev version of datatable, set env. variable `DT_BUILD_NUMBER`;

- In all other cases (except when building from sdist), the version of datatable will contain a local suffix of the form `{releaseN}a0+{buildmode}.{timestamp}.{username}`;

- Variables `CI_VERSION_SUFFIX`, `DTBL_GIT_HASH` are no longer used.

- Building wheels on CentOS machine now uses the standard docker container from https://github.com/pypa/manylinux

- The Jenkins pipeline is significantly refactored, including:
    - linux / ppc builds are now done in pypa-manylinux docker images;
    - the wheels produced are now "manylinux" type;
    - testing process simplified: functionality that was previously in Makefile moved to the Jenkins file;
    - the no-pandas, no-numpy steps removed (unnecessary);
    - some duplicate steps removed;
    - coverage steps disabled (not working properly yet);
    - remove dead code in jenkins file and Makefile.

WIP for #2258 
Closes #2265


ci/Jenkinsfile.groovy
```diff
@@ -814,8 +815,6 @@ def test_in_docker(String testtag, String pyver, String docker_image, boolean la
     try {
         def docker_args = ""
         docker_args += "--rm --init "
-        docker_args += "-u `id -u`:`id -g` "
```
mmalohlava: hmm, why it was removed - i cannot guarantee that runnign docker under root will be supported in future.

st-pasha: Getting error `PermissionError: [Errno 13] Permission denied: '/.local'` (http://mr-0xc1:8080/blue/organizations/jenkins/h2o-py-data-table/detail/PR-2287/23/pipeline/234)

I'm not saying that we should run under root, but whatever user is the default for docker, he must have sufficient permissions to do basic operations in the container.

st-pasha: I've put the `-u` flag back, trying a different approach

PR #2289, opened 2020-01-09
Improve progress reporting for FTRL

oleksiyskononenko:
- report loss and epoch while training the FTRL model;
- prevent `iteration_nrows` to become zero.


oleksiyskononenko:
Merging, will take care of the aggregator segfault in a separate PR.

PR #2290, opened 2020-01-10
New Sphinx directive: xfunction

st-pasha:
In this PR we introduce a new approach for auto-generating the documentation from sources.

## Before

Previously we relied on the standard `autodoc` Sphinx extension, which was easy to work with, but the results it produced were in many regards unsatisfactory:
- `autodoc` relies on importing the python module and locating the docs by inspecting runtime objects. This doesn't work well for us, because it would require either building datatable from source on RTD website, or somehow providing pre-built binaries of datatable from S3. Both of these options are non-trivial to implement, and even harder to maintain overtime.
- `autodoc` isn't doing a very good job at rendering the documentation even if it is able to import datatable: function signatures are missing, styling is inadequate, entities are not cross-referenced, overall appearance is bleak, occasionally the markup is broken completely.

<details>
<summary>Screenshot of `autodoc` content rendering</summary>

</details>

## After

The new `xfunction` directive is almost as easy to use as `autodoc`:
```
.. xmethod:: datatable.Frame.to_csv
    :src: c/frame/to_csv.cc Frame::to_csv
```
In contrast with `autodoc`, however, the following features are supported:
- The docs are retrieved directly from C++ source files, bypassing the need to build datatable. This means the docs can now always be up-to-date with the latest source changes;
- Each function/method has a link to the function's source on GitHub, for example: [to_csv() source](https://github.com/h2oai/datatable/blob/38f0b6239641ac94769c221abfd60a67238e5517/c/frame/to_csv.cc#L128-L227);
- Function signatures are extracted and rendered correctly;
- Overall appearance is improved: the important elements such as function's name and parameters are all properly emphasized, all code fragments are rendered in monospace font;
- Whenever a parameter is mentioned in the body of the docstring, it is automatically converted into a link to that parameter's description.
- More flexible: if we find later that some content is not being rendered quite properly, it is now easy to add more transforms into the `xfunction` extension.

<details>
<summary>Screenshot of `xfunction` content rendering</summary>
Screen Shot 2020-01-09 at 3 28 35 PM
</details>

## TODO
 After this PR, the next steps are:
- Convert all Frame methods into `xfunction` format;
- `frame.rst` should be the "birds-eye" overview of all Frame's methods;
- Add tests to ensure that all methods are actually documented -- this way as we add new methods in the future we won't forget to document them;
- Add pretty transforms for the "Examples" section.


angela0xdata:
@st-pasha wow, i really like the way the docs look with xfunction!

PR #2292, opened 2020-01-10
Fix a leak in aggregate_exemplars()

st-pasha:
`DataTable` object stored in `dt_counts` unique_ptr was never deleted.


oleksiyskononenko:
Nice catch, looks like an old bug from the times when we did aggregation in-place. Not sure this could be a reason for segfaults, though...

PR #2293, opened 2020-01-11
[BUILD] Minor refactoring of makefile.

mmalohlava:
Include:
  - Remove repetitive parts of makefile
  - include ulimit option for docker run (-c to enable core dump
  collection)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,20 @@ PLATFORM := $(ARCH)-$(OS)
 # Distribution directory
 DIST_DIR := dist/$(PLATFORM)
 
+# Docker settings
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -208,15 +222,8 @@ centos7_docker_publish: Dockerfile-centos7.$(PLATFORM).tag
```
achraf-mer: `DT_LARGE_TESTS_ROOT` env will be passed for centos7_in_docker, centos7_build_in_docker_impl, that's the only difference I can see.

st-pasha:
After #2258 all docker commands now run from the Jenkins file; this PR is now obsolete.

PR #2309, opened 2020-01-18
[WIP] Finishing AppVeyor Integration

### LINKED ISSUE #2308: Add AppVeyor Windows build to PR checks
Since now datatable compiles on Windows, each new PR should be checked on AppVeyor for not breaking the MSVC compatibility. This can be done with building datatable on AppVeyor and a subsequent test to import datatable module in Python. Once #2301 is done, we should also run a full cycle of tests on AppVeyor and produce a wheel, providing full support of datatable on Windows (#1114).

oleksiyskononenko:
Instead of setting up the build script on AppVeyor, add all the instructions needed to build datatable to  `appveyor.yml` file in the root folder.

Closes #2308


appveyor.yml
```diff
@@ -0,0 +1,25 @@
+version: 0.10.{build}
```
st-pasha: what is this `version`?

appveyor.yml
```diff
@@ -0,0 +1,25 @@
...
+    - master
+
+pull_requests:
+  do_not_increment_build_number: true
+
+build_script:
+- cmd: >-
+    SET PATH=C:\Python37-x64;C:\Python37-x64\Scripts;%PATH%
+
+    pip install packaging
```
st-pasha: I don't think we need `packaging` here.
We only depend on `blessed` and `typesentry` so far, but even those will be removed sooner or later

oleksiyskononenko: I use `packaging` in `compiler.py` to detect a proper version of WinSDK, but this module is not available on Windows by default. Not sure why, because it is available everywhere else where we test datatable.

appveyor.yml
oleksiyskononenko: It is just some dummy version that AppVeyor needs to identify builds. We may want to sync it with the `dev` versions we have, or may not.

appveyor.yml
oleksiyskononenko: Just checked and It seems that this module is not standard, so may be I will replace it with something from the Python standard library or will just write my own function to check the version validity. It is a pretty straightforward operation.

st-pasha: `packaging.version.parse()` is used to parse PEP440-compliant python package versions. It's not the proper tool to parse WinSDK versions -- whatever standard they use, it is certainly different from python's. Though it might work in some simple cases, it wouldn't be a reliable solution.

oleksiyskononenko: Well, here we don't need a very reliable solution, we just need to distinguish "version folders" and "some folders with generic text names". Normally under the WinSDK folder there is nothing else, but the folders containing SDK files, and their names are something like `10.0.18362.1` — packaging module detects these quite well. However, on AppVeyor I've seen a situation when SDK folder contains some other unrelated dirs with just some text names, that's why I needed the fix.

Anyways, I've already replaced `packaging.version.parse()` with a simple function that will filter out all these unrelated dirs.

PR #2327, opened 2020-01-27
Add Logger2 to xbuild

oleksiyskononenko:
- add a conservative Sublime friendly `Logger2` to xbuild; 
- fix all the compilation warnings produced by `clang` on Mac.

### DIFF
--- a/src/core/column/const.cc
+++ b/src/core/column/const.cc
@@ -159,9 +159,8 @@ class ConstFloat_ColumnImpl : public Const_ColumnImpl {
           return SType::FLOAT64;
 
         default:
-          xassert(0 && "Invalid stype for float constant");  // LCOV_EXCL_LINE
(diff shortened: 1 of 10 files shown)


src/core/column/const.cc
```diff
@@ -160,6 +160,8 @@ class ConstFloat_ColumnImpl : public Const_ColumnImpl {
 
         default:
           xassert(0 && "Invalid stype for float constant");  // LCOV_EXCL_LINE
+          throw RuntimeError() << "Wrong `stype0` in `normalize_stype()`: "
+                  << stype0;
```
st-pasha: if we're throwing an exception, then the previous `xassert` is no longer needed

oleksiyskononenko: Yeah, we can now remove `xassert` here and in the other places too.

PR #2338, opened 2020-01-30
Switch to C++17 language standard

st-pasha:
Today we compile `datatable` only with C++17 -compliant compilers, so there is no reason to keep using the old C++11 language standard.

This switch doesn't require any changes to the existing code, and we do not expect any externally-visible effects.

### DIFF
--- a/ci/ext.py
+++ b/ci/ext.py
@@ -248,7 +248,7 @@ def build_extension(cmd, verbosity=3):
             ext.compiler.add_linker_flag("/DEBUG:FULL")
     else:
         # Common compile flags
-        ext.compiler.add_compiler_flag("-std=c++11")
+        ext.compiler.add_compiler_flag("-std=c++17")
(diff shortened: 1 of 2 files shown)


st-pasha:
@mmalohlava After doing some digging, it appears that the C++17 standard is supported on MacOS only starting from version 10.13 (High Sierra). 
According to https://gs.statcounter.com/macos-version-market-share/desktop/worldwide the marker share of MacOS-10.13+ is 79.21% as of Dec-2019.

oleksiyskononenko:
For our Windows builds we are now using compiler provided by MS Visual Studio 2019, and this compiler only [supports](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019) C++14 and C++17 standards. It would be if we stick to the same C++ standard for all the operating systems.

Issue #2301, opened 2020-01-15
Fix failing functionality and tests on Windows


oleksiyskononenko:
Though datatable compiles with MSVC as of #2291, some tests are failing and memory mapping functionality doesn't seem to work. Here is a brief list of issues:

- [x] `long` and `long long` related issues
- [x] `long double` related issues
- [x] memory mapping issues for `.jay` files
- [x] memory mapping issues in `fread`
- [x] frame view issues
- [x] url related issues in `fread`
- [ ] file descriptor related issues (`test_fread_from_fileobj` disabled for now)
- [x] numpy related issues
- [x] misc 

Prerequisite for #1114

PR #2344, opened 2020-02-07
Fix frame to string conversion on Windows

oleksiyskononenko:
- fix the way terminal size is detected on Windows;
- since internally `dt.options.display.max_column_width` couldn't be less than `2`, to be consistent, also adjust the setter for this option.

WIP for #2301

### DIFF
--- a/src/core/utils/terminal/terminal.cc
+++ b/src/core/utils/terminal/terminal.cc
@@ -52,22 +52,22 @@ Terminal& Terminal::plain_terminal() {
   return term;
 }
 
-Terminal::Terminal(bool is_plain) {
+Terminal::Terminal(bool is_plain) : is_plain_(is_plain){
   // Note: there is no simple way to catch the terminal re-size event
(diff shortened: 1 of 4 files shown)


src/core/utils/terminal/terminal.cc
```diff
@@ -157,7 +157,7 @@ bool Terminal::unicode_allowed() const noexcept {
 
 TerminalSize Terminal::get_size() {
   #if DT_OS_WINDOWS
-    _detect_window_size();
+    if (!is_plain_) _detect_window_size();
   #else
     if (!size_.width || !size_.height) _detect_window_size();
   #endif
```
st-pasha: Can merge the two branches together:
```suggestion
  if (!is_plain_ &&
      (DT_OS_WINDOWS || size_.width == 0)) _detect_window_size();
```

oleksiyskononenko: Sure, this part is not performance critical, so won't loose much.

PR #2345, opened 2020-02-12
Do not use Popen.wait() in fread

oleksiyskononenko:
- `Popen.wait()` is [known](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait) to cause deadlocks when used in conjunction with pipes. Due to this fact some tests, like `test_fread_from_cmd2`, were constantly hanging on Windows. In this PR we adjust `fread()` not to call `Popen.wait()`.

- minor fix in the `terminal.py` to prevent it from hanging when used in the multiprocessing pool. Python terminal class is going to be fully replaced by the corresponding C++ code anyways.

WIP for #2301

### DIFF
--- a/src/datatable/fread.py
+++ b/src/datatable/fread.py
@@ -280,12 +280,10 @@ class GenericReader(object):
         proc = subprocess.Popen(cmd, shell=True,
                                 stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE)
-        ret = proc.wait()
         msgout, msgerr = proc.communicate()
-        if ret:
+        if msgerr:
             msgerr = msgerr.decode("utf-8", errors="replace").strip()
-            raise TValueError("Shell command returned error code %r: `%s`"
-                              % (ret, msgerr))
+            raise TValueError("Shell command returned an error: `%s`" % msgerr)
         else:
             self._text = msgout
             self._src = cmd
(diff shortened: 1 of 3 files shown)


src/datatable/fread.py
```diff
@@ -280,12 +280,10 @@ def _resolve_source_cmd(self, cmd):
         proc = subprocess.Popen(cmd, shell=True,
                                 stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE)
-        ret = proc.wait()
```
st-pasha: nice catch

src/datatable/fread.py
```diff
@@ -280,12 +280,10 @@ def _resolve_source_cmd(self, cmd):
...
         msgout, msgerr = proc.communicate()
-        if ret:
+        if msgerr:
```
st-pasha: It is conceivable to have a process that terminates with an error, but doesn't write anything to the stderr. Thus, checking the return code is more robust. However, the return code can be queried after `.communicate()` using `.poll()` or `.returncode`.

oleksiyskononenko: Yeah, let me also add a check for the return code.

st-pasha: Try `dt.fread(cmd="false")`

st-pasha: perhaps we can even add this as a test?

oleksiyskononenko: Sure, but better use `cmd="exit 1"` or something like that, because on Windows `cmd="false"` will pass with `Command not found` error message.

PR #2346, opened 2020-02-13
Read Excel date values

### LINKED ISSUE #1701: fread() reads dates from xlsx as float64
`dt.fread("~/Downloads/dates.xlsx")
`returns 

```
Week_start | Value_Sales
0 | 42008 | 3527.56
1 | 42015 | 4344.03
2 | 42022 | 4103.33
3 | 42029 | 21530.4
4 | 42036 | 4102.35
5 | 42043 | 3965.92
```

while pandas 

`pd.read_excel("../Downloads/dates.xlsx")
`
returns

```
Week_start	Value_Sales
0	2015-01-04 01:00:00	3527.56
1	2015-01-11 01:00:00	4344.03
2	2015-01-18 01:00:00	4103.33
3	2015-01-25 01:00:00	21530.41
4	2015-02-01 01:00:00	4102.35
5	2015-02-08 01:00:00	3965.92


```

[dates.xlsx](https://github.com/h2oai/datatable/files/2923197/dates.xlsx)

st-pasha:
- When creating a frame from list of mixed types containing bools/ints/floats and also strings, the resulting column will now have string type instead of object type. This is consistent with how fread reads CSV files, and also with the semantics of rowbinding.
- When reading an Excel file we now parse datetime and boolean values correctly.

Closes #1701

### DIFF
--- a/tests/fread/test_xls.py
+++ b/tests/fread/test_xls.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #-------------------------------------------------------------------------------
-# Copyright 2018 H2O.ai
+# Copyright 2018-2020 H2O.ai
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 #-------------------------------------------------------------------------------
 import pytest
 import random
+import datatable as dt
+from tests import find_file, assert_equals
 from datatable.xls import (
     _parse_row, _combine_ranges, _process_merged_cells,
     _range2d_to_excel_coords, _excel_coords_to_range2d)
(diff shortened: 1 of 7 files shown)


tests/fread/test_xls.py
```diff
@@ -88,3 +90,113 @@ def test__excel_coords_to_range2d(seed):
...
+    DT = dt.fread(filename)
+    assert DT.shape == (7, 1)
+    assert DT.names == ("Data",)
+    assert DT.to_list()[0] == ['2019-05-06 00:00:00',
+                               '2019-05-07 00:00:00',
+                               '2019-05-08 00:00:00',
+                               '2019-05-09 00:00:00',
+                               '2019-05-10 00:00:00',
+                               '2019-05-11 00:00:00',
+                               '2019-05-12 00:00:00']
```
oleksiyskononenko: What about also adding `assert DT.stype == dt.str32` to check the stype explicitly?

st-pasha: Eventually we will be creating a `date64` column here, so the test will have to be updated anyways.

PR #2349, opened 2020-02-14
Use `with` when working with the multiprocessing pool

oleksiyskononenko:
Using `with` statement, when working with the multiprocessing pool in `test_parallel.py::test_multiprocessing_threadpool`, seems to fix hangs in Python 3.8, as it properly closes the pool at exit.

### DIFF
--- a/tests/test_parallel.py
+++ b/tests/test_parallel.py
@@ -42,19 +42,18 @@ cpp_test = pytest.mark.skipif(not hasattr(core, "test_coverage"),
 #-------------------------------------------------------------------------------
 
 def test_multiprocessing_threadpool():
-    import atexit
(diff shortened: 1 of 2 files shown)


tests/test_parallel.py
```diff
@@ -42,12 +42,14 @@
 #-------------------------------------------------------------------------------
 
 def test_multiprocessing_threadpool():
+    import atexit
     # Verify that threads work properly after forking (#1758)
     import multiprocessing as mp
     from datatable.internal import get_thread_ids
     parent_threads = get_thread_ids()
     n = 4
     pool = mp.Pool(processes=n)
```
st-pasha: better use `with mp.Pool(...) as pool:` here

PR #2350, opened 2020-02-15
Fix sort-related issues

### LINKED ISSUE #2348: AssertionError when grouping & sorting
```
from datatable import dt, f, by, sort
DT = dt.Frame(A=[1, 2, 3, 1, 2, 3], B=list('akdfnv'), C=[0.5]*6)
DT[:, :, by(f.A, f.B), sort(f.A, f.B)]
```
throws
```
AssertionError: Assertion 'is_writable()' failed in src/core/buffer.cc, line 861
```

Attribute-to: @arnocandel

st-pasha:
- Column materialization is adjusted so that a non-virtual column which has refcount>1 wouldn't get copied needlessly. This should improve the performance of sorting: currently we materialize all columns before using them for sorting/grouping, and during this process we currently accidentally create a copy of the data.

- Fixed the use of `DTDEBUG` macro: this macro is always defined, and equal either 0 or 1. However, using it as `#ifdef DTDEBUG` is incorrect, it should be `#if DTDEBUG` instead. Because of this misunderstanding, we were including some code that was meant for debug builds only into production, specifically code for tracking/untracking all memory allocation. After this fix we expect the performance to improve, especially the sort function which allocates many memory ranges internally.

- There was an issue with combining a `by()` operator with a `sort()` when the sort function involves more than one column. The issue was manifesting as an assertion error after we switched from an array to buffer implementation in one of the recent PRs (however it seems the problem existed before too, only it was masked). With certain configuration of the data it was even possible to cause a segfault. This has now been fixed. 

Closes #2348

### DIFF
--- a/tests/test_dt_sort.py
+++ b/tests/test_dt_sort.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #-------------------------------------------------------------------------------
-# Copyright 2018-2019 H2O.ai
+# Copyright 2018-2020 H2O.ai
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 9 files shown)


tests/test_dt_sort.py
```diff
@@ -931,3 +931,16 @@ def test_h2oai7014(tempfile_jay):
...
+    DT = dt.Frame(A=[1, 2, 3, 1, 2, 3], B=list('akdfnv'),
+                  C=[0.1, 0.2, 0.3, 0.4, 0.5, 0.6],
+                  D=[11]*6, E=[2]*6)
+    # Check that these expressions do not crash
+    DT[:, :, by(f.A), sort(f.A, f.E)]
+    # DT[:, :, by(f.A, f.B), sort(f.A, f.B)]
+    # assert_equals(DT[:, dt.count(), by(f.D), sort(f.E, f.A)],
+    #               dt.Frame([[11], [6]],
+    #                        names=["D", "count"],
+    #                        stypes=[dt.int32, dt.int64]))
```
oleksiyskononenko: Are these lines commented out intentionally? Seems like we're not asserting anything in this test.

st-pasha: nope, I uncommented them locally but then forgot to commit

PR #2359, opened 2020-02-21
Docs2

### LINKED ISSUE #2357: Deleting a key in a frame results in a segfault
```python
>>> import datatable as dt
>>> DT = dt.Frame(A=[3, 4, 1], B=["one", "two", "three"])
>>> DT.key = "A"
>>> del DT.key
Segmentation fault: 11
```

### LINKED ISSUE #2367: Values in a string column may become NA after rbinding
```
import datatable as dt
DT0 = dt.Frame(["2017-11-10 07:50:18" for i in range(14887665)])
DT1 = dt.rbind(*[DT0 for i in range(8)])
assert DT1[-1, "C0"] is not None # Fails
```

```
>>> print(DT1[113025454, 0])
2017-11-10 07:50:18
>>> print(DT1[113025455, 0])
None
```
the length of each string is 19 chars, and 113025454*19 == 2147483626 == 0x7fffffea. The string column overflows the int32 limit

Attribute-To: @arnocandel

st-pasha:
- Fixed crash from `del DT.key`;
- Added docs for `.head()`, `.tail()` and `.key`;
- Implemented keyed frames for `..dtframe` Sphinx directive;
- Added ability to supply source-link for property setters;
- Added ability to detect types line when parsing a frame from docs;

Closes #2357

### DIFF
--- a/src/core/frame/key.cc
+++ b/src/core/frame/key.cc
@@ -35,12 +35,12 @@ static const char* doc_key =
 R"(
 The tuple of column names that are the primary key for this frame.
 
+If the frame has no primary key, this property returns an empty tuple.
+
 The primary key columns are always located at the beginning of the
-frame, and therefore the following identity always holds::
+frame, and therefore the following holds::
 
     DT.key == DT.names[:len(DT.key)]
-
-If the frame is not keyed, this property returns an empty tuple.
 
 Assigning to this property will make the Frame keyed by the specified
 column(s). The key columns will be moved to the front, and the Frame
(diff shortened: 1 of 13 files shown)


src/core/frame/key.cc
```diff
@@ -31,16 +31,63 @@
...
+str32 | int32
+----- + -----
+five  |     4
+four  |     3
+one   |     0
+three |     2
+two   |     1
+--
+[5 rows x 2 columns]
+)";
```
oleksiyskononenko: Are you sure you've committed the code to render the column types?

st-pasha: The column types are not rendered in the console (yet). In the example I've added those manually. However, the doc-parser is now able to recognize these types (if present) and convert them into a proper `..dtframe` directive that will be rendered into HTML with the correct type strip.

PR #2381, opened 2020-03-10
Implement the BOM mark for .to_csv()

### LINKED ISSUE #2379: Add option to include BOM mark in to_csv()
This should be an optional flag, e.g.
```
DT.to_csv("out.csv", bom=True)
```
with default value of `False`.

The BOM mark allows some programs (notably, Excel) to recognize files as UTF-8 encoded (which they should assume by default, but for historical reasons they don't).

st-pasha:
Added argument `bom=False` to the `Frame.to_csv()` method. This argument controls whether or not to write BOM (Byte-Order Mark) into the output csv file.

Closes #2379

### DIFF
--- a/src/core/frame/to_csv.cc
+++ b/src/core/frame/to_csv.cc
@@ -42,8 +42,8 @@ static void change_to_lowercase(std::string& str) {
 
 static const char* doc_to_csv =
 R"(to_csv(self, path=None, *, quoting="minimal", append=False,
-       header=..., hex=False, compression=None, verbose=False,
-       _strategy="auto")
+       header=..., bom=False, hex=False, compression=None,
+       verbose=False, _strategy="auto")
 --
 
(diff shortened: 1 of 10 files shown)


src/core/frame/to_csv.cc
```diff
@@ -87,6 +87,17 @@ header: bool | ...
     will be written in all cases except when appending content into
     an existing file.
 
+bom: bool
+    If True, then insert the byte-order mark into the output file
```
oleksiyskononenko: May be add that this is `the UTF-8 byte-order mark`?

st-pasha: BOM is the unicode character U+FEFF, and this is the character that we're adding. The output file is written in UTF-8 however, so the BOM character becomes a 3-byte sequence `EF BB BF`. But this is just the consequence of the fact that we're writing a UTF-8 encoded output.

PR #2388, opened 2020-03-13
WIP on new-sort

st-pasha:
This adds ability to sort by multiple columns, the most important sorter so far, however it still seems to fail in some cases.

The new sorter is still work in progress.


samukweku:
hi @st-pasha, working on a doc/mini-tutorial for ```pydatatables```, and I wish to know how to sort on multiple columns and varying the order of the sorting - say I want the first column to be in ascending order, and the next column in descending order. How do I achieve that? I look at the docs for sort,  it  has a ```reverse``` parameter, but I am not sure how that works - if it accepts a list of booleans?

PR #2390, opened 2020-03-21
Add explicit end-of-file checks to fread

oleksiyskononenko:
On Windows there is no simple way to make the "overmapping" functionality and the corresponding `Overmap_BufferImpl` class to work. This functionality was needed to ensure that the last byte when memory mapping a particular file was always `\0`, and in such a way to avoid explicit checks for the end-of-file in many places.

In this PR we
- add explicit end-of-file checks in fread and get rid of the `Overmap_BufferImpl` class
- improve Jenkins scripts, especially the part related to the larger fread tests
- fix minor issues with the tests

Benchmarking this PR against master on various datasets we found that there is almost no difference in terms of performance, for example on 2.5 GHz Quad-Core Intel Core i7

| Dataset                            | Size [MB]       | this PR [s] | master [s] |
| :---------------------- | ------------- |------------|-----------|
| credit_card.csv               | 3                      | 0.006277   | 0.006114 |
| allyears.1987.2013.csv  | 15000             | 109.4026   | 109.0010 |

However, in contrast to master this PR has no issues related to fread on Windows.

WIP for #2301


ci/Jenkinsfile.groovy
```diff
@@ -62,8 +62,11 @@ DOCKER_IMAGE_X86_64_MANYLINUX = "quay.io/pypa/manylinux2010_x86_64"
 // Note: global variables must be declared without `def`
 //       see https://stackoverflow.com/questions/6305910
 
-// Needs invocation of larger tests
-needsLargerTest = false
+// Larger tests are invoked if forced or if isModified() routine
+// returns `true`, and unless the larger tests are explicitely
+// disabled. Here we only define the default value for `needsLargerTests`
+// adjusting it when we check for the actual code changes.
+needsLargerTests = params.FORCE_LARGER_TESTS || params.FORCE_ALL_TESTS
```
st-pasha: Can we please call this variable `doFreadTests`, similar to all other variables in lines 71-75.
Likewise, the param name can be `FORCE_FREAD_TESTS`. This would reduce the confusion, I think.

oleksiyskononenko: Yeah, I already renamed it. Just didn't push as waiting for the larger tests to finish.

ci/Jenkinsfile.groovy
```diff
@@ -309,7 +313,7 @@ ansiColor('xterm') {
                                     unstash 'x86_64-manylinux-wheels'
                                     test_in_docker("x86_64-manylinux-py37", "37",
                                                    DOCKER_IMAGE_X86_64_MANYLINUX,
-                                                   needsLargerTest)
+                                                   needsLargerTests)
```
st-pasha: I somehow think we don't need to pass this variable as a parameter -- it's global after all. Also, we don't pass any other variables like `doPpcTests` or `doExtraTests`.

src/core/buffer.cc
```diff
@@ -176,6 +178,20 @@ class BufferImpl
...
+      return !(size_%page_size_);
+    }
+
+    static size_t calc_page_size() {
+      #if DT_OS_WINDOWS
+        SYSTEM_INFO sysInfo;
+        GetSystemInfo(&sysInfo);
+        const size_t page_size = sysInfo.dwPageSize;
+      #else
+        const size_t page_size = static_cast<size_t>(sysconf(_SC_PAGE_SIZE));
```
st-pasha: no need for a local variable - can return the result immediately

ci/Jenkinsfile.groovy
oleksiyskononenko: As for the `FORCE_FREAD_TESTS`, I'm not sure. Can we have larger tests for other datatable functions? For example larger sort, join or groupby tests that may be optional and controlled with the same flag? 

In any case I guess it's better to keep the word `LARGER` in this variable, because it doesn't control the other fread tests. Something like `FORCE_LARGER_FREAD_TESTS`.

ci/Jenkinsfile.groovy
oleksiyskononenko: Probably yes. We can also do a lot of other re-factoring and fixes in this groovy file.

src/core/buffer.cc
oleksiyskononenko: That's true. But I normally prefer to have just one `return` statement, in this way it's easier to control what and where we actually return.

ci/Jenkinsfile.groovy
st-pasha: I mean, "large tests" doesn't really reflect the purpose of the variable. It really only controls extra fread tests. So we could call it `doLargeFreadTests` or `doExtraFreadTests` or `doFreadTests`, whereas "do large tests" could be used to execute tests with >2B rows frames or >2GB string columns or similar.

oleksiyskononenko: Right now — yes. I thought there was an idea for that variable to control some other larger tests? Or it got its name just by chance?

st-pasha: If we need to control the execution of really large tests, that should probably be a separate variable.

ci/Jenkinsfile.groovy
st-pasha: feel free

src/core/buffer.cc
st-pasha: Yeah, but given that this is a `#if .. #else` case, the compiler will really see only 1 return anyways.

ci/Jenkinsfile.groovy
oleksiyskononenko: Make sense. I will rename it once the larger tests finish.

ci/Jenkinsfile.groovy
oleksiyskononenko: I've already fixed the `isModified()` routine, it actually never returned `true` as only checked the base file names and not the full file names to match the pattern. That is probably a reason why the larger tests were never triggered automatically and some broken tests leaked.

src/core/buffer.cc
oleksiyskononenko: For compiler, yes. I'm talking more about myself looking at the code :) It seems easier to catch just one `return` line than going through all the macro branches.

ci/Jenkinsfile.groovy
oleksiyskononenko: Btw, I guess the variable name is coming from the datatable environment variable `DT_LARGE_TESTS_ROOT`. This one has no indication it is only applicable to fread tests.

src/core/csv/reader_parsers.cc
```diff
@@ -428,35 +434,35 @@ void parse_float64_extended(FreadTokenizer& ctx) {
...
-  if (ch[0]=='n' && ch[1]=='a' && ch[2]=='n' && (ch += 3)) goto return_nan;
-  if (ch[0]=='i' && ch[1]=='n' && ch[2]=='f' && (ch += 3)) goto return_inf;
-  if (ch[0]=='I' && ch[1]=='N' && ch[2]=='F' && (ch += 3)) goto return_inf;
-  if (ch[0]=='I' && ch[1]=='n' && ch[2]=='f' && (ch += 3)) {
-    if (ch[0]=='i' && ch[1]=='n' && ch[2]=='i' && ch[3]=='t' && ch[4]=='y') ch += 5;
+  if (ch + 2 < ctx.eof && ch[0]=='n' && ch[1]=='a' && ch[2]=='n' && (ch += 3)) goto return_nan;
+  if (ch + 2 < ctx.eof && ch[0]=='i' && ch[1]=='n' && ch[2]=='f' && (ch += 3)) goto return_inf;
+  if (ch + 2 < ctx.eof && ch[0]=='I' && ch[1]=='N' && ch[2]=='F' && (ch += 3)) goto return_inf;
+  if (ch + 2 < ctx.eof && ch[0]=='I' && ch[1]=='n' && ch[2]=='f' && (ch += 3)) {
+    if (ch + 3 < ctx.eof && ch[0]=='i' && ch[1]=='n' && ch[2]=='i' && ch[3]=='t' && ch[4]=='y') ch += 5;
```
st-pasha: `ch + 4`

src/core/csv/reader_parsers.cc
```diff
@@ -509,25 +515,25 @@ void parse_float64_hex(FreadTokenizer& ctx) {
   bool Eneg, subnormal = 0;
   ch += (neg = (*ch=='-')) + (*ch=='+');
```
st-pasha: eof check here?

src/core/csv/reader_parsers.cc
```diff
@@ -428,35 +434,35 @@ void parse_float64_extended(FreadTokenizer& ctx) {
...
+    if (ch + 6 < ctx.eof && ch[1]=='D' && ch[2]=='I' && ch[3]=='V' && ch[4]=='/' && ch[5]=='0' && ch[6]=='!' && (ch += 7)) goto return_nan;
+    if (ch + 6 < ctx.eof && ch[1]=='V' && ch[2]=='A' && ch[3]=='L' && ch[4]=='U' && ch[5]=='E' && ch[6]=='!' && (ch += 7)) goto return_nan;
+    if (ch + 5 < ctx.eof && ch[1]=='N' && ch[2]=='U' && ch[3]=='L' && ch[4]=='L' && ch[5]=='!' && (ch += 6)) goto return_na;
+    if (ch + 5 < ctx.eof && ch[1]=='N' && ch[2]=='A' && ch[3]=='M' && ch[4]=='E' && ch[5]=='?' && (ch += 6)) goto return_na;
+    if (ch + 4 < ctx.eof && ch[1]=='N' && ch[2]=='U' && ch[3]=='M' && ch[4]=='!' && (ch += 5)) goto return_na;
+    if (ch + 4 < ctx.eof && ch[1]=='R' && ch[2]=='E' && ch[3]=='F' && ch[4]=='!' && (ch += 5)) goto return_na;
+    if (ch + 3 < ctx.eof && ch[1]=='N' && ch[2]=='/' && ch[3]=='A' && (ch += 4)) goto return_na;
   }
   parse_float64_simple(ctx);
   return;
```
st-pasha: Missing eof check on line 477 below `if (quoted && *ch!=ctx.quote)`

src/core/csv/reader_parsers.cc
```diff
@@ -47,7 +47,7 @@ void parse_bool8_numeric(FreadTokenizer& ctx) {
   // *ch=='1' => d=1,
   // *ch==(everything else) => d>1
   uint8_t d = static_cast<uint8_t>(*ch - '0');
```
st-pasha: missing eof check

src/core/csv/reader_parsers.cc
```diff
@@ -122,13 +122,13 @@ void parse_int_simple(FreadTokenizer& ctx) {
   int sd = 0;                 // number of significant digits (without initial 0s)
```
st-pasha: eof check on line 118

src/core/csv/reader_parsers.cc
```diff
@@ -177,8 +177,8 @@ void parse_intNN_grouped(FreadTokenizer& ctx) {
   int sf = 0;              // number of significant digits (without initial 0s)
```
st-pasha: eof check on line 170

oleksiyskononenko:
Thanks, let me go through this one more time. Seems like it is easy to miss.

src/core/csv/reader_parsers.cc
```diff
@@ -573,21 +605,24 @@ void parse_string(FreadTokenizer& ctx) {
 
   // need to skip_whitespace first for the reason that a quoted field might have space before the
   // quote; e.g. test 1609. We need to skip the space(s) to then switch on quote or not.
-  if (*ch==' ' && ctx.strip_whitespace) while(*++ch==' ');  // if sep==' ' the space would have been skipped already and we wouldn't be on space now.
+  if (ch < ctx.eof && *ch==' ' && ctx.strip_whitespace) {
+    while(*++ch==' ' && ch < ctx.eof);  // if sep==' ' the space would have been skipped already and we wouldn't be on space now.
```
st-pasha, line 609: this doesn't look right: `*++ch==' '` is equivalent to `{ch++; *ch==' '}`. If you compare with eof afterwards, then you can still have an invalid past-the-end-of-file read. Try rewriting it as
```suggestion
if (ctx.strip_whitespace) { 
  while (ch < ctx.eof && *ch == ' ') ch++;
}
```

src/core/csv/reader_parsers.cc
```diff
@@ -652,8 +687,8 @@ void parse_string(FreadTokenizer& ctx) {
           // if there is a ", afterwards but before the next \n, use that; the field was quoted and it's still case (i) above.
           // Otherwise break here at this first sep as it's case (ii) above (the data contains a quote at the start and no sep)
           ch2 = ch;
-          while (*++ch2 && *ch2!='\n' && *ch2!='\r') {
-            if (*ch2==quote && (ch2[1]==sep || ch2[1]=='\r' || ch2[1]=='\n')) {
+          while (ch < ctx.eof && *++ch2 && *ch2!='\n' && *ch2!='\r') {
+            if (ch + 1 < ctx.eof && *ch2==quote && (ch2[1]==sep || ch2[1]=='\r' || ch2[1]=='\n')) {
```
st-pasha, line 691: - the eof check should use `ch2` variable instead of `ch`;
- replace pre-increment `++ch2` with an increment within the body of the while loop;
- i believe the `*++ch2` condition (which means `*++ch2!='\0') was an old-style eof check, so should be removed

src/core/csv/reader_parsers.cc
oleksiyskononenko: Yeah, you're right. This was an old change and I didn't think it through.

src/core/csv/reader_parsers.cc
oleksiyskononenko: Yes, this is a typo, here `ch2` should be checked, of course.

oleksiyskononenko:
Yeah, I have already addressed those just didn’t push it yet. Tried to force the ppc build and tests, but they always fail for some reasons.

src/core/csv/reader_parsers.cc
oleksiyskononenko: As for the increment, I guess the same is true for line 679, but to replace it with an increment one should also take care about the initial value.

oleksiyskononenko:
All tests are green except for several ppc machines that seem to have problems when running larger fread tests. Doesn't seem to be related to the changes on this PR. Mergin.

Issue #1950, opened 2019-08-05
xread: an "extract+read"


st-pasha:
This is a proposal for implementing a new function `xread()`, which would be conceptually similar to `fread()`, but much lazier. In particular, `xread()` would parse only the first `n_sample_lines=100` lines of the file, detecting the general information such as parse settings, the number columns, their names and types. After that, `xread()` returns a "lazy frame" object, which can be used with the standard `[i,j]` notation:

- `[:n, :]` returns just the first `n` rows of the dataset (equivalent to `max_nrows` parameter);
- `[1000:2000, :]` returns rows from 1000 to 2000. Generally, we should allow the user to request consecutive ranges on the same lazy frame. This will be equivalent to "chunked reading", which is a popular request;
- `[-100:, :]` returns the last 100 rows. For a "file object" sources, this would require that the file is read in its entirety. For files on disk (or in memory), we could try to parse from the end of the file.
- `[:, :5]` return only the first 5 columns.
- `[:, ["A", "B", "C"]]` return columns named `A`, `B` and `C`.
- `[:, f + {"wage_per_hour": f.salary/f.hours}]` return all columns + an additional column containing the `salary` divided by `hours`.
- `[f.status != '', :]` return only the rows where field `status` is not empty.
- `[dt.random.randu() < 0.2, :]` randomly sample 20% of the rows.
- ...

These are just some of the examples of what could be possible. Obviously, the `i` and `j` selectors can be combined into a single `[i, j]` selector too. It should be even possible to add join operations and groupbys into the mix (provided that we use single-pass hash-based grouping).

st-pasha:
# Implementation note

Whereas the standard `DT[i,j,by]` selector takes advantage of the fact that the data is readily present in memory and allows random access and multiple passes, this is not so with the "lazy frame" as produced by `xread`. The key property of this lazy frame is that it is essentially single-pass (with some buffering), and therefore the algorithm for applying the `[i,j,by]` selector must be rethought accordingly.

Another crucial property of the xread-frame is that the principal driver of the materialization process is the ordered read loop. Thus, the process can be visualized as follows:
```
+--------+           +---------+
|  raw   |  acquire  |  main   |   filter    +--------+
|  data  |<----------| ordered |------------>| Result |
| source |   data    |  loop   |  transform  +--------+
+--------+           +---------+
```
[...]

PR #2396, opened 2020-04-03
iread() function

### LINKED ISSUE #2008: Handle zip-files with multiple files.

### LINKED ISSUE #2392: fread should read .tar.gz files properly
We already have support for .gz files, but not for .tar.gz.

st-pasha:
- Added new function `iread()`, which is similar to `fread()`, but is suitable for reading multiple files at once. The function returns an iterator of Frames.
- This function can be used to read multi-file archives, glob patterns, multi-sheet Excel files, etc. In particular, `.zip` and `.tar.gz` files are supported.
- `fread()` now always returns a single Frame. If multiple sources are provided, then either the first will be used, or an error thrown, depending on the parameter `multiple_sources=`.
- fread now throws `IOError`s instead of `RuntimeError`s.

- Fixed a bug with the warnings handler (the warnings were not properly displayed);
- Simplified internal API of `progress_bar` class: now the progress bar is always destroyed when the last task is popped off the stack. In order to display "cancelled"/"error" status on the progress bar we now set that indicator when the SIGINT is caught.

Closes #2392
Closes #2008


src/core/progress/progress_manager.cc
```diff
@@ -62,8 +62,9 @@ void progress_manager::finish_work(work* task, bool successfully) {
   tasks.pop();
 
   std::lock_guard<std::mutex> lock(mutex);
-  if (successfully && tasks.empty()) {
-    pbar->set_status_finished();
+  if (tasks.empty()) {
+    if (successfully) pbar->set_status_finished();
+    else              pbar->set_status_error(false);
```
oleksiyskononenko, line 67: I see you did some changes to the progress bar code. What was the reason and do you have any clue as to how [this](http://mr-0xc1:8080/blue/rest/organizations/jenkins/pipelines/h2o-py-data-table/branches/PR-2390/runs/26/nodes/165/log/?start=0) could happen?

```
[2020-04-01T00:20:33.437Z]             assert 0 <= p.progress <= 1
[2020-04-01T00:20:33.437Z]             assert p.status in ("running", "finished", "cancelled", "error")
[2020-04-01T00:20:33.437Z]             progress_reports.append(p)
[2020-04-01T00:20:33.437Z]     
[2020-04-01T00:20:33.437Z]         with dt.options.progress.context(callback = progress_fn,
[2020-04-01T00:20:33.437Z]                                          enabled=True,
[2020-04-01T00:20:33.437Z]                                          min_duration=0):
[2020-04-01T00:20:33.437Z]             [d_exemplars, d_members] = aggregate(d_in, min_rows=0,
[2020-04-01T00:20:33.437Z]                                                  nd_max_bins=div, seed=1)
[2020-04-01T00:20:33.437Z]             messages = ("", "Preparing", "Aggregating", "Sampling", "Finalizing")
[2020-04-01T00:20:33.437Z]             message_index = 0
[2020-04-01T00:20:33.437Z]             for i, p in enumerate(progress_reports):
[2020-04-01T00:20:33.437Z]                 if i > 0:
[2020-04-01T00:20:33.437Z]                     assert p.progress >= progress_reports[i-1].progress
[2020-04-01T00:20:33.437Z]                 message_index_new = messages.index(p.message)
[2020-04-01T00:20:33.438Z] >               assert message_index <= message_index_new or p.status == "finished"
[2020-04-01T00:20:33.438Z] E               AssertionError: assert (1 <= 0 or 'running' == 'finished'
[2020-04-01T00:20:33.438Z] E                 - running
[2020-04-01T00:20:33.438Z] E                 + finished)
```

It looks like the progress bar was still running, but the message it sent was empty. It may not be a big deal since the the same message is normally sent multiple times.

src/core/read/py_fread.cc
```diff
@@ -90,11 +92,85 @@ static py::oobj fread(const py::PKArgs& args) {
...
+}
+
+
+
+//------------------------------------------------------------------------------
+// iread() python function
+//------------------------------------------------------------------------------
+
+static const char* doc_iread =
+R"(iread(anysource=None, *, file=None, text=None, cmd=None, url=None,
```
oleksiyskononenko, line 110: btw, what `i` in `iread` means? and what is the main reason to separate `iread` functionality from `fread`?

src/datatable/utils/fread.py
```diff
@@ -238,21 +248,47 @@ def _resolve_archive(filename, subpath, tempfiles):
...
+        else:
+            return (None, None, None, None), extracted_files
+
+    elif filename.endswith(".tar.gz"):
+        import tarfile
+        zf = tarfile.open(filename, mode="r:gz")
+        zff = [entry.name for entry in zf.getmembers() if entry.isfile()]
+        if subpath:
+            if subpath in zff:
+                filename += "/" + subpath
```
oleksiyskononenko, line 273: `os.path.join()` here too?

src/datatable/utils/fread.py
```diff
@@ -238,21 +248,47 @@ def _resolve_archive(filename, subpath, tempfiles):
...
+                              % (subpath, filename))
+        extracted_files = []
+        for entryname in zff:
+            if logger:
+                logger.debug("Extracting %s/%s to temporary directory %s"
+                             % (filename, entryname, tempfiles.tempdir))
+            newfile = tempfiles.create_temp_file()
+            with zf.extractfile(entryname) as inp, open(newfile, "wb") as out:
+                out.write(inp.read())
+            srcname = filename + "/" + entryname
```
oleksiyskononenko, line 286: `os.path.join()`?

src/datatable/utils/fread.py
```diff
@@ -238,21 +248,47 @@ def _resolve_archive(filename, subpath, tempfiles):
...
-        out_file = newfile
+                raise IOError("File `%s` does not exist in archive `%s`"
+                               % (subpath, filename))
+        extracted_files = []
+        for zf_file in zff:
+            if logger:
+                logger.debug("Extracting %s/%s to temporary directory %s"
+                             % (filename, zf_file, tempfiles.tempdir))
+            newfile = zf.extract(zf_file, path=tempfiles.tempdir)
+            srcname = filename + "/" + zf_file
```
oleksiyskononenko, line 259: What about using `os.path.join()` here? This may be helpful for Windows support.

src/core/progress/progress_manager.cc
st-pasha: The main reason was to have a  well-defined point when the `progress_bar` object in the progress manager should be destroyed. Now it is always destroyed when the last task is popped from the stack. Before we had the progress bar destroyed when the `tasks` becomes empty, but only if the last task finished successfully. If not, then we kept the progress bar instance was kept, and then cleaned up when the exception was caught on the python boundary.
However, this presents a problem if we catch exceptions earlier: basically we have to add some progress-bar-cleanup logic into every exception handler? This just seems utterly error-prone.

src/core/read/py_fread.cc
st-pasha: "i" means iterator: the return value is an iterator producing Frames.
The main reason to have this as a separate function is to make a clear distinction of what is returned by each function: `fread` returns a single Frame, `iread` returns an iterator of Frames.

src/datatable/utils/fread.py
st-pasha: Yeah, I've thought about it, but not sure about the correct approach.
Because  these are "fake" paths, where the name of an archive is treated as a folder: `data.zip/train.csv`. Is it better to use forward or backslash here? Do we need to perform slash-normalization? How does `zff` array look like, esp. if the archive has some nested directory structure inside?

src/core/read/py_fread.cc
oleksiyskononenko: I see. Then it could be somehow inconsistent with the existing `fread` function, where `f` stands for file (?) — that is an input parameter, while for `iread` the first letter `i` stands for the output type.

st-pasha: "fread" came from R data.table, where it meant "fast & friendly file finagler". Personally, I don't like the word "finagler", so I'd prefer "f" to mean something else. It could be "file" or "frame", or remain unspecified.

For `iread()` the letter `i` does mean "iterator", but it applies both to the input and the output: after all, in order to produce multiple frames in the output, you need to have multiple sources in the input as well.

That being said, I'm open to suggestions how the `iread` function can be called instead (renaming `fread` is impossible at this point though).

oleksiyskononenko: I see. To me, if a function is called `something_read()`, I would assume this `something` is the input type. Because then we say: read me a file, read me an archive, etc. Here, if `iread` is going to read various types, it could be called `xread` meaning that it could read from extra sources comparing to `fread`, that only reads files.

oleksiyskononenko: It could also prevent confusion when typed in uppercase, like now the title of this PR is `Iread ...` that first made me think that the function name starts with `L` :)

src/datatable/utils/fread.py
oleksiyskononenko: Good questions. I remember sometimes depending on the source OS an archive could have different slashes :/ So this needs to be investigated in more details.

src/core/read/py_fread.cc
st-pasha: well, `xread` is already reserved for "extract-read", see #1950

st-pasha: Also, `iread` reads from same sources as `fread`, except that it allows a multiple of them.

oleksiyskononenko: I see, then may be `mread`?

st-pasha: I considered `mread` as well, but ultimately thought that `iread` would be more familiar to python users: after all, there are many functions like `izip()`, `islice()`, `ifilter()`, `imap()`, etc (some of them from py2.7) which are iterator versions of functions `zip()`, `map()`, etc.  In this sense, `iread` is like an iterator version of `fread`.

PR #2399, opened 2020-04-06
Fix misc tests on Windows 

oleksiyskononenko:
Fix misc tests on Windows by
- using `os.path.join()` as much as possible in `fread` function and the relevant tests;
- using `pathlib` when creating local URL's;
- using `newline="\n"` when opening files in python;
- using a context manager when working with the zip files;
- deleting memory-mapped files only after the corresponding frames have been deleted.

Also, on Windows temporarily disable `test_fread_from_fileobj` test, that kills the entire pytest session when launched.

WIP for #2301

### DIFF
--- a/tests/fread/test_fread_api.py
+++ b/tests/fread/test_fread_api.py
@@ -207,9 +207,10 @@ def test_fread_from_stringbuf():
     assert d0.to_list() == [[1, 4], [2, 5], [3, 6]]
 
 
-@pytest.mark.skipif(sys.platform == "win32", 
-                    reason="On Windows this test makes pytest to stop")
-def test_fread_from_fileobj(tempfile):
+# This test is temporarily disabled in Windows, 
+# because there it makes pytest to stop abruptly.
+def test_fread_from_fileobj(tempfile, nowin):
+
     with open(tempfile, "w") as f:
         f.write("A,B,C\nfoo,bar,baz\n")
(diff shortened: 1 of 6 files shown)


tests/fread/test_fread_api.py
```diff
@@ -207,20 +210,20 @@ def test_fread_from_fileobj(tempfile):
...
-        frame_integrity_check(d0)
-        assert d0.source == tempfile
-        assert d0.names == ("A", "B", "C")
-        assert d0.to_list() == [["foo"], ["bar"], ["baz"]]
+    # with open(tempfile, "r") as f:
+    #     d0 = dt.fread(f)
+    #     frame_integrity_check(d0)
+    #     assert d0.source == tempfile
+    #     assert d0.names == ("A", "B", "C")
+    #     assert d0.to_list() == [["foo"], ["bar"], ["baz"]]
```
st-pasha, line 218: If this does not work on Windows just yet, how about making a pytest fixture that will skip this test on windows?

oleksiyskononenko: Yeah, this one doesn't work yet. It is commented out, so that when launched, this test doesn't kill pytest while I'm working on other things. I still need to understand why this one fails so bad, but if can't, we can always temporarily disable it on Windows only, agreed.

PR #2401, opened 2020-04-09
Fix type bump message generation

oleksiyskononenko:
`fread` could generate type bump messages that contain a sample of data. To limit the size of the message we have used `BUF_SIZE` variable, but this approach doesn't work well when the data size becomes sufficiently large. 

In this PR we fix the type bump message generation by applying the `BUF_SIZE` (now it is called `MAX_FIELD_LEN`) to the data only and not to the whole message. In such a way the other useful information in the message never gets truncated.

Also fix several issues introduced in the previous PR's and several larger fread tests.

WIP for #2301


oleksiyskononenko:
>So, if I understand correctly the issue here is that when the "field" is too long, it can truncate the message, in which case we won't see the line number?

Yes.

Well, the code throws two exceptions just because I think it is better to check the return value of `snprintf`. For the moment I can’t imagine the situation when it throws, so we can easily replace those with `xasserts`.

The reason I’ve added these checks is because In the current code launched on Windows `n` became `-1` and this value was passed along like it is normal.

Agreed on the partially repeated code.

oleksiyskononenko:
Btw, I didn't use streams here because thought there was a reason to stick to `snprintf`? Or this is just some old code that didn't use streams for historical reasons?

PR #2402, opened 2020-04-10
Cleanup & refactoring of dt::read::Column/Columns classes

st-pasha:
- Created `_dt.h` which forward-declares many classes used throughout the codebase. This can be used in place of "manual" forward-declaring that we do now.
- Renamed `dt::read::Column -> PreColumn` (in order to avoid confusion with top-level `Column` class)
- Renamed `dt::read::Columns -> PreFrame`
- general refactoring of both classes

This is a preparatory work before implementing the "fread directly to Jay file" functionality.


src/core/read/preframe.cc
```diff
@@ -0,0 +1,227 @@
...
+  return allocsize;
+}
+
+
+
+
+//------------------------------------------------------------------------------
+// Finalizing
+//------------------------------------------------------------------------------
+using dtptr = std::unique_ptr<DataTable>;
```
oleksiyskononenko, line 207: Since we're now doing some code cleaning, what about moving this definition to `_dt.h` (or some other header file), so that we include this file from everywhere else? Right now I see at least four duplicated definitions of `dtptr`, namely, `reader.h`, `aggregator.h`, `dt_ftrl_base.h` and `preframe.cc`.

src/core/_dt.h
```diff
@@ -0,0 +1,95 @@
...
+class RowIndex;
+class WritableBuffer;
+
+struct CString;
+enum class LType : uint8_t;
+enum class SType : uint8_t;
+
+using std::size_t;
+using colvec = std::vector<Column>;
+using strvec = std::vector<std::string>;
```
oleksiyskononenko, line 43: I like `_dt.h` quite a lot, may be also get rid of the duplicated `strvec` definitions in the code and include `_dt.h` instead? Right now there are at least 6 identical `strvec` definitions.

st-pasha: Absolutely, my thinking is that this header file can be used whenever we need to forward-declare any other class in a header. As such, "_dt.h" could eventually be included nearly everywhere. However, this comprehensive refactoring is not the purpose of this PR. It's just something that can be adjusted in the future as we go.

src/core/read/preframe.cc
st-pasha: yep, if you see duplicated definitions like that, then it's perfectly fine to add them to the _dt.h file. This file is just a collection of all forward declarations (all that I've found so far). It can be expanded in future revisions.

PR #2408, opened 2020-04-17
Use platform specific tolerance in certain fread tests

oleksiyskononenko:
Not all the platforms have a proper long double type that is used by fread to convert double literals into double numbers. In this PR we make use of platform specific tolerances for Windows and PowerPC64 that allow certain fread tests to not fail. 

Also, set up AppVeyor config file to launch tests for every Windows build.

WIP for #2301

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -54,6 +54,15 @@ def nowin():
         pytest.skip("Disabled on Windows")
 
 
+
+@pytest.fixture(scope="session")
+def tol():
+    # Tolerances for float comparisons on different OS's
+    tols = {"win32": 1e-14}
+
+    return tols.get(sys.platform, 0)
+
+
 @pytest.fixture(scope="session")
 def nocov():
     """Skip this test when running in the 'coverage' mode"""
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -54,6 +54,20 @@ def nowin():
...
+def tol():
+    """
+    This fixture returns a tolerance to compare floats
+    on a particular operating system. The reason we sometimes use
+    OS specific tolerance is that some platforms don't have a proper
+    long double type, resulting in a loss of precision
+    when fread converts double literals into double numbers.
+    """
+    tols = {"win32": 1e-15}
+    return tols.get(sys.platform, 0)
```
oleksiyskononenko, line 68: @st-pasha Do you think we also want to add ppc tolerance here? Right now I see only one test that is disabled in ppc due to the loss of precision.

st-pasha: sure

oleksiyskononenko: Looks like it is not fun to detect ppc https://github.com/h2oai/datatable/blob/master/tests/conftest.py#L33-L47 :)

Issue #2413, opened 2020-04-28
Fix multiple warnings generated by msvc


oleksiyskononenko:
msvc generates dozens of various warnings which should all be taken care of
- [x] template instantiation issues
- [x] libs related issues
- [ ] unsigned integer related issues
- [x] `if constexpr` warnings since msvc compiles in C++14 mode
- [x] unknown pragmas warnings
- [x] unreachable code warnings
- [ ] cast related warnings
- [x] empty control structures

Prerequisite for #1114

PR #2415, opened 2020-04-29
Fix warnings issued by msvc for libs, pragmas and some casts

oleksiyskononenko:
- fix warnings issued by msvc for libs, pragmas, template definitions and some casts;
- introduce and make use of `DT_COMPILER_CLANG`, `DT_COMPILER_GCC` and `DT_COMPILER_MSVC` everywhere, except for the libs code; 
- rename `DTDEBUG` to `DT_DEBUG`.

WIP for #2413

### DIFF
--- a/src/core/csv/reader_parsers.cc
+++ b/src/core/csv/reader_parsers.cc
@@ -10,7 +10,6 @@
 #include "read/fread/fread_tokenizer.h"  // FreadTokenizer
 #include "read/constants.h"              // hexdigits, pow10lookup
 #include "utils/assert.h"                // xassert
-#include "utils/macros.h"                
 
 static constexpr int8_t   NA_BOOL8 = -128;
 static constexpr int32_t  NA_INT32 = INT32_MIN;
--- a/src/core/csv/reader_parsers.cc
+++ b/src/core/csv/reader_parsers.cc
@@ -147,8 +147,8 @@ void parse_int_simple(FreadTokenizer& ctx) {
       (sd == MAX_DIGITS && value <= MAX_VALUE))
   {
     T x = static_cast<T>(value);
-    *reinterpret_cast<T*>(ctx.target) = (x ^ -static_cast<int>(negative))
-                                        + negative;
+    T neg = static_cast<T>(negative);
+    *reinterpret_cast<T*>(ctx.target) = (x ^ -neg) + neg; // same as neg? -x : x
     ctx.ch = ch;
   } else {
     *reinterpret_cast<T*>(ctx.target) = NA_VALUE;
--- a/src/core/csv/reader_parsers.cc
+++ b/src/core/csv/reader_parsers.cc
@@ -285,7 +285,7 @@ void parse_float32_hex(FreadTokenizer& ctx) {
       else if (E == 126 && Eneg && acc) /* subnormal */ E = 0;
       else goto fail;
     } else {
-      E = 127 + (E ^ -static_cast<uint32_t>(Eneg)) + Eneg;
+      E = 127 + (E ^ -static_cast<int>(Eneg)) + Eneg;
       if (E < 1 || E > 254) goto fail;
     }
     ctx.target->uint32 = (neg << 31) | (E << 23) | (acc);
@@ -576,7 +576,7 @@ void parse_float64_hex(FreadTokenizer& ctx) {
       else if (E == 1022 && Eneg && acc) /* subnormal */ E = 0;
       else goto fail;
     } else {
-      E = 1023 + (E ^ -static_cast<uint64_t>(Eneg)) + Eneg;
+      E = 1023 + (E ^ -static_cast<int>(Eneg)) + Eneg;
       if (E < 1 || E > 2046) goto fail;
     }
     ctx.target->uint64 = (neg << 63) | (E << 52) | (acc);
--- a/src/core/csv/reader_parsers.cc
+++ b/src/core/csv/reader_parsers.cc
@@ -10,7 +10,7 @@
 #include "read/fread/fread_tokenizer.h"  // FreadTokenizer
 #include "read/constants.h"              // hexdigits, pow10lookup
 #include "utils/assert.h"                // xassert
-#include "utils/macros.h"                // xassert
+#include "utils/macros.h"                
 
 static constexpr int8_t   NA_BOOL8 = -128;
 static constexpr int32_t  NA_INT32 = INT32_MIN;
--- a/src/core/csv/reader_parsers.cc
+++ b/src/core/csv/reader_parsers.cc
@@ -147,7 +147,8 @@ void parse_int_simple(FreadTokenizer& ctx) {
       (sd == MAX_DIGITS && value <= MAX_VALUE))
   {
     T x = static_cast<T>(value);
-    *reinterpret_cast<T*>(ctx.target) = (x ^ -negative) + negative;
+    *reinterpret_cast<T*>(ctx.target) = (x ^ -static_cast<int>(negative))
+                                        + negative;
     ctx.ch = ch;
   } else {
     *reinterpret_cast<T*>(ctx.target) = NA_VALUE;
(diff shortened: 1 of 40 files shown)


src/core/utils/macros.h
```diff
@@ -160,4 +160,18 @@ struct alignas(CACHELINE_SIZE) cache_aligned {
...
+#if DT_OS_WINDOWS
+  // Disable C4996 warning ("This function or variable may be unsafe")
+  // issued by MSVC for a fully valid and portable code
+  #pragma warning(disable : 4996)
+
+  // Disable C4127 warning ("consider using 'if constexpr' statement instead")
+  // as 'if constexpr' is not available in C++11
+  #pragma warning(disable : 4127)
+#endif
+
```
st-pasha, line 176: For other compilers we select which warnings to emit and which to hide in the build script: https://github.com/h2oai/datatable/blob/master/ci/ext.py#L297-L316. We can do the same with MSVC using `/wdnnnn` compile flag(s), see https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level.

Using pragmas like this is less optimal, because it means that (a) this include file has side-effects, (b) occasionally we include this file for the purpose of its side-effects only, (c) if during a subsequent refactor the include is removed, it would be very difficult to notice any problem (esp. since we don't watch for compile warnings on CI machines).

src/core/wstringcol.cc
```diff
@@ -115,7 +116,11 @@ char* writable_string_col::buffer_impl<T>::strbuf_ptr() const {
 template <typename T>
 void writable_string_col::buffer_impl<T>::write(const char* ch, size_t len) {
   if (ch) {
-    if (sizeof(T) == 4) xassert(len <= Column::MAX_ARR32_SIZE);
+    if (sizeof(T) == 4) {
+      (void) len;
```
st-pasha, line 120: Not sure what warning is this preventing? The variable `len` is clearly used on lines 124-126 below.

src/core/rowindex_slice.cc
```diff
@@ -41,8 +41,8 @@ bool check_slice_triple(size_t start, size_t count, size_t step, size_t max)
   return (start <= max) &&
          (count <= RowIndex::MAX) &&
          (count <= 1 || (step == 0) ||
-          (step <= RowIndex::MAX? step <= (max - start)/(count - 1)
-                                : step >= -start/(count - 1)));
+         (step <= RowIndex::MAX? step <= (max - start)/(count - 1)
+                               : step >= -start/(count - 1)));
```
st-pasha, line 45: I thought we decided that `-step <= start/(count - 1)` is better?

src/core/py_encodings.cc
```diff
@@ -50,7 +50,10 @@ int init_py_encodings(PyObject*) {
   for (int k = 0; k < 3; k++) {
     uint32_t* map = (k == 0)? win1252_map : (k == 1)? win1251_map : iso8859_map;
     for (unsigned int i = 0; i < 256; i++) {
-      if (i < 0x80) xassert(map[i] == i);  // check ASCII compatibility
+      if (i < 0x80) {
+        (void) i;
```
st-pasha, line 54: what warning does this fix?

src/core/lib/zlib/zutil.cc
```diff
@@ -3,12 +3,14 @@
...
  */
 #include "lib/zlib/zutil.h"
-namespace zlib {
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wcast-qual"
-#pragma clang diagnostic ignored "-Wold-style-cast"
+namespace zlib {
 
+#if defined(__clang__)
```
st-pasha, line 9: I think we should define a triple of macros `DT_COMPILER_CLANG`, `DT_COMPILER_GCC` and `DT_COMPILER_MSVC` in macros.h, similar to all other macros there. That way we'll have a more uniform and easy to remember spelling.

src/core/csv/reader_parsers.cc
```diff
@@ -216,9 +218,15 @@ void parse_intNN_grouped(FreadTokenizer& ctx) {
...
-    if (sizeof(T) == 4) ctx.target->int32 = negative? -x : x;
-    if (sizeof(T) == 8) ctx.target->int64 = negative? -x : x;
+    if (sizeof(T) == 4) {
+      int32_t x = static_cast<int32_t>(acc);
+      ctx.target->int32 = negative? -x : x;
+    }
+    if (sizeof(T) == 8) {
+      int64_t x = static_cast<int64_t>(acc);
+      ctx.target->int64 = negative? -x : x;
+    }
```
st-pasha, line 228: Can use the same approach as in line 150 above

src/core/csv/reader_parsers.cc
```diff
@@ -147,7 +148,8 @@ void parse_int_simple(FreadTokenizer& ctx) {
       (sd == MAX_DIGITS && value <= MAX_VALUE))
   {
     T x = static_cast<T>(value);
-    *reinterpret_cast<T*>(ctx.target) = (x ^ -negative) + negative;
+    *reinterpret_cast<T*>(ctx.target) = (x ^ -static_cast<int>(negative))
+                                        + negative;
```
st-pasha, line 152: if `x` is of type `T`, then shouldn't `negative` be cast to the same type?
```
T neg = static_cast<T>(negative);
*reinterpret_cast<T*>(ctx.target) = (x ^ -neg) + neg;  // same as neg? -x : x
```

src/core/utils/macros.h
oleksiyskononenko: Nice, didn't notice we have this capability in the `ext.py`, let me use it.

src/core/wstringcol.cc
oleksiyskononenko: I guess this one emitted a warning about an empty statement when you do `make build`, so I just used a noop on `len` to get rid of it.

src/core/rowindex_slice.cc
oleksiyskononenko: Yeah, but at some point I stopped fixing unsigned integer issues as we have quite a lot of them and they all need special attention (somewhere we may even have a bug). So will address all them in the subsequent PR.

src/core/py_encodings.cc
oleksiyskononenko: See above.

src/core/lib/zlib/zutil.cc
oleksiyskononenko: Yeah, could be useful. I also found that it is very easy to miss an include like `utils/macros.h` and then to have an incorrect `#if DT_SOMETHING` behavior without any getting any warnings. Not sure what could be a good way around it.

src/core/csv/reader_parsers.cc
oleksiyskononenko: I guess I casted it to `int` because `T` could only be `int32_t` or `int64_t`, we can of course cast it to `T` for flexibility and clarity.

src/core/wstringcol.cc
st-pasha: yeah, seems like ignoring it is a simpler alternative. We can even make it conditional on the build type: ignore in regular builds, leave as-is in debug builds

oleksiyskononenko: I guess `(void) len;` was just less code :)

src/core/lib/zlib/zutil.cc
st-pasha: For me it produces a warning:
```
warning: 'DT_SOMETHING' is not defined, evaluates to 0 [-Wundef]
  #if DT_SOMETHING
```

It's actually `#ifdef S0METHING` that never produces a warning, even if you make a typo in the spelling of the macro. In this respect our current approach is safer.

oleksiyskononenko: For some reason on Windows I never got the warning you post :/

src/core/wstringcol.cc
st-pasha: I think I've seen multiple examples like this in this PR. Plus, consider this also from the perspective of trying to eliminate future such warnings as well.

oleksiyskononenko: Yes, but now we have one additional line of code per such a case. With push/disable/pop pragmas it will be at lease seven lines per a case. Though we can define push/disable/pop as another macro :)

src/core/csv/reader_parsers.cc
oleksiyskononenko: The old code actually casted `x` to `T` here, but when `T` is `int64` msvc still think it is possible we assign `x` to `ctx.target->int32` (even though we never would), and issues a warning.

src/core/lib/zlib/zutil.cc
oleksiyskononenko: https://stackoverflow.com/questions/1643820/why-no-warning-with-if-x-when-x-undefined

src/core/csv/reader_parsers.cc
```diff
@@ -277,7 +284,7 @@ void parse_float32_hex(FreadTokenizer& ctx) {
       else if (E == 126 && Eneg && acc) /* subnormal */ E = 0;
       else goto fail;
     } else {
-      E = 127 + (E ^ -Eneg) + Eneg;
+      E = 127 + (E ^ static_cast<uint32_t>(-static_cast<int>(Eneg))) + Eneg;
```
st-pasha, line 287: much clearer:
```suggestion
      E = Eneg? 127 - E : 127 + E;
```
or use `127u` if compiler still complains.

src/core/csv/reader_parsers.cc
```diff
@@ -568,7 +575,7 @@ void parse_float64_hex(FreadTokenizer& ctx) {
       else if (E == 1022 && Eneg && acc) /* subnormal */ E = 0;
       else goto fail;
     } else {
-      E = 1023 + (E ^ -Eneg) + Eneg;
+      E = 1023 + (E ^ static_cast<uint64_t>(-static_cast<int>(Eneg))) + Eneg;
```
st-pasha, line 578: same as in line 287

src/core/csv/reader_parsers.cc
oleksiyskononenko: Sure, I thought there was some reason for the original cryptic formulation?

Issue #1750, opened 2019-03-27
streaming dt.fread(path).to_jay()


arnocandel:
Implement parsing a large .csv to a .jay file in a streaming fashion, without requiring more than a few GBs of RAM at most.

arnocandel:
https://github.com/h2oai/h2oai/issues/3518

st-pasha:
One way this can be achieved is by having a special parameter `memory_bound=` which would recommend a limit on how much RAM can be used by fread. If the dataset appears to be larger than this bound, a smaller frame will be created and saved to a temporary Jay file on disk. Then we will continue reading the remaining data, again possibly dumping it into a temporary Jay file. And so on. At the end, we will have 0 or more datasets residing in temporary folder on disk, and 1 "tail" dataset in memory. At this point we can perform "virtual rbind" of all these frames, returning a virtual Frame that contains all data from the input csv. It would be almost indistinguishable from the old fread behavior, except that the returned dataset will be virtual not material.

Saving this Frame to a Jay file would materialize the data directly on disk, essentially copying chunks from all temporary jay frames into the final destination.

Prerequisite: virtual column support (#1369)
Prerequisite: proper handling of temporary Jay files (i.e. we have to know when they are safe to delete).

arnocandel:
a script to push the boundaries on the rbind(many tables) memory limit and grouping:
https://github.com/h2oai/h2oai/blob/bf046974e812707e546815e0c46aecab69e3106a/tests/test_ts_support/big_data_stock.py

Issue #1097, opened 2018-06-05
fread should gather basic column stats while reading


st-pasha:
These should be: `nacount`, `min` and `max`.
This will allow us to avoid slower-than-usual first sort operation on a dataset (since min/max will already be calculated).

PR #2416, opened 2020-05-01
Fread to jay 

st-pasha:
This is the current state of the "streaming fread to Jay" feature. It is mostly working, however one test currently still fails. The tests for the feature itself still missing.

The following features are introduced:
- [x] New `memory_limit=` argument for fread can be used as a suggestion of how much RAM can be used during reading in order to maintain intermediate results / final output. This limit is not strict;
- [x] The `dt::read::PreColumn` class maintains a vector of column chunks that were captured during the reading (this design will also allow us to eliminate rereading pass, and allow the possibility of adding new columns on-the-fly);
- [x] New `TemporaryFile` class allows to create a temporary file on disk, which will be automatically deleted when it is no longer in use. New class `TemporaryFile_BufferImpl` is similar to `Mmap_BufferImpl`, but backed by a temporary file object. When the last buffer referring to a temporary file is gc'd, the file will be deleted; 
- [x] A "chunked" PreColumn will produce an `Rbound_ColumnImpl` column in the end;
- [x] The "rbound column" can be saved to Jay file without materializing.
- [x] setting `.nrows` should create a virtual "truncated column" instead of modifying the `.nrows` parameter on the existing column;

In order to make the feature working nicely, the following still missing:

- [ ] need support for calculating basic stats while freading, at the minimum maintain na count per column (#1097), this is currently a major source of slowdown;
- [ ] when opening a Jay file, the frame should keep track of this fact, especially if the file is temporary. Saving an opened temporary Jay file into another Jay file should be done via rename.

WIP for #1750

### DIFF
--- a/src/core/_dt.h
+++ b/src/core/_dt.h
@@ -96,6 +96,9 @@ namespace read {
   enum BT : uint8_t;
 
   struct FreadTokenizer;
+  union field64;
+
+  class ChunkCoordinates;
   class GenericReader;
   class PreColumn;
   class PreFrame;
--- a/src/core/_dt.h
+++ b/src/core/_dt.h
@@ -104,6 +104,7 @@ namespace read {
   class PreFrame;
   class ThreadContext;
 }}
+class FreadReader;
 
 
 #endif
--- a/src/core/_dt.h
+++ b/src/core/_dt.h
@@ -34,6 +34,7 @@ class DataTable;
 class Groupby;
 class RowIndex;
 class Stats;
+class TemporaryFile;
 class WritableBuffer;
 
 struct CString;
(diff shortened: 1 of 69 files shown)


dmillar:
On my system, (32GB RAM, Ubuntu 18.04) I get segfaults when reading larger delimited files, say around 15-20GB. To clarify the memory_limit implementation, is this in bytes or what measurement? If I set this to 16 or 16000000000, I get the same result (quick saturation of RAM and segfault).

If interested, some background on the segfaults: the only way I can read this size of data is to read smaller chunks and write immediately to jay then read back in. I get segfaults when rbinding chunks to one large frame.

st-pasha:
The `memory_limit=` parameter is in bytes, though it is not very strict when it comes to string columns (this is why this feature is still WIP). I'd recommend going down to 8GB limit and looking at the verbose logs to see whether anything unusual happens.

The largest remaining part for this feature to work properly, however, is implementation of #1097 in order to avoid accidental materialization of the frame when saving to Jay.

PR #2419, opened 2020-05-05
No pycode in random attacker

st-pasha:
Remove python code generation from random attacker; replaced with more detailed logging.

The reason for this change is that the effort required to maintain this functionality far outweighs its benefits.

### DIFF
--- a/tests_random/utils.py
+++ b/tests_random/utils.py
@@ -76,7 +76,7 @@ def assert_equals(frame1, frame2):
         assert same_iterables(frame1.to_list(), frame2.to_list())
 
 
-def log(fn):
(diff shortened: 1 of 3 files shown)


tests_random/utils.py
```diff
@@ -76,6 +76,20 @@ def assert_equals(frame1, frame2):
         assert same_iterables(frame1.to_list(), frame2.to_list())
 
 
+def log(fn):
```
oleksiyskononenko, line 79: May be use a name different from `math.log`?

st-pasha: sure

PR #2420, opened 2020-05-06
Add -pthread compile/link flag

st-pasha:
This seems to provide some benefits to multithreaded libraries, according to https://stackoverflow.com/questions/2127797


oleksiyskononenko:
Are you sure there is any effect on dt, since we're using `std::thread`s only?

st-pasha:
I'm not sure at all. However, I don't think there is any chance of having a detrimental effect either. And if there is even 5% chance that it will cause the standard library to use a more thread-safe alternative somewhere, then why not take that chance?

And btw here's another post that recommends using `-pthread`: https://gcc.gnu.org/legacy-ml/gcc-help/2007-01/msg00138.html

PR #2429, opened 2020-05-07
Escape column names with backticks in error messages

### LINKED ISSUE #2406: Column names containing backticks are not rendered properly in the error messages
When a column name contains a backtick and an exception is thrown, the backticks are either stripped out of the message or mess up the message format. Here is an MRE:

```python
from datatable import dt
DT0 = dt.Frame()

try:
  DT1 = DT0["e`fg"]
except Exception as e:
  print(e)

print(DT0["e`fg"])
```

The code above outputs
```
Column efg does not exist in the Frame
  backticks.py:9 in <module>  print(DT["e`fg"])

KeyError: Column efg` does not exist in the Frame
```

This issue is also a reason as to why some random attack tests (for instance, seed `3561896165173212362`) are failing now.

Attribute-To: @arnocandel

st-pasha:
Closes #2406

### DIFF
--- a/ci/xbuild/logger.py
+++ b/ci/xbuild/logger.py
@@ -151,6 +151,7 @@ class Logger1(Logger0):
             ch_future = sz - ch_queued - ch_done
             line += " [%s\x1B[1;97m%s\x1B[0;90m%s]" \
                     % ("#" * ch_done, "#" * ch_queued, "." * ch_future)
+            line += " %d/%d" % (self._n_finished, self._n_total)
         line += "\x1B[m\x1B[K"  # clear the rest of the line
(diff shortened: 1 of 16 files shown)


oleksiyskononenko:
Fun error on Windows, not sure why it appeared only on this PR
```
src\core\utils\file.cc(146): error C3861: 'S_ISDIR': identifier not found
```

st-pasha:
Perhaps because this definition was included transitively via exceptions.h -> types.h, but that include is now gone?

st-pasha:
Btw, I have a strong feeling that we'll need a windows-specific followup for this PR, in order to properly escape backslashes in file names...

PR #2430, opened 2020-05-07
Switch to C++14 language standard

oleksiyskononenko:
It seems that for [various reasons](https://github.com/h2oai/datatable/pull/2338) we can't switch to C++17, so in this PR I propose we switch to at least C++14. The reasons for this are
- we are now about to fully support Windows, and MSVC [doesn't support](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019) anything below C++14, so datatable is already built on Windows with the C++14 standard. Switching to C++14 for other operating systems/compilers will make things more consistent.
- C++14 has some [nice features](https://en.wikipedia.org/wiki/C%2B%2B14#New_language_features) comparing to C++11, for instance, [variable templates](https://en.cppreference.com/w/cpp/language/variable_template) and many others, that we can start using right away.

Besides of switching to C++14, this PR introduces the following changes
- `utils/c+++.h` header file has been removed and `dt::make_unique` calls have been replaced with the `std::make_unique` ones;
- `operator "" _s` definition has been removed from `utils/misc.h`;
- `itarget()` function template has been replaced with the polymorphic lambda in  `Ftrl<T>::fit_regression()`;
- documentation adjustments, in particular, support for `gcc 4.8` has been dropped because C++14 is supported only as of `gcc 5.0`.


st-pasha:
According to https://en.cppreference.com/w/cpp/compiler_support, the C++14 standard is supported from:
- GCC 5.0 (we use GCC 8.1 in docker images);
- CLang 3.4;
- MSVC 19.0 (which is VS 2015);
- Apple (supported in all known versions).

st-pasha:
> Switching to C++14 doesn't require any changes to the code, though a separate PR should be made to adjust documentation.

While current code happily compiles under C++14, there are some places which can be simplified under the new standard. For example, file `utils/c+++.h` can now be completely removed.

oleksiyskononenko:
Yes, `utils/c+++.h` has only a definition of `make_unique` so can be safely removed if we decide to switch to C++14.

st-pasha:
@mmalohlava I've checked [PEP-571](https://www.python.org/dev/peps/pep-0571/), and the manylinux2010 wheels that we produce are compatible with CentOS-6+.

PR #2434, opened 2020-05-12
Fread encoding= parameter

### LINKED ISSUE #2395: Support `encoding=` argument for fread
The default encoding is "auto", which almost always means "UTF-8", except for the cases when the encoding can be guessed from the BOM mark embedded into the file.

Otherwise we should support all encodings supported by python, and recode the file from the input encoding into UTF-8 temporary file.

See: https://github.com/h2oai/h2oai/issues/14126

st-pasha:
Added option `encoding=` which allows the input to be recoded from a non-standard encoding before it will be parsed by fread.

Closes #2395


src/core/csv/reader.cc
```diff
@@ -419,46 +429,43 @@ py::oobj GenericReader::read_all(py::robj pysources)
...
+  for (int i = 0; i < 5 && ch < eof; i++) {
+    if (newline) trace("%s", repr_source(ch, 100));
+    else         trace("...%s", repr_source(ch, 97));
+    const char* start = ch;
+    const char* end = std::min(eof, ch + 10000);
+    while (ch < end) {
+      char c = *ch++;
+      // simplified newline sequence. TODO: replace with `skip_eol()`
+      if (c == '\n' || c == '\r') {
+        if ((*ch == '\r' || *ch == '\n') && *ch != c) ch++;
```
oleksiyskononenko, line 454: Since we do `*ch++` above, shouldn't we also add a check `ch < end` here?

docs/releases/v0.11.0.rst
```diff
@@ -90,6 +90,11 @@
   -[new] It is now possible to read multi-file ``.tar.gz`` files using
     :func:`iread()`. [#2392]
 
+  -[new] Added parameter ``encoding`` which will force fread to recode the
+    input using the specified encoding before attempting to read it. The
+    recoding process uses standard python codecs, and is therefore
+    single-threaded.
```
oleksiyskononenko, line 96: What about using words `decode`/`decoding` instead of `recode`/`recoding` here? Since we're only decoding the input and not encode it again, also, the progress bar message is `Decoding ...`.

src/core/csv/reader.h
```diff
@@ -161,7 +162,7 @@ class GenericReader
     virtual ~GenericReader();
 
     py::oobj get_tempfiles() const;
-    py::oobj read_all(py::robj pysources);
+    // py::oobj read_all(py::robj pysources);
```
oleksiyskononenko, line 165: Could now be removed?

src/core/read/source.cc
```diff
@@ -55,7 +56,45 @@ Source_Python::Source_Python(const std::string& name, py::oobj src)
 
 
 py::oobj Source_Python::read(GenericReader& reader) {
-  return reader.read_all(src_);
+  reader.source_name = &name_;
+  // return reader.read_all(src_);
```
oleksiyskononenko, line 60: Could now be removed?

src/core/csv/reader.cc
```diff
@@ -656,79 +663,51 @@ const char* GenericReader::repr_binary(
...
-      if (ch == end && ch < eof) {
-        ch = start + 100;
-        newline = false;
-      } else {
-        newline = true;
-      }
-    }
-    trace("=====================");
-  }
-  t_open_input = wallclock() - t0;
```
oleksiyskononenko, line 716: I don't see this line moved anywhere, don't we measure `t_open_input` anymore?

src/core/csv/reader.cc
```diff
@@ -151,6 +154,17 @@ void GenericReader::init_nthreads(const py::Arg& arg) {
   }
 }
 
+
+void GenericReader::init_encoding(const py::Arg& arg) {
+  if (arg.is_none_or_undefined()) return;
+  encoding_ = arg.to_string();
+  if (!PyCodec_KnownEncoding(encoding_.c_str())) {
+    throw ValueError() << "Unknown encoding " << encoding_;
```
oleksiyskononenko, line 162: Do we have documentation added somewhere for the new `encoding` parameter? I guess it would be worth to mention that it is case insensitive, since It seems that both `PyCodec_KnownEncoding()` and `PyCodec_IncrementalDecoder()` are case insensitive.

st-pasha: added mention in the changelog

src/core/csv/reader.cc
st-pasha: good catch!

src/core/csv/reader.cc
st-pasha: yep, added it back

PR #2435, opened 2020-05-13
Logger class

### LINKED ISSUE #1902: Better logging infrastructure
This is a proposal to restructure and organize our current logging facilities under a single central authority. Currently we have log messages emitted from fread (via C-like `_message()` method), from csv writer (via `LogMessage` class), warnings via the `Warning` class, and possibly other places. Ideally, these should all use the same utility "Logger" class.

The API can be similar to what LogMessage / Exception / Warning classes do now. The critical part here is that the Logger should communicate to python in a thread-safe manner, especially when messages are sent from worker threads in a parallel region. The monitor thread should participate in this.

There should also be python-level options for controlling the logging: whether it is enabled at all, whether to print console, or to send messages to a custom logger object. Also there should be an option for anonymizing user data, as well as a helper methods for doing so (could be taken from fread's code).

st-pasha:
Update logging infrastructure in fread to use C++ stream-oriented writing methods (instead of old printf-style functions).

- Created a centralized `Logger` class in "utils/logger.h" which can handle logging from any C++ operation.
- Use this new class in fread()
- Use this new class in to_csv()

Closes #1902


src/core/utils/logger.cc
```diff
@@ -0,0 +1,273 @@
...
+
+template <>
+Message& Message::operator<<(const plural& pl) {
+  out_ << pl.count_ << ' ';
+  out_ << pl.str_;
+  if (pl.count_ != 1) {  // see misc.py for more advanced pluralization rules
+    out_ << 's';
+  }
+  return *this;
+}
```
oleksiyskononenko, line 155: May be one day we will port the advanced pluralization code from Python to C++, and then expose it back to Python as well as use it from C++.

st-pasha: Yeah, I think we can defer that to that time when the more advanced rules are actually needed. Until now we've been mostly pluralizing words like `row`, `column`, `thread` or `file`.

PR #2436, opened 2020-05-13
Improve fread EOL detection

### LINKED ISSUE #1343: fread may sometimes detect incorrect newline character
File having `\r` newline characters, but also `\n` in one of the fields (correctly quoted) is not parsed correctly:
```
>>> dt.fread("a,b,c\r1,4,foo\r7,99,\"wha\ntt?\"\r4,-1,nvm")
   | a         b  c.1   4  foo.7  99  "wha
-- + -------  --  ---  --  -----  --  ----
 0 | tt?"\r4  -1  nvm  NA     NA  NA    NA

[1 row x 7 columns]
```

### LINKED ISSUE #2074: Cannot parse airline sentiment dataset
* Tried to parse the airlines sentiment dataset from https://www.figure-eight.com/wp-content/uploads/2016/03/Airline-Sentiment-2-w-AA.csv but got the following:

```
In [3]: dt.fread("Airline-Sentiment-2-w-AA.csv")                                                                                                                                    
100% |ââââââââââââââââââââââââââââââââââââââââââââââââââ| (error) data
RuntimeError                              Traceback (most recent call last)
<ipython-input-3-046e05c26643> in <module>
----> 1 dt.fread("Airline-Sentiment-2-w-AA.csv")

~/h2oai/env/lib/python3.6/site-packages/datatable/fread.py in fread(anysource, file, text, cmd, url, columns, sep, dec, max_nrows, header, na_strings, verbose, fill, encoding, skip_to_string, skip_to_line, skip_blank_lines, strip_whitespace, quotechar, save_to, nthreads, logger, **extra)
     62     del params["extra"]
     63     freader = GenericReader(**params)
---> 64     return freader.read()
     65 
     66 

~/h2oai/env/lib/python3.6/site-packages/datatable/fread.py in read(self)
    676                 return res
    677             else:
--> 678                 return core.gread(self)
    679         finally:
    680             self._clear_temporary_files()

RuntimeError: Too few fields on line 1: expected 3 but found only 2 (with sep=';'). Set fill=True to ignore this error.  <<_unit_id,_golden,_unit_state,_trusted_judgments,_last_judgment_at,airline_sentiment,airline_sentiment:confidence,negativereason,negativereason:confidence,airline,airline_sentiment_gold,name,negativereason_gold,retweet_count,text,tweet_coord,tweet_created,tweet_id,tweet_location,user_timezone\r681448150,FALSE,finalized,3,2/25/15 5:24,neutral,1,,,Virgin America,,cairdin,,0,@VirginAmerica What @dhepburn said.,,2/24/15 11:35,5.70306E+17,,Eastern Time (US & Canada)\r681448153,FALSE,finalized,3,2/2...>>
```

st-pasha:
Improved logic for newline detection in fread. Now files with `\r`-newlines are recognized more reliably.

Closes #1343 
Closes #2074


src/core/csv/reader_fread.cc
```diff
@@ -754,30 +754,57 @@ void FreadReader::detect_header() {
...
+  * character is treated as a line separator. When the value is false,
+  * a standalone '\\r' does not start a new line.
+  *
+  * Only the first 64Kb of text will be scanned. If any \\n is found,
+  * then cr_is_newline will be set to false. However, if at least 10
+  * \\r are found without encountering a single \\n, then
+  * cr_is_newline will be set to true. If after reading 64Kb of input
+  * only \\r's were seen, then cr_is_newline will again be set to
+  * true. If neither \\r's nor \\n's were seen, then cr_is_newline
+  * will be set to false (i.e. newlines are \\n-based).
```
oleksiyskononenko, line 768: What about creating python options for `64Kb` and `10`?

st-pasha: If the need ever arises, it is much easier to expose the `eol` parameter to fread.
Because, what we're really trying to detect here are files that use `\r` as a line separator, which doesn't happen very often. According to Wikipedia (https://en.wikipedia.org/wiki/Newline), this was used for "Commodore 8-bit machines (C64, C128), Acorn BBC, ZX Spectrum, TRS-80, Apple II series, Oberon, the classic Mac OS, MIT Lisp Machine and OS-9".  While apparently some csv files still remain from those times, it is highly unlikely that they would have line lengths > 64Kb.

Going in the opposite direction, we're running the risk that a `\n`-using file will have an unquoted field with 10+ `\r`s in the very first line (which has to be the header line usually). Again, incredibly unlikely, and the authors of the file are supposed to quote those fields anyways.

oleksiyskononenko: Yes, so why not to expose the eol parameter to fread? I was like it is already exposed and on this PR we just improve auto-detection.

st-pasha: We should, it's just not within the scope of this PR. Another parameter that we also want exposed is the `quote_rule`.

PR #2439, opened 2020-05-14
Support building wheels on Windows

### LINKED ISSUE #2432: Support building wheels on Windows
Prerequisite for #1114

oleksiyskononenko:
- improve `ABI` detection in xbuild to allow building wheels on Windows;
- build/test wheels for all the Python versions on AppVeyor.

Closes #2432

### DIFF
--- a/ci/xbuild/wheel.py
+++ b/ci/xbuild/wheel.py
@@ -410,17 +410,43 @@ class Wheel:
     #---------------------------------------------------------------------------
 
     def _get_python_tag(self):
+        from packaging import tags
         impl = sys.implementation.name
         assert impl == "cpython"
-        major = sys.version_info.major
-        minor = sys.version_info.minor
-        return "cp" + str(major) + str(minor)
+        return "cp" + tags.interpreter_version()
+
 
     def _get_abi_tag(self):
-        soabi = sysconfig.get_config_var("SOABI")
-        parts = soabi.split("-")
-        assert parts[0] == "cpython"
-        return "cp" + parts[1]
+        """
+        Return the ABI tag if available, otherwise just emulate it.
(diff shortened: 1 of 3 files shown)


ci/xbuild/wheel.py
```diff
@@ -416,11 +416,38 @@ def _get_python_tag(self):
...
+                self.log.report_abi_variable_missing(var)
+                return fallback
+            return val
+
+        soabi = sysconfig.get_config_var('SOABI')
+        if soabi:
+            parts = soabi.split("-")
+            assert parts[0] == "cpython"
+            abi = 'cp' + parts[1]
+        elif hasattr(sys, 'maxunicode'):
```
st-pasha, line 438: According to https://docs.python.org/3/library/sys.html, `sys.maxunicode` was changed in 3.3, which means it existed even before that. Since we're not targeting pythons before 3.5, this branch should never happen. So I'd suggest changing this line to simple `else:`

ci/xbuild/wheel.py
```diff
@@ -416,11 +416,38 @@ def _get_python_tag(self):
...
+            d = ''
+            m = ''
+            if get_flag('Py_DEBUG', hasattr(sys, 'gettotalrefcount')):
+                d = 'd'
```
st-pasha, line 442: it's probably simpler to start with `abi = self._get_python_tag()`, and then do `abi += 'd'` and `abi += 'm'` as necessary. That way we don't introduce too many extra variables

ci/xbuild/wheel.py
oleksiyskononenko: Yeah, I've cleaned up most of the old branches and this one could also be cleaned up. This PR is still WIP as I wanted to make sure wheels can be generated on all the platforms first. We can now refactor and adjust it.

ci/xbuild/wheel.py
oleksiyskononenko: Sure, these are artifacts from the old code.

Issue #2440, opened 2020-05-14
Use same versioning scheme in AppVeyor as in Jenkins


st-pasha:
The wheels built in AppVeyor have to have the same internal version number as those that are produced on Jenkins. In order to achieve this, we need to:

1. Use `git rev-list --count master` as a build number both on Jenkins and AppVeyor (as opposed to Jenkins internal counter). 
2. Adjust the AppVeyor script so that it sets up the variables `DT_RELEASE`, `DT_BUILD_NUMBER`  and `DT_BUILD_SUFFIX` before building wheels. See https://github.com/h2oai/datatable/blob/master/ci/ext.py#L74-L87 for how they are used. The logic for setting up those variables should be the same as in Jenkins.
3. The wheel files should also be listed as artifacts in the AppVeyor manifest.

PR #2442, opened 2020-05-15
Wheel generation adjustments

oleksiyskononenko:
As of this PR, instead of the Jenkins internal counter
- master builds will use `git rev-list --count master` as a `DT_BUILD_NUMBER`
- other branches will use `git rev-list --count master..` in the `DT_BUILD_SUFFIX`

This is now valid for both AppVeyor and Jenkins making the wheel names consistent. Travis still uses its own naming scheme. 

In addition, we now do not use `scm` when checking out the code on Jenkins and do a simple `git clone` instead. This makes the structure of the source code directory cleaner and consistent with github.

Closes https://github.com/h2oai/datatable/issues/2440


ci/Jenkinsfile.groovy
```diff
@@ -142,10 +143,20 @@ ansiColor('xterm') {
...
+                                sh "git checkout master"
+                                sh "git pull"
+                                sh "git checkout ${env.CHANGE_BRANCH}"
+                                def BRANCH_BUILD_ID = sh(script:
+                                  'git rev-list --count master..',
+                                  returnStdout: true
+                                )
+                                sh "git checkout master"
+                                DT_BUILD_SUFFIX = env.BRANCH_NAME.replaceAll('[^\\w]+', '') + "." + BRANCH_BUILD_ID
+                            }
```
st-pasha, line 159: I'm still not convinced using a shared repository is a good idea. Consider the following scenario: branch A is forked from master, there are few commits there, and then it is converted into a PR. All is good:
```
      /---*---*---* (A)
-----o  (master)
```
Now, it turns out we need to make some fixes in the RP, so a new commit is pushed into branch A. However, in the meanwhile, some other PR was merged to master, and the tree now looks like this:
```
      /---*---*---*---* (A)
-----o----o  (master)
```
There is no longer a direct path between A and master, and the rev-list-count command will fail (I think)

st-pasha: Or another scenario: 2 separate PRs attempt to run their tests at the same time, and consequently they run `git pull` on the same local repository at the same time. Since git was not designed for a parallel local access, a data corruption may occur.

oleksiyskononenko: I'm also not sure this is a good idea. May only work if we have 1 PR running on 1 machine at the same time, and if no one touches the `/home/jenkins/repos/datatable` folder meanwhile (I don't think we have a guarantee for that). For the moment I want to see if the new PR build id suffix will work as expected on both AppVeyor and Jenkins, and it seems it does work. Now we just need to make it safe.

>There is no longer a direct path between A and master, and the rev-list-count command will fail (I think)

This is exactly the situation we have now, and I don't think anything failed because of the fact https://github.com/h2oai/datatable/pull/2445 was merged.

oleksiyskononenko: Well, there is actually a problem for the last commit Jenkins has `18` as the PR build suffux and AppVeyor has `19` (it seems that AppVeyor is wrong...).

ci/Jenkinsfile.groovy
```diff
@@ -138,14 +137,23 @@ ansiColor('xterm') {
...
+
                         if (isRelease()) {
                             DT_RELEASE = 'True'
                         }
                         else if (env.BRANCH_NAME == 'master') {
-                            DT_BUILD_NUMBER = env.BUILD_ID
+                            DT_BUILD_NUMBER = sh(
+                              script: "git rev-list --count master | tr -d '\n'",
+                              returnStdout: true
+                            )
```
st-pasha, line 149:
```suggestion
                            DT_BUILD_NUMBER = sh(
                              script: "git rev-list --count master",
                              returnStdout: true
                            ).trim()
```
(and same below)

oleksiyskononenko: Yeah, let's do it.

PR #2468, opened 2020-06-03
Debug logger feature

### LINKED ISSUE #2452: Add DataTable to support verbose mode for all operations
DataTable supports verbose mode for limited operations now ( fread() / to_csv() ) but it would be great to add verbose support for all operations of DT.

as part of verbose mode, if we can get performance metrics (time, memory, disk), it would be great.  With this enhancement our performance engineers will be able to develop/debug performance issues fast.  

Thanks DT team.

st-pasha:
- New option `dt.options.debug.enabled` can be used to turn on verbose logging to each function/method call in datatable's core. Additionally, `dt.options.debug.logger` can be set to a user-provided logger object;
- By default, the logger reports the name of each function called, and how long it took;
- Option `dt.options.debug.report_args` will additionally print the arguments of all function calls;
- Option `dt.options.debug.max_arg_size` allows you to control the repr size of each argument. This option is needed in order to prevent objects with excessively long reprs (e.g. some arrays or strings) to overwhelm the output.
- Adjusted repr of `Frame` to make it less verbose while more informative.
- Fixed bug with reprs of "rowwise" functions.

Example log output:
```
<Option debug.report_args>.set() # 6.17e-07 s
<Frame#102ba02d0>.__init__(range(0, 10)) # 3.774e-05 s
<Frame#102ba02d0 10x1>.nrows = 15 # 1.0703e-05 s
<Frame#102ba02d0 15x1>.nrows # 3.94e-07 s
<Frame#102ba02d0 15x1>[::2, :] {
  <Frame#102ba0120>.__init__() # 4.01e-07 s
} # 0.000126431 s
<Frame#102ba0120 8x1>.__del__() # 1.885e-06 s
<Frame#102ba02d0 15x1>[<class 'int'>] = <class 'str'> # 0.000275428 s
dt.cbind(<Frame#102ba02d0 15x1>, <Frame#102ba02d0 15x1>, force=True) {
  <Frame#102ba0300>.__init__() # 2.33e-06 s
  dt.apply_color('yellow', 'DatatableWarning: ') # 9.003e-06 s
  dt.apply_color('grey', "Duplicate column name found, and was assigned a unique name: 'C0' -> 'C1'") # 7.059e-06 s
} # 0.000250906 s
<Frame#102ba0300 15x2>.__del__() # 1.091e-06 s
<Frame#102ba02d0 15x1>.__del__() # 7.711e-06 s
```

Closes #2452


oleksiyskononenko:
I guess there is something wrong after ` dt.options.debug.logger` has been set to invalid logger object, see below

```
>>> import datatable as dt
>>> dt.options.debug.logger=1
>>> dt.options.debug.logger
unable to emit log message
1
>>> dt.options.debug.logger="default"
unable to emit log message
```

Btw, what about having this option `False` by default, replace `default` with `True` and then have a separate option for the custom logger object?

st-pasha:
You're right, it somehow gets stuck in an invalid mode...

As for having 2 options, I tried that first, but it was super annoying where you had to change 2 options in order to get logging to work; and setting the other option's value automatically doesn't work will with option contexts.

oleksiyskononenko:
I guess the most popular use case will be just to turn on the debug logger, and if this option is set to `False` by default then it would be obvious, from the user perspective, how to turn it on. Right now, one would first need to spend some time reading documentation and figure out the magic `"default"` value.

I agree that setting two options could be annoying, but this will be required only in the case of the custom logger object. Only advanced users will do that, so it shouldn't be a problem for them to set another option to `True`. Especially that this operation is normally needed to be done once for the whole project.

st-pasha:
Fair point, I'll try to go back to two options

oleksiyskononenko:
You might also want to adjust the initial message on this PR, as it normally becomes a commit message.

PR #2475, opened 2020-06-12
Fix documentation of Frame.colindex()

st-pasha:
For some reason the docs are saying `nrows` when really it is `ncols`.


st-pasha:
Failing tests are due to a broken external dependency: https://github.com/pypa/virtualenv/issues/1857. They are fixing it right now

PR #2477, opened 2020-06-16
Enable building debug-mode wheels

st-pasha:
Currently the wheels are built in "production" mode. However, occasionally there may be a need to produce a wheel with a "debug" version of the core library. This is now possible via
```
python ci/ext.py debug
python ci/ext.py wheel --nobuild
```

### DIFF
--- a/ci/ext.py
+++ b/ci/ext.py
@@ -512,20 +512,24 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
     assert isinstance(wheel_directory, str)
     assert isinstance(config_settings, dict)
     assert metadata_directory is None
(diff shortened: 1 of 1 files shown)


oleksiyskononenko:
Nice, what about the wheel file name in this case? Will we be able to distinguish between the "build" wheel and the "custom" wheel?

st-pasha:
This is how the wheel is called when I'm building it locally: `datatable-0.11.0a0+custom.1592343172.pasha-cp36-cp36m-macosx_10_9_x86_64.whl`. In this case "custom" indicates that it used a non-standard mechanism for wheel building.

However, if env.variables like `DT_BUILD_SUFFIX` were defined, then the build version would not have any indication.

PR #2479, opened 2020-06-18
Macros for compiler warnings handling

st-pasha:
- Defined macros `DISABLE_CLANG_WARNING(N)`, `DISABLE_GCC_WARNING(N)`, `DISABLE_MSVC_WARNING(N)` and corresponding `RESTORE_*` macros, to reduce the amount of boilerplate needed to switch off a warning for a portion of code;
- Renamed `DT_OS_DARWIN` -> `DT_OS_MACOS`;
- Rename member variables in class `monitor_thread`;

### DIFF
--- a/src/core/column/sentinel_fw.cc
+++ b/src/core/column/sentinel_fw.cc
@@ -139,11 +139,8 @@ size_t SentinelFw_ColumnImpl<T>::memory_footprint() const noexcept {
 // Data access
 //------------------------------------------------------------------------------
 
-#if DT_COMPILER_MSVC
-  #pragma warning(push)
-  // unreachable code
-  #pragma warning(disable : 4702)
-#endif
+DISABLE_MSVC_WARNING(4702)
+
 
 template <typename T>
 bool SentinelFw_ColumnImpl<T>::get_element(size_t i, int8_t* out) const {
@@ -201,9 +198,8 @@ bool SentinelObj_ColumnImpl::get_element(size_t i, py::robj* out) const {
   return !x.is_none();
 }
 
-#if DT_COMPILER_MSVC
-  #pragma warning(pop)
-#endif
+
+RESTORE_MSVC_WARNING(4702)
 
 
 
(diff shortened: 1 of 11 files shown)


src/core/column/sentinel_fw.cc
```diff
@@ -139,11 +139,8 @@ size_t SentinelFw_ColumnImpl<T>::memory_footprint() const noexcept {
 // Data access
 //------------------------------------------------------------------------------
 
-#if DT_COMPILER_MSVC
-  #pragma warning(push)
-  // unreachable code
-  #pragma warning(disable : 4702)
-#endif
+DISABLE_MSVC_WARNING(4702)
```
oleksiyskononenko, line 142: May be don't remove the comment line `// unreachable code`, so that we know what warning it is.

src/core/utils/misc.cc
```diff
@@ -23,16 +23,16 @@
...
 // within if() conditions that get statically ignored.
-#if DT_COMPILER_GCC
-  #pragma GCC diagnostic push
-  #pragma GCC diagnostic ignored "-Wshift-count-overflow"
-#endif
+DISABLE_GCC_WARNING("-Wshift-count-overflow")
+// 4293: shift count negative or too big, undefined behavior
+DISABLE_MSVC_WARNING(4293)
+// 4333: right shift by too large amount, data loss
+DISABLE_MSVC_WARNING(4333)
```
oleksiyskononenko, line 35: Where do we restore these?

st-pasha: Technically we don't need to worry about restoring warnings at the end of translation unit, since the compiler exits at that point anyways.
But I can restore them if you want.

src/core/column/sentinel_fw.cc
st-pasha: Oh, right. I didn't realize that comment was describing the warning.

src/core/utils/misc.cc
oleksiyskononenko: Yeah, but originally these warnings were silenced only for several lines of code, and now they're silenced for the whole translation unit. Who knows how the code in `misc.cc` will evolve, so why not to silence the warnings only on the required lines?

PR #2491, opened 2020-06-26
Fix broken links on documentation site

st-pasha:


oleksiyskononenko:
We may still have some issues after re-organizing pages. For insance, switching to the `latest` build on this page https://datatable.readthedocs.io/en/v0.10.1/using-datatable.html results in 404.

PR #2508, opened 2020-06-28
CString refactoring

### LINKED ISSUE #2502: Replacing NA strings in a column
When NA strings are replaced in a column, all empty strings get replaced too
```
import datatable as dt
DT = dt.Frame(A=["foo", "", None, "bar"])
DT.replace(None, "buzz")
print(DT.to_list())

# Produces:
# [['foo', 'buzz', 'buzz', 'bar']]

# Expected:
# [['foo', '', 'buzz', 'bar']]
```

st-pasha:
- class `CString` moved into file "cstring.h" / "cstring.cc";
- class `CString` moved into the `dt::` namespace;
- all constructors in `CString` are now explicit;
- fields `.ch` and `.size` are now private;
- field `.size` changed into `size_t` instead of `int64_t`;
- fixed a bug with replacing NA strings in function `replace()` (#2502);

Closes #2502


src/core/models/dt_ftrl.cc
```diff
@@ -970,8 +970,8 @@ dtptr Ftrl<T>::create_p(size_t nrows) {
   for (size_t i = 0; i < nlabels; ++i) {
     CString val;
     bool isvalid = col0_str64.get_element(i, &val);
-    labels_vec[i] = isvalid? std::string(val.ch, static_cast<size_t>(val.size))
-                           : std::string();
+    labels_vec[i] = val.to_string();
+    xassert(isvalid == !val.isna());  (void)isvalid;
```
oleksiyskononenko, line 974: For invalid elements, do we always return `val` that has `ptr_ == nullptr`? Or there could be some garbage?

st-pasha: You're right, there could be garbage (as with all other types).

st-pasha: I fixed it

PR #2509, opened 2020-06-29
[DOC] Fread usage examples

samukweku:
This is a collection of usage examples for `fread`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "restructuredtext.confPath": "${workspaceFolder}/docs"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "restructuredtext.confPath": "${workspaceFolder}/docs"
-}
--- a/docs/manual/fread_examples.rst
+++ b/docs/manual/fread_examples.rst
@@ -1,14 +1,36 @@
-
+.. py:currentmodule:: datatable
 
 Fread Examples
 =================
 
+This function is capable of reading data from a variety of input formats (text files, plain text, files embedded in archives, excel files, ...), producing a Frame as the result. You can even read in data from the command line.
+
+See :func:`fread()` for all the available parameters.
+
+**Note:** If you wish to read in multiple files, use :func:`iread()`; it returns an iterator of Frames.
+
 Read Data
 ----------
 
-Read text data directly::
+- Read from text file::
+
+
 
     from datatable import dt, fread
+
+    result = fread('iris.csv')
+    result.head(5)
+
+        sepal_length	sepal_width	petal_length	petal_width	species
+    0	5.1	            3.5	            1.4	            0.2	        setosa
+    1	4.9	            3	            1.4	            0.2	        setosa
+    2	4.7	            3.2	            1.3	            0.2	        setosa
+    3	4.6	            3.1	            1.5	            0.2	        setosa
+    4	5	            3.6	            1.4	            0.2	        setosa
+
+
+- Read text data directly::
+
 
     data = ('col1,col2,col3\n'
             'a,b,1\n'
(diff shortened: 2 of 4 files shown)


docs/manual/fread_examples.rst
```diff
@@ -0,0 +1,490 @@
...
+     #You specify the `cmd` parameter
+     #Here we filter data for the year 2015
+     fread(cmd = """cat netflix.tsv | awk 'NR==1; /^2015-/'""")
+
+The command line can be very handy with large data; you can do some of the preprocessing before reading in the data to `datatable`.
+
+Dealing with Unicode Data
+-------------------------
+
+`Fread` is pretty powerful and can automatically detect the encoding in a number of cases.::
```
st-pasha, line 70: Actually, we do very little encoding guessing. Only if it's a UTF16 file with a BOM mark then we can guess the encoding automatically. In all other cases we assume the file is in UTF8 (which most files should be nowadays). If the file is in any other encoding, that should be declared explicitly via the `encoding=` parameter.

docs/manual/fread_examples.rst
```diff
@@ -0,0 +1,490 @@
...
+        1	2	bøb
+        2	3	cárol
+
+You can also specify the encoding; `fread` will recode into `UTF-8` before reading.
+Check `here <https://docs.python.org/3/library/codecs.html#standard-encodings>`_ for Python's list of standard encodings.
+
+Detect Thousand Separator
+-------------------------
+
+Thousand separators are automatically detected::
```
st-pasha, line 91: Actually we don't detect the thousand separator, we just assume it's `,`. 
Looks like this parameter is not even properly exposed to the user, though it exists internally.

I guess we could set it automatically based on `dec`.

docs/manual/fread_examples.rst
```diff
@@ -0,0 +1,490 @@
...
+
+        ID	    Charges  Payment_Method
+    0	634-VHG	    28	     Cheque
+    1	365-DQC	    33.5     Credit card
+    2	264-PPR	    631	     NA
+    3	845-AJO	    42.3     NA
+    4	789-KPO	    56.9     Bank Transfer
+
+
+For rows with null values, set `fill=True`::
```
st-pasha, line 148: Better say "incomplete rows", i.e. rows with less values than in other rows. This is different from null values.

docs/manual/fread_examples.rst
```diff
@@ -0,0 +1,490 @@
...
+You can change `some` of the column names via a `dictionary`::
+
+    fread(data, columns={"a":"A", "b":"B"})
+
+        A	B	c	d
+    0	1	2	3	4
+    1	5	6	7	8
+    2	9	10	11	12
+
+By deafult, `fread` takes the first line in the data as the header. If, however, you do not want the first line as the header, set the `header` parameter to False::
```
st-pasha, line 223: Fread will use heuristics to determine whether the first row is data or not (for example see your first example in this section). It may occasionally guess incorrectly, in which case `header=` parameter will be useful.

docs/manual/fread_examples.rst
```diff
@@ -0,0 +1,490 @@
...
+
+- Select with a `dictionary`::
+
+        data = ('a,b,c,d\n'
+                '1,2,3,4\n'
+                '5,6,7,8\n'
+                '9,10,11,12')
+
+        #pass Ellipsis:None to discard any columns that are not needed
+        fread(data, columns={"a":"a", Ellipsis:None})
```
st-pasha, line 347: `Ellipsis` is the same as `...`, but the latter looks better: `{"a": "a", ...: None}`

samukweku:
Thanks @st-pasha! I’ll work on your review and update

docs/manual/fread_examples.rst
```diff
@@ -0,0 +1,490 @@
...
+
+    url = "https://raw.githubusercontent.com/Rdatatable/data.table/master/vignettes/flights14.csv"
+    fread(url)
+
+Read from an archive. If there are multiple files, only the first will be read; you can specify the path to the specific file you are interested in::
+
+
+    fread("data.zip/mtcars.csv")
+
+Read from `.xlsx` files - `.xls` files are not supported::
```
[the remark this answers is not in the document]
samukweku: An error on my part.

docs/manual/fread_examples.rst
```diff
@@ -22,73 +44,66 @@ Read text data directly::
...
+You can specify the encoding via the ``encoding`` parameter; ``fread`` will recode into ``UTF-8`` before reading::
 
     data = ("num,name\n"
             "1,alïce\n"
             "2,bøb\n"
             "3,cárol\n"
             )
 
-    fread(data)
+    fread(data, encoding = "latin-1")
```
st-pasha, line 94: The concept of "encoding" means how certain abstract characters are stored at byte level. As such, it makes sense to talk about the encoding of a file (because a file is just a sequence of bytes), or the encoding of a `bytes` object, but there is no such thing as "encoding" of a python string (a python string has an underlying encoding, of course, but it's complicated, and mutable, and invisible to the user).

In short, when input is a string the concept of encoding does not apply. Or at least shouldn't. In your example it will actually garble the input, which is a bug.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "restructuredtext.confPath": "${workspaceFolder}/docs"
+}
```
st-pasha, line 3: Looks like this was included by accident. Feel free to change `.gitignore` to add this path to the list of ignored files.

docs/manual/fread_examples.rst
samukweku: ok. had a different understanding of encoding. Thanks for clearing that.

.vscode/settings.json
samukweku: Will do

oleksiyskononenko:
@samukweku I've just merged the most recent `master` onto your branch, this should resolve issues with some failing checks for your PR.

samukweku:
Thanks @oleksiyskononenko !, I'll run another push

.vscode/settings.json
st-pasha: Could you delete this file from the PR too, please?

PR #2513, opened 2020-06-30
Debug build

### LINKED ISSUE #2501: Add a Jenkins job for building/testing a debug version of datatable
All current artifacts are built & tested in production mode, we should make at least one to also build in debug mode where the internal asserts will be properly checked. Also, the debug wheel can then be used to help other users debug any issues they might see in their pipelines.

Use python version 3.6 and manylinux2010_x86_64 platform.

st-pasha:
Closes #2501

### DIFF
--- a/tests/test_parallel.py
+++ b/tests/test_parallel.py
@@ -28,14 +28,8 @@ import subprocess
 import sys
 import time
 from datatable.lib import core
+from tests import (cpp_test, skip_on_jenkins)
 
-
-#-------------------------------------------------------------------------------
-# Check if we need to run C++ tests
-#-------------------------------------------------------------------------------
-
-cpp_test = pytest.mark.skipif(not hasattr(core, "test_coverage"),
-                              reason="C++ tests were not compiled")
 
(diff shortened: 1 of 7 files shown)


tests/test_parallel.py
```diff
@@ -145,6 +139,7 @@ def test_progress(parallel_type, nthreads):
 
 # Send interrupt signal and make sure process throws KeyboardInterrupt
 @cpp_test
+@skip_on_jenkins
```
oleksiyskononenko, line 142: Any ideas as to why this test stopped working for a debug build? I assume it will now be skipped even for the normal builds on Jenkins...

st-pasha: All `@cpp_test`s are skipped on Jenkins/Travis/AppVeyor anyways, because they require a debug build... As to what went wrong - i don't know. Just the progress bar was never interrupted until the very end. Could be either that monitor's priority was so low that it never had a chance to run, or that somehow the signal couldn't get delivered to the monitor thread.

PR #2530, opened 2020-07-10
[DOC]Groupby examples documentation

### LINKED ISSUE #2363: Documentation Request: using the new groupby expressions feature
I have been v interested in this feature since datatable came out - and now it's here! (actually since November).   While I do plan to experiment with it an actual list of illustrative usages would be quite handy to add somewhere to the docs.  The authors may know about nifty abilities/usages that I would not likely discover just ad-hoc.  Thanks!

samukweku:
This is a collection of examples for groupby - it aims to improve the documentation for datatable and show various uses of the `by` function.

Closes #2363

### DIFF
--- a/docs/manual/groupby_examples.rst
+++ b/docs/manual/groupby_examples.rst
@@ -1,0 +1,755 @@
+.. py:currentmodule:: datatable
+
+Grouping with ``by``
+====================
+
+The ``by`` modifier splits a dataframe into groups, either via the provided column(s) or an ``f-expression``, and then applies ``i`` and ``j`` within each group.  This `split-apply-combine <https://www.jstatsoft.org/article/view/v040i01#:~:text=Abstract%3A,all%20the%20pieces%20back%20together.>`_ strategy allows for a number of operations:
+
+- Aggregations per group
+- Transformation of a column or columns, where the shape of the dataframe is maintained
+- Filtration, where some data are kept and the others discarded, based on a condition or conditions
+
+See :func:`by()` for more details.
+
+Aggregation
+-----------
+
+The aggregate function is applied in the ``j`` section.
+
+- Group by one column::
+
+    from datatable import dt, f, by, ifelse, update, sort, count, min, max, mean, sum, rowsum, count
+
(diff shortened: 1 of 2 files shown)


samukweku:
Do I need to manually add my name to the list of contributors? If so, what does the number next to the names signify? number of approved PRs?

st-pasha:
The list of contributors is generated semi-automatically via script `ci/gh.py`. Once you run the script for the correct milestone, it will report all contributors, which then needs to be pasted into the changelog file. The process cannot be fully automated because it would have to modify the files in the repository, which has lots of problems.

In general, we would run the script as part of the release process to ensure that the counts are accurate; but it can also be run before then to update the current contributor list.

samukweku:
I read that part in the docs, and it was not too clear. I surmised then that the script does the update of the contributors. I suggest, if that is ok with you, that that section be made clearer, so contributors know that they don’t need to include their names in the contributor section, as that would be covered by the project maintainers via the script

docs/manual/groupby_examples.rst
```diff
@@ -0,0 +1,755 @@
...
+                01-02   TMAX    71
+                01-02   TMIN    26"""
+
+    # Task : group by Day and find minimum Data_Value for TMIN 
+    # and maximum Data_Value for TMAX
+
+    df[:,
+       f.Day.extend({"TMAX" : ifelse(f.Element=="TMAX", f.Data_Value, None),
+                     "TMIN" : ifelse(f.Element=="TMIN", f.Data_Value, None)})
+      ][:, [max(f.TMAX), min(f.TMIN)], by('Day')]
```
st-pasha, line 642: You can combine these two expressions into a single one:
```
df[:, {"TMAX": max(ifelse(f.Element=="TMAX", f.Data_Value, None)),
       "TMIN": min(ifelse(f.Element=="TMIN", f.Data_Value, None))},
   by(f.Day)]
```

samukweku:
Thanks for the feedback @st-pasha , I"ll make the edits and push

st-pasha:
Thanks Samuel, looks great now.

Issue #2358, opened 2020-02-21
Implement a linear model with stochastic gradient descent learning


oleksiyskononenko:
A linear model we are going to add to `datatable.models` should support binomial and multinomial classifications, as well as numeric regression. It should be a regularized model with stochastic gradient decent (SGD) learning similar to the implementation provided by [scikit-learn](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.linear_model).

The model should support all the column types that are currently supported by `datatable`, i.e. numeric and string columns, and will include 

**I. The preparation step: feature binning**
- for string columns we will do one hot encoding, it should also be possible to specify category groups for a particular column;
- for numeric columns we will do binning first, the number of bins or the corresponding intervals will be an input to the model. Second, we will do one hot encoding based on the binned data.
These preparation steps will ensure the model has high interpretability. 

**II. Training the model**
Each row of data will be converted on-the-fly to a set of zeros and ones — an input to our model. Then, we will train the model by using the SGD method. Its API should be similar to the datatable FTRL API: regularization parameters, learning rate, precision, etc. but will exclude parameters related to the hashing trick.

XiaomoWu:
First, let me be clear that I'm more than happy to see new features added to `pydatatable`. However, do you mind talking more about why you consider linear models to be necessary for `pydatatable`, which is primarily a data manipulation package instead of an ML package? I'm just curious about which direction will `pydatatable` take in the future: a dedicated 2D data manipulation tool, or a combination of data manipulation and ML tool? At last, thanks again for bringing the amazing `Rdatatable` to python!

oleksiyskononenko:
Dear @XiaomoWu, yes, the main purpose of datatable is to be a dedicated 2D data manipulation tool. However, we also have a limited number of models under the datatable.models, namely, data aggregator, the FTRL model and the k-fold splitter.

For the moment we are not going to make datatable.models as complete as possible in terms of the ML tools. This is because, first, there are already quite a lot of excellent tools available to the community; second, we do not have enough resources and see no reason to bring all of the existing algos under the same umbrella. 

[...]

PR #2535, opened 2020-07-11
Implement cut() function

### LINKED ISSUE #2483: Implement equivalent of Pandas cut()
1D aggregator does a similar job, but it would be nice to have [cut()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.cut.html) as a frame or frame-proxy method along with `qcut()` requested on this issue https://github.com/h2oai/datatable/issues/1680

oleksiyskononenko:
In this PR we add a `cut(cols, nbins=10, right_closed=True)` function that can be applied to a Frame or to an f-expression. The function behavior is very close to the following Pandas code: `pd.cut(input, bins, right, labels=False)`.

Also, some improvements to the expression documentation and minor fixes to `src/datatable/__init__.py` and `src/datatable/sphinxext/xfunction.py`.

WIP for #2358 
Closes #2483


src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+  * Virtual column for data binning.
+  */
+template <typename T_elems, typename T_stats>
+class Cut_ColumnImpl : public Virtual_ColumnImpl {
+  private:
+    Column col_dbl_;
+    double bin_factor_, bin_shift_;
+
+  public:
+    Cut_ColumnImpl(Column col, size_t bins)
```
st-pasha, line 42: I think in all virtual columns we pass children columns via move: `Column&& col`, so that no extra copy is created when not needed.

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+    {
+      col_dbl_ = Column(new dt::FuncUnary2_ColumnImpl<T_elems, double>(
+                   Column(col),
+                   [](T_elems x, bool x_isvalid, double* out) {
+                     *out = static_cast<double>(x);
+                     return x_isvalid && _isfinite(x);
+                   },
+                   col.nrows(),
+                   dt::SType::FLOAT64
+                 ));
```
st-pasha, line 55: Looks like you're trying to convert the column into float64 type here -- it's easier (and potentially more efficient) to use `col.cast_inplace(SType::FlOAT64)`.

I know you're converting infinities to NAs here too, but that cost should not be incurred unless actually needed, because columns with infinities are extremely rare. So, the simplest approach would be to ignore them completely; while a more advanced one is to proceed as if there are no infinities, but if min/max operations below return infinities, then treat this as a special case.

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+#include "column/func_unary.h"
+#include "column/virtual.h"
+#include "stype.h"
+
+namespace dt {
+
+
+/**
+  * Virtual column for data binning.
+  */
```
st-pasha, line 34: Maybe explain a little more, what kind of binning, and specifically how the edge cases are treated.

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+template <typename T_elems, typename T_stats>
+class Cut_ColumnImpl : public Virtual_ColumnImpl {
+  private:
+    Column col_dbl_;
+    double bin_factor_, bin_shift_;
+
+  public:
+    Cut_ColumnImpl(Column col, size_t bins)
+      : Virtual_ColumnImpl(col.nrows(), col.stype()),
```
st-pasha, line 43: This declares that your column has `col.nrows()` rows (ok), and stype `col.stype()`, which is not ok because it seems that you wanted to create a column of INT32 stype

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+                   col.nrows(),
+                   dt::SType::FLOAT64
+                 ));
+
+      // Get stats
+      T_stats tmin, tmax;
+      col.stats()->get_stat(Stat::Min, &tmin);
+      col.stats()->get_stat(Stat::Max, &tmax);
+      double min = static_cast<double>(tmin);
+      double max = static_cast<double>(tmax);
```
st-pasha, line 62: In theory we should implement conversion of (some) stats during type-cast.
However, in this case you can get the min/max stats of `col_dbl_` (i.e. after the type cast), so that you can avoid your class being dependent on `T_stats`: `col_dbl_.stats()->min_double(&isvalid)`.

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+        col_dbl_(col),
+        bin_factor_(bin_factor),
+        bin_shift_(bin_shift) {}
+
+    ColumnImpl* clone() const override {
+      return new Cut_ColumnImpl<T_elems, T_stats>(col_dbl_, bin_factor_, bin_shift_);
+    }
+
+    bool allow_parallel_access() const override {
+      return col_dbl_.allow_parallel_access();
```
st-pasha, line 79: This override is not needed: the base implementation will use `n_children()` / `child()` to get the same outcome.

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+      double value;
+      bool is_valid = col_dbl_.get_element(i, &value) && _notnan(bin_factor_);
+
+      if (is_valid) {
+        *out = static_cast<int32_t>(bin_factor_ * value + bin_shift_);
+      } else {
+        return false;
+      }
+
+      return true;
```
st-pasha, line 102: Because NAs are rare, and it is allowed to write garbage into `out` in case of NAs, a simpler approach would be
```
// no ifs and buts here!
*out = static_cast<int32_t>(bin_factor_ * value + bin_shift_);
return is_valid;
```

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+      // Get stats
+      T_stats tmin, tmax;
+      col.stats()->get_stat(Stat::Min, &tmin);
+      col.stats()->get_stat(Stat::Max, &tmax);
+      double min = static_cast<double>(tmin);
+      double max = static_cast<double>(tmax);
+
+      // Set up binning coefficients
+      set_cut_coeffs(bin_factor_, bin_shift_, min, max, bins);
+    }
```
st-pasha, line 66: I would remove this constructor entirely, replacing it with a static factory function `static ColumnImpl* make(Column&&, size_t nbins)`. Then after computing all the coefficients you can use the constructor below to return the proper instance. Unless you have one of the special cases: infinities, or `bin_factor_` ended up being NA. Because if bin_factor_ is NA, then `get_element()` will produce NA values only, and in such a case it's much easier to return a `ConstNa_ColumnImpl` instance.

src/core/column/cut.h
```diff
@@ -0,0 +1,131 @@
...
+
+    const Column& child(size_t i) const override {
+      xassert(i == 0);  (void)i;
+      return col_dbl_;
+    }
+
+
+    bool get_element(size_t i, int32_t* out)  const override {
+      double value;
+      bool is_valid = col_dbl_.get_element(i, &value) && _notnan(bin_factor_);
```
st-pasha, line 94: `_notnan()` check will not be needed if you always return a ConstNA_ColumnImpl when bin_factor_ is NA.

st-pasha:
Note that after the proposed changes your virtual column won't need to be a template, which is much easier to handle!

src/core/column/cut.h
oleksiyskononenko: In this case I would need to cast to float64 first and then calculate stats. I guess there is nothing wrong if we keep the template and calculate stats on the original column (it could be faster for some datatypes, and if min/max are inf/nan we won't really need to cast at all — just return the `ConstNa_ColumnImpl`).

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+  if (args[0].is_none_or_undefined()) {
+    throw TypeError() << "Function `cut()` requires one positional argument, "
+                         "but none were given";
+  }
+  oobj arg0 = args[0].to_oobj();
+  oobj arg1 = args[1].is_none_or_undefined()? py::None() : args[1].to_oobj();
+  oobj arg2 = args[2].is_none_or_undefined()? py::True() : args[2].to_oobj();
+
+  if (arg0.is_frame()) {
+    return cut_frame(arg0, arg1, arg2);
```
st-pasha, line 193: what if we had a generic method that takes `Expr(op, (DT,), params)` and converts it into `DT[:, Expr(op, (f[:],), params)]`.

tests/ijby/test-cut.py
```diff
@@ -0,0 +1,213 @@
...
+	DT_cut_tuple = cut(DT, nbins = tuple(nbins))
+	DT_cut_list_left = cut(DT, nbins = nbins, right_closed = False)
+	assert_equals(DT_ref_right, DT_cut_list)
+	assert_equals(DT_ref_right, DT_cut_tuple)
+	assert_equals(DT_ref_left, DT_cut_list_left)
+
+
+
+@pytest.mark.xfail(reason="This test may fail rarely due to pandas "
+                   "inconsistency, see test_cut_pandas_issue_35126 test")
```
st-pasha, line 165: Usually we mark `xfail` test that is known to fail right now, but there is an expectation that the functionality will be fixed in the future. In this case, is there an expectation that this test will ever work? I mean, did pandas acknowledge they are going to fix their issue?

tests/ijby/test-cut.py
```diff
@@ -0,0 +1,213 @@
...
+def test_cut_error_groupby():
+    msg = r"cut\(\) cannot be used in a groupby context"
+    DT = dt.Frame(range(10))
+    with pytest.raises(NotImplementedError, match=msg):
+        DT[:, cut(f[0]), f[0]]
+
+
+def test_cut_empty_frame():
+	DT_cut = cut(dt.Frame())
+	assert_equals(DT_cut, dt.Frame())
```
st-pasha, line 85: Looks like the files uses tabs mixed with spaces. There is "convert indentation to spaces" function in Sublime.

src/core/column/cut.h
```diff
@@ -0,0 +1,170 @@
...
+
+    bool get_element(size_t i, int32_t* out)  const override {
+      double value;
+      bool is_valid = col_.get_element(i, &value);
+      *out = static_cast<int32_t>(a_ * value + b_) + shift_;
+      return is_valid;
+    }
+
+
+    static void set_cut_coeffs(double& a, double& b, int32_t& shift,
```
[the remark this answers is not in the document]
oleksiyskononenko: Yeah, compute should be better, I guess I called it set because it sets `a`, `b` and `shift` that are passed by reference.

tests/ijby/test-cut.py
oleksiyskononenko: No, they didn't answer. We can disable this particular test for the moment, but I would like to keep it, because it fails very rare and at the same time helps to identify bugs if we change something in future.

tests/ijby/test-cut.py
oleksiyskononenko: Yeah, I had tabsize equal two in this file for some reason :/

tests/ijby/test-cut.py
st-pasha: Yeah, so that's my concern: how would we notice that something went wrong if the test is marked as "it's ok for something to go wrong"?

src/core/column/cut.h
```diff
@@ -0,0 +1,170 @@
...
+ *       Then, multiply the normalized values by the number of the requested
+ *       bins and add a proper shift to compute the final bin ids. The following
+ *       formula is used for that (note, casting to an integer will truncate
+ *       toward zero)
+ *
+ *         bin_id_i = static_cast<int32_t>(x_i * a + b) + shift
+ *
+ *       2.1) if `max == min`, all values end up in the central bin which id
+ *            is determined based on the `right_closed` parameter being
+ *            `true` or `false`, i.e.
```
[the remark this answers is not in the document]
oleksiyskononenko: Yeah, should work, let me simplify it here.

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+    Frame or f-expression consisting of numeric columns.
+nbins: int | list of ints | tuple of ints
+    When a single number is specified, this number of bins
+    will be used to bin each column of `cols`.
+    When a list or a tuple is provided, each column will be binned
+    by using its own number of bins. In the latter case,
+    the list/tuple length must be equal to the number of columns
+    in `cols`.
+right_closed: bool
+    Indicates whether bins should include the rightmost edge or not.
```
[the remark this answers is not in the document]
oleksiyskononenko: Are you sure hyperlinks will work from the `.cc` file? I'm getting the following warning and they're not rendered properly 

```
datatable/docs/api/dt/cut.rst:24: WARNING: Inline literal start-string without end-string.
```

src/core/column/cut.h
```diff
@@ -0,0 +1,170 @@
...
+ *              a = (1 - epsilon) * nbins / (max - min)
+ *              b = -a * min + (epsilon - 1) * nbins
+ *              shift = nbins - 1;
+ *
+ *            scaling data to [epsilon - 1; 0], multiplying them by `nbins`,
+ *            and then shifting the resulting values by `nbins - 1`. The final
+ *            shift converts the auxiliary negative bin ids to the corresponding
+ *            positive bin ids.
+ *
+ */
```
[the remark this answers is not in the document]
oleksiyskononenko: Hm, It seems that we now have two indentation formats in use. Both work fine on my end. When did we change from one space to two?

src/core/column/cut.h
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+
+    static void set_cut_coeffs(double& a, double& b, int32_t& shift,
+                               const double min, const double max,
+                               const size_t nbins, const bool right_closed)
+    {
+      const double epsilon = static_cast<double>(
+                               std::numeric_limits<float>::epsilon()
+                             );
```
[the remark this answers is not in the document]
oleksiyskononenko: Double-precision epsilon doesn't always work, it seems that this value is too small and after all the all the manipulations with the bin ids the result is identical as if I use `epsilon == 0`.

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+
+    size_t i = 0;
+    for (auto py_nbin : py_nbins) {
+      size_t nbin = py_nbin.to_size_t();
+      nbins[i++] = nbin;
+    }
+    xassert(i == ncols);
+
+  } else {
+    if (defined_nbins) nbins_default = py_nbins_.to_size_t();
```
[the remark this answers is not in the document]
oleksiyskononenko: For the moment, when `nbins == 0` I just return all bins as `NA` and don't throw. Yes, we can do `to_int32_strict()` here.

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+      case dt::SType::INT64:   coli = Column(Cut_ColumnImpl::make<int64_t>(std::move(coli), nbins[i], right_closed_));
+                               break;
+      case dt::SType::FLOAT32: coli = Column(Cut_ColumnImpl::make<double>(std::move(coli), nbins[i], right_closed_));
+                               break;
+      case dt::SType::FLOAT64: coli = Column(Cut_ColumnImpl::make<double>(std::move(coli), nbins[i], right_closed_));
+                               break;
+      default:  throw TypeError() << "cut() can only be applied to numeric "
+                  << "columns, instead column `" << i << "` has an stype: `"
+                  << coli.stype() << "`";
+    }
```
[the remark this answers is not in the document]
oleksiyskononenko: Yeah, I had an ltype check at some point, but then forgot to restore it. Thanks!

src/datatable/models.py
```diff
@@ -21,8 +21,8 @@
 # IN THE SOFTWARE.
 #-------------------------------------------------------------------------------
 from datatable.lib._datatable import (
-	Ftrl,
 	aggregate,
+	Ftrl,
 	kfold,
 	kfold_random,
 )
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, and also tabsize was two. Will fix.

tests/ijby/test-cut.py
oleksiyskononenko: No way for the moment, my intention is to just keep this test somewhere and enable it when I do some related development. I can mark it with `skip`.

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+  {
+    "cols", "nbins", "right_closed"
+  },
+  "cut",
+
+R"(cut(cols, nbins=10, right_closed=True)
+--
+
+Cut all the columns in a Frame/f-expression by binning
+their values into equal width discrete intervals.
```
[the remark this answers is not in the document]
oleksiyskononenko: I see both usages, not sure which one is correct.

tests/ijby/test-cut.py
st-pasha: ok, whatever you think is best

oleksiyskononenko: Ok, I will mark it with `skip` so that it doesn't spam with all the xfail messages.

src/core/column/cut.h
```diff
@@ -0,0 +1,170 @@
...
+ *              shift = 0
+ *
+ *            scaling data to [0; 1 - epsilon] and then multiplying them
+ *            by `nbins`.
+ *
+ *       2.3) When `min != max`, and `right_closed == false`, set
+ *
+ *              a = (1 - epsilon) * nbins / (max - min)
+ *              b = -a * min + (epsilon - 1) * nbins
+ *              shift = nbins - 1;
```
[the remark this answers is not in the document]
oleksiyskononenko: There is also a case when `min == max` which should be treated separately...

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+----------
+cols: Frame | f-expression
+    Frame or f-expression consisting of numeric columns.
+nbins: int | list of ints | tuple of ints
+    When a single number is specified, this number of bins
+    will be used to bin each column of `cols`.
+    When a list or a tuple is provided, each column will be binned
+    by using its own number of bins. In the latter case,
+    the list/tuple length must be equal to the number of columns
+    in `cols`.
```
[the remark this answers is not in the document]
oleksiyskononenko: Pandas only supports 1D array-like input, while we support frame/expressions. It means that we may want to use different `nbins` depending on a column. Also, Pandas is mixing two things together `nbins` and `bins`, and I propose we have separate parameters for that.

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,214 @@
...
+  */
+static oobj pyfn_cut(const PKArgs& args)
+{
+  if (args[0].is_none_or_undefined()) {
+    throw TypeError() << "Function `cut()` requires one positional argument, "
+                         "but none were given";
+  }
+  oobj arg0 = args[0].to_oobj();
+  oobj arg1 = args[1].is_none_or_undefined()? py::None() : args[1].to_oobj();
+  oobj arg2 = args[2].is_none_or_undefined()? py::True() : args[2].to_oobj();
```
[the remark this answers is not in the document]
oleksiyskononenko: It doesn't seem we have this method for `py::Arg`

src/core/expr/head_func_cut.cc
```diff
@@ -0,0 +1,221 @@
...
+--------
+:func:`qcut()` -- function for quantile binning.
+
+.. _`half-open`: https://en.wikipedia.org/wiki/Interval_(mathematics)#Terminology
+
+)";
+
+
+static PKArgs args_cut(
+  1, 2, 0, false, false,
```
st-pasha, line 177: Maybe this should be `1, 0, 2, ...` (or at least `1, 1, 1`): require arguments `nbins` and `right_closed` to be explicitly named. Consider call `cut(f.A, 10)` -- this seems ok, cut into 10 bins. But `cut(f.A, 10, True)` is already very confusing. Plus, once we have the `bins` argument as well, then it would be unclear why we require `bins=` to be explicitly named while allowing `nbins=` to stay unnamed.

tests/ijby/test-cut.py
```diff
@@ -0,0 +1,209 @@
...
+    msg = r"cut\(\) can only be applied to numeric columns, instead column 1 has an stype: str32"
+    with pytest.raises(TypeError, match=msg):
+        cut(DT)
+
+
+def test_cut_error_negative_bins():
+    msg = "Number of bins must be positive, instead got: -10"
+    DT = dt.Frame(range(10))
+    with pytest.raises(ValueError, match=msg):
+        cut(DT, -10)
```
st-pasha, line 51: Also: add test for 0 bins

src/core/expr/head_func_cut.cc
oleksiyskononenko: Sure, let's do it `1, 0, 2`.

tests/ijby/test-cut.py
oleksiyskononenko: I guess I had it, but it got somehow removed in one of the re-factoring iterations. WIll add some more.

PR #2536, opened 2020-07-15
Auto-generate links to artifacts published on S3

### LINKED ISSUE #952: Add download links for datatable

st-pasha:
When publishing artifacts to S3, we will now generate links to those artifacts and put those links into https://h2o-release.s3.amazonaws.com/datatable/index.html. This file acts as a simple python repository, allowing one to install one of the latest dev wheels. 

Closes #952

### DIFF
--- a/ci/Jenkinsfile.groovy
+++ b/ci/Jenkinsfile.groovy
@@ -581,8 +581,7 @@ ansiColor('xterm') {
                             def s3cmd = ""
                             def pyindex_links = ""
                             dir("dist") {
-                                def distDir = new File(".")
-                                distDir.traverse {
+                                findFiles(glob: "*").each {
                                     def sha256 = sh(script: """python -c "import hashlib;print(hashlib.sha256(open('${it.name}', 'rb').read()).hexdigest())" """,
                                                     returnStdout: true).trim()
                                     def s3path = "${S3_BASE}/dev/datatable-${versionText}/{it.name}"
@@ -598,8 +597,7 @@ ansiColor('xterm') {
                             pyindex_content += pyindex_links
                             pyindex_content += "</ul>\n"
                             pyindex_content += "</body></html>\n"
-                            File pyindex_file = new File("index.html")
-                            pyindex_file.write(pyindex_content)
+                            writeFile(file: "index.html", text: pyindex_content)
                             s3cmd += "s3cmd put -P index.html ${S3_BASE}/index.html"
 
                             docker.withRegistry("https://harbor.h2o.ai", "harbor.h2o.ai") {
--- a/ci/Jenkinsfile.groovy
+++ b/ci/Jenkinsfile.groovy
@@ -42,10 +42,13 @@ NODE_MACOS   = 'osx'
 NODE_PPC     = 'ibm-power'
 NODE_RELEASE = 'docker && linux && !micro'
 
+// Paths for S3 artifacts
+S3_BASE = "s3://h2o-release/datatable"
+S3_URL = "https://h2o-release.s3.amazonaws.com/datatable"
 
 // Paths should be absolute
 S3_URL_STABLE = "s3://h2o-release/datatable/stable"
-HTTPS_URL_STABLE = "https://h2o-release.s3.amazonaws.com/datatable/stable"
+
 // Data map for linking into container
 LINK_MAP = [
     "Data"     : "h2oai-benchmarks/Data",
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile.groovy
```diff
@@ -569,25 +572,49 @@ ansiColor('xterm') {
...
+                            docker.withRegistry("https://harbor.h2o.ai", "harbor.h2o.ai") {
+                                withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: "awsArtifactsUploader"]]) {
+                                    docker.image("harbor.h2o.ai/library/s3cmd").inside {
+                                        sh s3cmd
+                                    }
+                                }
+                            }
+
+                            // TODO: remove?
                             s3upDocker {
```
mmalohlava, line 614: regarding S3UpDocker:

The strategy which we are using in other projects:
1. prepare structure which we would like to see in the target bucket locally. E.g.:
```
dist/
dist/project.whl
dist/docs
dist/docs/index.html
dist/index.html
```
2. s3cmddocker is doing simple sync into target location defined by release/version/artifactid

mmalohlava: the lesson learned from that that we minimize code in Jenkinsfiles (which are hard to debug) and simply prepare the structure locally (which also enable some testing if necessary)

st-pasha: I understand benefits of code sharing, but the problems with `s3upDocker` are:
- It is impossible to know (from reading the code) where the artifacts will end up. There is literally no S3 url in its parameters;
- It alters the paths of the files that it receives, splitting them into sub-directories. This is problematic because the index.html file needs to end up at the exact path, while other artifacts should be at some paths that are known beforehand;
- The function has neither documentation nor source code (at least GitHub cannot locate the source code: https://github.com/search?q=s3updocker&type=Code). It might be the permission issue.

At the same time, the alternative -- to construct the `s3cmd` command manually -- is perfectly simple, straightforward and understandable. The only code that would be useful is to reuse would be a function like
```
def s3docker(cmd) {
  docker.withRegistry("https://harbor.h2o.ai", "harbor.h2o.ai") {
    withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: "awsArtifactsUploader"]]) {
      docker.image("harbor.h2o.ai/library/s3cmd").inside {
        sh cmd
      }
    }
  }
}
```

oleksiyskononenko:
Several questions:
1. Do we really need to add links for every PR and not just for the master builds?
2. How big will this file be in a week, month, year? I guess eventually it will be time consuming to load it in a browser. 
3. Rather than on s3, why not to post these links on our docs page?

st-pasha:
> Do we really need to add links for every PR build and not just for the master builds?

We only add links to artifacts that get published to S3, which are typically only master builds, but also those that get built with `FORCE_S3_PUSH` parameter.

> How big will index.html be in a week, month, year? I guess eventually it will be time consuming to load it in a browser.

It adds about 3Kb of code per commit. If we started this thing from the beginning of the project, there would be about 5Mb of code by now. Plus, if we ever feel like the file became too big, we could manually remove some of the older versions.

> Rather than posting it on s3, why not having these links posted on our docs page?

Because doc pages live on GitHub, and thus in order to "post" anything there, you'd have to make a new commit into the master branch. And making a new commit would trigger a new master build...

oleksiyskononenko:
>We only add links to artifacts that get published to S3, which are typically only master builds, but also those that get built with FORCE_S3_PUSH parameter.

That's great.

>It adds about 3Kb of code per commit. If we started this thing from the beginning of the project, there would be about 5Mb of code by now. Plus, if we ever feel like the file became too big, we could manually remove some of the older versions.

I still think this file will once become pretty big and won't be easy for navigation. May be we at least add the current date next to the build?

>Because doc pages live on GitHub, and thus in order to "post" anything there, you'd have to make a new commit into the master branch. And making a new commit would trigger a new master build...

I've been thinking that Jenkins could make an additional commit to the current branch, but it will probably become cumbersome.

PR #2538, opened 2020-07-16
Update installation instructions

st-pasha:
- Installation instructions improved:
  - Add `pip` as a dependency;
  - Describe platform requirements;
  - More details on how to install dev wheels;
  - Expanded troubleshooting section.

- Improvements to `..xcode` directive:
  - Disable processing of code from `smartquotes` plugin;
  - Make sure prompt includes the next space (so that it is not selectable too);
  - Update styles.


docs/start/install.rst
```diff
@@ -1,129 +1,279 @@
...
+tab, copy the wheel URL that corresponds to your Python version, and finally
+install it as:
 
 .. xcode:: winshell
 
-   C:\> pip install YOUR_WHEEL_URL
+    C:\> pip install YOUR_WHEEL_URL
 
+For macOS and Linux, development wheels can be found at our `S3 repository`_.
+Scroll to the bottom of the page to find the latest links, and then download
```
oleksiyskononenko, line 92: Have you been thinking about adding builds to the `index.html` file in a reverse order? Then one won't need to scroll all the time.

docs/start/install.rst
```diff
@@ -1,129 +1,279 @@
...
 .. xcode:: shell
 
    $ pip install git+https://github.com/h2oai/datatable
 
-Since ``datatable`` is written mostly in C++, you will need to have a C++
-compiler on your computer. We recommend either `Clang 4+`, or `gcc 6+`,
-however in theory any compiler that supports C++14 should work.
+Since ``datatable`` is written mostly in C++, your computer must be set up for
+compiling C++ code. The build script will attempt to find the compiler
+automatically, searching for GCC, Clang, or MSVC on Windows. It it fails, or
```
oleksiyskononenko, line 121: `It it fails` -> `If it fails`

oleksiyskononenko:
While we're on updating instructions, what about also updating https://github.com/h2oai/datatable/blob/master/src/core/expr/!readme.md adding a note that the new function should also be added to `__init__.py` or to the proper `module_name.py` file?

docs/start/install.rst
st-pasha: It's much easier (and more reliable) to append text at the bottom of the page, than insert somewhere in the middle. Plus, it's not like this is such a common way of installation anyways...

st-pasha:
> what about also updating https://github.com/h2oai/datatable/blob/master/src/core/expr/!readme.md 

I thought you were going to update that readme as part of your cut() PR.

oleksiyskononenko:
Sure, I can push it to my PR.

PR #2545, opened 2020-07-23
Fix saving an Rbound string column to Jay

### LINKED ISSUE #2543: Corrupted Frame created when reading m_r10000.csv
```pycon
>>> import datatable as dt
>>> DT = dt.fread('~/Downloads/m_r10000.csv')
>>> dt.internal.frame_integrity_check(DT)

>>> DT[0].materialize()
AssertionError: Assertion '!newimpl->stats_' failed in src/core/column.cc, line 128

>>> dt.fread(DT[0].to_jay())
AssertionError: Assertion 'mb.size() >= sizeof(T) * (n + 1)' failed in src/core/column/sentinel_str.cc, line 62
```
where the dataset is https://h2oai.slack.com/files/U03QXBZCM/F017VPTGJ0Z/m_r10000.csv

_Originally posted by @pseudotensor at https://h2oai.slack.com/archives/C0BECF81M/p1595377527493100_

st-pasha:
When an rbound string column was saved to Jay, the offsets were not properly concatenated, creating a corrupted column.

Closes #2543

### DIFF
--- a/src/core/column.cc
+++ b/src/core/column.cc
@@ -125,8 +125,10 @@ dt::ColumnImpl* Column::_get_mutable_impl(bool keep_stats) {
   xassert(impl_);
   if (impl_->refcount_ > 1) {
     auto newimpl = impl_->clone();
-    xassert(!newimpl->stats_);
-    if (keep_stats && impl_->stats_) {
+    if (newimpl->stats_) {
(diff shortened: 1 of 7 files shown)


oleksiyskononenko:
May be we also add a test where we explicitly `rbind` two frames and then save the result to `jay`?  Because the internal logic of `fread` could change at some point.

PR #2553, opened 2020-07-24
Make ..xfunction compatible with Sphinx 3

st-pasha:
Use `PythonDomain.note_object()` instead of manipulating domain's data explicitly.
As a side-effect, this change is incompatible with Sphinx prior to version 3.0 (there was no `node_id=` back then). So, whoever works on documentation will need to upgrade their Sphinx installation.


st-pasha:
cc @samukweku  After this PR, Sphinx 3.0+ will be required for building docs.

PR #2555, opened 2020-07-29
Refactor parallel ordered for loop

st-pasha:
A parallel for-ordered loop will now require an explicit class that implements methods start()/order()/finish(), instead of the current approach with multiple nested lambdas.

The reason for this change is because the previous approach required instantiating thread contexts by recursively calling the lambda, which not only was conceptually difficult to understand, but also produces stack traces that are hard to debug (partly because of the many levels of nestedness, partly because the thread state was kept in auto-generated variables which are hard to inspect). Lastly, if there are too many threads (ppc64 may have 100+), there is a danger of causing a stack overflow.


st-pasha:
Merging, because the only failure is a timeout in a debug build when running large fread tests (and it wasn't an actuall stall or anything, the tests were running, just slowly...)

PR #2559, opened 2020-08-05
Implement qcut() function

### LINKED ISSUE #1680: Implement equivalent of Pandas qcut()
- Describe what feature you would like to see implemented.

> Pandas equivalent of `qcut()`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.qcut.html

- If possible, give an example of how it may look in the code and what result
  will be produced.

> Similar to Pandas `qcut()`. Maybe some improvements involved API wise but unsure of those at the moment.

oleksiyskononenko:
In this PR we add a `qcut(cols, nquantiles=10)` function that can be applied to a Frame or to an f-expression. The function behavior is very close to the following Pandas code: `pd.qcut(input, q, labels=False)`.

WIP for #2358 
Closes #1680


src/core/column/qcut.h
```diff
@@ -0,0 +1,178 @@
...
+    Column col_;
+    RowIndex ri_;
+    Groupby gb_;
+    int32_t nquantiles_;
+    size_t : 32;
+
+  public:
+    static ColumnImpl* make(Column&& col, int32_t nquantiles) {
+      xassert(nquantiles > 0);
+      auto res = group({col}, {SortFlag::NONE});
```
st-pasha, line 68: sorting/grouping is a quite expensive operation, so it would be better to hide inside the `Qcut_ColumnImpl`, so that it would get materialized only when it's actually needed.

src/core/column/qcut.h
```diff
@@ -0,0 +1,178 @@
...
+
+      return new Latent_ColumnImpl(new Qcut_ColumnImpl(
+                   std::move(col), nquantiles, std::move(ri), std::move(gb)
+                 ));
+
+    }
+
+
+    void materialize(Column& col_out, bool) override {
+      Column col_out_ = Column::new_data_column(col_.nrows(), SType::INT32);
```
st-pasha, line 93: By convention, a variable ending with `_` is a class member, in order to distinguish them from local variables. So, maybe call this `newcol` (or anything without an underscore at the end)?

src/core/column/qcut.h
```diff
@@ -0,0 +1,178 @@
...
+      auto data_out_ = static_cast<int32_t*>(col_out_.get_data_editable());
```
st-pasha, line 94: (same here)

src/core/column/qcut.h
```diff
@@ -0,0 +1,178 @@
...
+      {
+        size_t row;
+        bool row_valid = ri_.get_element(0, &row);
+        xassert(row_valid); (void) row_valid;
+        has_na_group = !col_.get_element_isvalid(row);
+      }
+
+      // Set up number of valid groups and the quantile coefficient.
+      size_t ngroups = gb_.size() - has_na_group;
+      double a = nquantiles_ * (1 - epsilon) / (ngroups - 1);
```
st-pasha, line 111: what if `ngroups` is 1 (after removing the na group)?

src/core/column/qcut.h
oleksiyskononenko: Yes, but then we have this problem returning a virtual column from the `Latent_ColumnImpl`.

src/core/column/qcut.h
oleksiyskononenko: Sure.

src/core/column/qcut.h
oleksiyskononenko: Right, will handle this case separately.

src/core/column/qcut.h
oleksiyskononenko: Yep.

src/core/column/qcut.h
st-pasha: You could always just do `.materialize()` on that column before returning.
Plus, we discussed yesterday how to enable virtual columns with Latent column: create a special "placement clone" virtual method.

oleksiyskononenko: Yes, I first had materialization turned on for these columns, but didn't like that solution. May be let me revert it back, as it is still more "optimal" than doing a groupby.

As for the Latent column improvements, I propose we do it in a separate PR along with some other changes to `materialize` API.

src/core/column/qcut.h
```diff
@@ -0,0 +1,172 @@
...
+
+          gb.get_group(i, &j0, &j1);
+
+          for (size_t j = j0; j < j1; ++j) {
+            size_t row;
+            bool row_valid = ri.get_element(j, &row);
+            xassert(row_valid); (void) row_valid;
+            data_tmp[row] = q;
+          }
+      });
```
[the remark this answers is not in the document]
oleksiyskononenko: Yes, I first even had it enabled for string columns but then thought if this could be in any way useful... not sure for the moment.

src/core/expr/head_func_qcut.cc
```diff
@@ -0,0 +1,196 @@
...
+R"(qcut(cols, nquantiles=10)
+--
+
+Bin all the columns in a Frame/f-expression into equal-population
+discrete intervals, i.e. quantiles. In reality, for some data
+these quantiles may not have exactly the same population.
+
+Parameters
+----------
+cols: Frame | f-expression
```
st-pasha, line 144: I guess this should be saying just `f-expression` now, with the rest of the doc updated accordingly

oleksiyskononenko: Why? This is an input parameter, that can also be a `Frame`.

oleksiyskononenko: I guess this one should be changed to `Expr` instead

```
return: Frame | Expr
    Frame/f-expression, where each column is filled with
    the respective quantile ids.
```

st-pasha: The argument can actually be either an FExpr, or Frame, or pandas DataFrame/Series, or numpy array, or a python range, or a simple number, etc. In reality we will take the argument and attempt to convert into an FExpr, which works for many different types of values (and the set of types is going to only expand in the future). The resulting FExpr must also implement the `evaluate_n()` method, and most of them do but some don't (like `Head_Literal_Type`).

st-pasha: So, the CORRECT way would be to come up with a new term (like `FExpr-like` or `FExpr-able` or I don't know), and then create a separate help page to describe what that term actually means, then make it so that this type would automatically link to its corresponding help page in Sphinx.

A shorter way would be to simply write the new term here, and then leave it unexplained for now, creating a new ticket for finishing the documentation part.

oleksiyskononenko: Don't we already have any function that support the same input type?

PR #2560, opened 2020-08-05
Encode unsafe characters in URLs passed to fread

st-pasha:
If the user passes a URL with unsafe characters (such as spaces) to fread, then such URLs will now be read correctly. For example:
```
import datatable as dt
dt.fread("https://github.com/OpportunityInsights/EconomicTracker/raw/main/data/Google Mobility - County - Daily.csv")
```

### DIFF
--- a/src/core/read/source.cc
+++ b/src/core/read/source.cc
@@ -209,8 +209,24 @@ py::oobj Source_Url::read(GenericReader& reader) {
   {
     dt::progress::work job(1);
     job.set_message("Downloading " + url_);
+    // Characters [0-9a-zA-Z._~-] are always considered "safe".
+    // Further, "/" is typically used as a path separator, and ":"
+    // is in the protocol. Both can be used in URLs safely.
+    // The "%" character, on the other hand, is not "safe" -- it is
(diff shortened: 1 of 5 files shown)


src/core/read/source.cc
```diff
@@ -209,8 +209,24 @@ py::oobj Source_Url::read(GenericReader& reader) {
...
+    // Characters [0-9a-zA-Z._~-] are always considered "safe".
+    // Further, "/" is typically used as a path separator, and ":"
+    // is in the protocol. Both can be used in URLs safely.
+    // The "%" character, on the other hand, is not "safe" -- it is
+    // used to indicate an escape character. However, we designate it
+    // as "safe" so that the user may be able to pass a URL that is
+    // already correctly encoded. The downside is that if the url
+    // is not encoded, but happens to have a % character in it, then
+    // this function will throw an exception, but I consider such
+    // cases to be very rare. If they do occur, then the user could
```
oleksiyskononenko, line 221: Yeah, normally URLs are copy-pasted from a browser, that does encoding on-the-fly.

PR #2565, opened 2020-08-12
Fix warnings produced by Clang 10

st-pasha:
After switching to the recent version of Clang compiler, there were over 1000 warnings, mostly from the `-Wimplicit-int-float-conversion` category. These warning are now fixed and the source compiles cleanly with the recent Clang.

### DIFF
--- a/ci/ext.py
+++ b/ci/ext.py
@@ -342,7 +342,6 @@ def build_extension(cmd, verbosity=3):
                 "-Wno-switch-enum",
                 "-Wno-weak-template-vtables",
                 "-Wno-weak-vtables",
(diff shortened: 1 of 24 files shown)


oleksiyskononenko:
Don't you want to disable it with `-Wno-implicit-int-float-conversion` when compiler is `Clang10`?

st-pasha:
Well, the problem is that there is no way for "ext.py" to know the version of Clang reliably. We could in theory run `clang --version`, but Apple's Clang versions do not correspond to actual Clang version; plus there is no guarantee that we won't have to do the same for GCC in the near future. So it's easier to just fix the warnings.

Issue #2562, opened 2020-08-06
Refactor Expr architecture


st-pasha:
Our current approach for creating new `Expr`s is overly complicated, as outlined here: https://github.com/h2oai/datatable/blob/master/src/core/expr/!readme.md. This complexity stems mostly from the fact that the `Expr` class which performs arithmetic on `f-expressions` is defined in pure python, and then needs to be bridged into the C++ core.

A more sane approach would be to define everything in C++, eliminating most of the "middle-man" code. In particular, the following architecture is proposed:
- C++ class `py::FExpr` to replace current python `Expr` class;
- C++ class `py::ColumnNamespace` to replace current python `FrameProxy` class;
- C++ class `dt::expr::FExpr` is a merged version of current `dt::expr::Expr` and `dt::expr::Head`. The class is virtual, with the hierarchy following that of the `Head` class;
- Each `py::FExpr` contains a `shared_ptr<dt::expr::FExpr>`;
- The `dt::expr::FExpr` class defines virtual methods for evaluation and reproing;
- The `Op` enum is removed.

samukweku:
Hi @st-pasha, I think it would be helpful as well if a documentation could be created for code contributors , to know what the building blocks are, so to speak. That way when dabbling in and writing the code, I can sort of know which parts to use and what. Admittedly, my knowledge of C++ is limited, so maybe the documentation is clear and the challenge is with me.

st-pasha:
Hi @samukweku, this new refactoring effort aims at simplifying code inside the `core/expr` folder. It is quite a significant part of the codebase, but it's not the only part. So if you want to try and contribute code, feel free to look around. For example, #2166 is both relatively simple and completely independent of the Expr architecture. On the other hand if you wanted to do any PR that involves creating new functions that act within `DT[...]`, then it's best to wait a week or so until this upgrade is complete.

st-pasha:
Sure, everything is possible, though making it properly handle international text would be very tricky...
However, we can start with an ASCII version first.

Do you have a specific API in mind for this function (i.e. function name, parameters, etc)?

Peter-Pasta:
Hi @st-pasha, I don't have a specific API in mind, although I would be looking forward to using this with the upcoming case_when functionality.

PR #2573, opened 2020-08-17
qcut() converted into FExpr

st-pasha:
WIP for #2562


src/core/expr/fexpr_qcut.cc
```diff
@@ -0,0 +1,198 @@
...
+equal populations. Thus, the intervals are chosen according to
+the sample quantiles of the data.
+
+If there are duplicate values in the data, they will all be placed
+into the same been. In extreme cases this may cause the bins to be
+highly unbalanced.
+
+Parameters
+----------
+cols: FExpr
```
oleksiyskononenko, line 153: I remember we discussed that the input can be also different from `FExpr`?

src/core/expr/fexpr_qcut.cc
```diff
@@ -0,0 +1,198 @@
...
+    Input data for quantile binning.
+
+nquantiles: int | List[int]
```
oleksiyskononenko, line 156: Can also be a tuple...

st-pasha: True, anywhere where we accept a list, we also accept a tuple. 
However, when I tried to explicitly state this in argument's type, it quickly became a mess. Not here (this is a simple case), but in some other places where we have for example `List[bool | int | float | str]`. So I decided that it's just easier to document that something is a list, and leave tuple as "an extra thing that also happens to work".

src/core/expr/fexpr_qcut.cc
st-pasha: Yeah, we'll change this (and other places) once we have a documentation for what it means to be "FExpr-like".

PR #2584, opened 2020-08-25
Implement function round()

### LINKED ISSUE #2285: Implement a function to round off floating values in pydatatable
I have already explained the requirement of this case on StackOverflow , 

Here it's reference.

https://stackoverflow.com/questions/59560330/how-to-round-off-floating-values-in-pydatatable

Regards,
Mallesham

st-pasha:
The function is equivalent to standard python's `round()`. It allows rounding to the specified number of decimal digits (see the included documentation for details). 

Closes #2285

### DIFF
--- a/src/core/expr/fexpr_round.cc
+++ b/src/core/expr/fexpr_round.cc
@@ -30,32 +30,7 @@
 namespace dt {
 namespace expr {
 
-
 static constexpr int NODIGITS = std::numeric_limits<int>::min();
-
-static const int64_t pow10s[] = {
-  1,
-  10,
-  100,
-  1000,
-  10000,
-  100000,
-  1000000,
-  10000000,
-  100000000,
-  1000000000,
-  10000000000,
-  100000000000,
-  1000000000000,
-  10000000000000,
-  100000000000000,
-  1000000000000000,
-  10000000000000000,
-  100000000000000000,
-  1000000000000000000,
-  9223372036854775807,
-};
-
 
 
 //------------------------------------------------------------------------------
@@ -66,10 +41,10 @@ template <typename T>
 class RoundNeg_ColumnImpl : public Virtual_ColumnImpl {
   private:
     Column arg_;
-    int64_t scale_;
+    double scale_;
 
   public:
-    RoundNeg_ColumnImpl(Column&& arg, int64_t scale)
+    RoundNeg_ColumnImpl(Column&& arg, double scale)
       : Virtual_ColumnImpl(arg.nrows(), arg.stype()),
         arg_(std::move(arg)),
         scale_(scale)
@@ -88,7 +63,8 @@ class RoundNeg_ColumnImpl : public Virtual_ColumnImpl {
       T value;
       bool isvalid = arg_.get_element(i, &value);
       if (isvalid) {
-        *out = static_cast<T>((value + scale_/2) / scale_ * scale_);
+        *out = static_cast<T>(
+                std::rint(static_cast<double>(value) / scale_) * scale_);
       }
       return isvalid;
     }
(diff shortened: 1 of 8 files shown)


src/core/expr/fexpr_round.cc
```diff
@@ -0,0 +1,384 @@
...
+
+Parameters
+----------
+cols: FExpr
+    Input data for rounding. This could be an expression yielding
+    either a single or multiple columns. The `round()` function will
+    apply to each column independently and produce as many columns
+    in the output as there were in the input.
+
+    Only numeric columns are allowed: boolean, integer or floating.
```
oleksiyskononenko, line 326: float?

src/core/expr/fexpr_round.cc
```diff
@@ -0,0 +1,384 @@
...
+-----
+Values that are exactly half way in between their rounded neighbors
+are converted towards their nearest even value. For example, both
+`7.5` and `8.5` are rounded into `8`, whereas `6.5` is rounded as `6`.
+
+Rounding integer columns may produce unexpected results for values
+that are close to the min/max value of that column's storage type.
+For example, when an `int8` value `127` is rounded to nearest 10, it
+becomes `130`. However, since `130` cannot be represented as `int8` a
+warp-around occurs and the result becomes `-126`.
```
oleksiyskononenko, line 358: wrap-around?

Also, what about promoting column to `int16` in this case?

src/core/expr/fexpr_round.cc
```diff
@@ -0,0 +1,384 @@
...
+
+Notes
+-----
+Values that are exactly half way in between their rounded neighbors
+are converted towards their nearest even value. For example, both
+`7.5` and `8.5` are rounded into `8`, whereas `6.5` is rounded as `6`.
+
+Rounding integer columns may produce unexpected results for values
+that are close to the min/max value of that column's storage type.
+For example, when an `int8` value `127` is rounded to nearest 10, it
```
oleksiyskononenko, line 356: backquotes around `10`?

src/core/expr/fexpr_round.cc
st-pasha: Then the stype of the result would not be the function of the stype of the input anymore, and it opens a HUGE can of worms. On the other hand, numpy also does the same thing as described here:
```
>>> np.array([127], dtype='int8').round(decimals=-1)
array([-126], dtype=int8)
```

st-pasha:
The macOS failures are due to Jenkins problems, unrelated to this PR. So, merging.

src/core/expr/fexpr_round.cc
oleksiyskononenko: If input is `int8` then we can always promote to `int16`, etc, unrelated to the actual data. Otherwise it seems very dangerous to me. No one really cares about the stypes, but when `round(127) == -126` this could make a difference.

st-pasha: It is reasonable to promote int8 or int16 into int32, but it will fix the problem only for those stypes. Obviously, int32 values near the INT32_MAX will have the same issues. And int64 values near the INT64_MAX.

oleksiyskononenko: `int32` can be promoted to `int64`, but for `int64` yes, there is no way... but this is a hard limit

st-pasha: I don't think the promotion `int32` -> `int64` is desirable. After all, most users store integers that are of reasonable magnitude, far away from the boundaries. Doubling the column's storage requirements to potentially handle few corner cases is just not worth it.

oleksiyskononenko: >most users store integers that are of reasonable magnitude

I guess this is true for `int32` and `int64`, but may not be true for `int8` and `int16`. For instance, `int8` can be automatically assigned by `fread` during data import.

>Doubling the column's storage requirements to potentially handle few corner cases is just not worth it.

Well, this column is virtual anyways.

st-pasha: I don't think fread creates int8 columns right now; however we may do so in the future. But even if we do, it will only be for a 0/1 column only

PR #2590, opened 2020-08-27
Expand tests for binary operations

oleksiyskononenko:
This set of tests covers the basic binary operations on columns and scalars, it is not fully comprehensive though.

### DIFF
--- a/tests/expr/fbinary/test-add.py
+++ b/tests/expr/fbinary/test-add.py
@@ -32,7 +32,7 @@ from tests import assert_equals
 
 @pytest.mark.parametrize("type", [dt.obj64])
 def test_add_error_wrong_one_type(type):
-    DT = dt.Frame(A=[[None]*100]/type, B=[range(100)]/dt.int32)
+    DT = dt.Frame(A=[]/type, B=[]/dt.int32)
     type_str = str(type)[6:]
     msg = r"Operator \+ cannot be applied to columns of types %s and int32" % \
           (type_str,)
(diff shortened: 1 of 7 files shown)


tests/expr/fbinary/test-add.py
```diff
@@ -21,10 +21,39 @@
...
 from tests import assert_equals
 
 
+#-------------------------------------------------------------------------------
+# Errors
+#-------------------------------------------------------------------------------
+
+@pytest.mark.parametrize("type", [dt.obj64])
+def test_add_error_wrong_one_type(type):
+    DT = dt.Frame(A=[[None]*100]/type, B=[range(100)]/dt.int32)
```
st-pasha, line 35: That's a very strange frame... Have you looked at it?
Anyways, the easiest way to create a frame with given types is either
```
DT = dt.Frame(A=[]/dt.obj64, B=[]/dt.int32)
```
or
```
DT = dt.Frame(names=["A", "B"], stypes=[dt.obj64, dt.int32])
```

tests/expr/fbinary/test-add.py
```diff
@@ -21,10 +21,39 @@
...
 from datatable import dt, f
 from tests import assert_equals
 
 
+#-------------------------------------------------------------------------------
+# Errors
+#-------------------------------------------------------------------------------
+
+@pytest.mark.parametrize("type", [dt.obj64])
+def test_add_error_wrong_one_type(type):
```
st-pasha, line 34: no point in making this parameterized...

tests/expr/fbinary/test-add.py
oleksiyskononenko: Indeed, that was probably a typo at some point, not sure how it went through. Though it did created a frame with the needed types, I will simplify it, but we may want to look at what really happened here.

tests/expr/fbinary/test-add.py
oleksiyskononenko: We may have some new types added, then it will be easier to adjust this test.

tests/expr/fbinary/test-add.py
oleksiyskononenko: Ah, I guess it just converted the `[range(100)]` to one `int32` element, and `[[None]*100]` to one `obj64` element, due to `[]`. Will fix this anyways.

Issue #2586, opened 2020-08-26
API Documentation checklist


st-pasha:
List of symbols exported from the main `datatable` namespace:

## Classes
- [x] `Frame`
    - [x] `__copy__`, `__deepcopy__`
    - [x] `__getitem__`, 
    - [x] `__setitem__`, 
    - [x] `__delitem__`
    - [x] `__getstate__`, `__setstate__`
    - [x] `__init__`
    - [x] `__iter__`
    - [x] `__len__`
    - [x] `__repr__`, `_repr_html_`, `_repr_pretty_`
    - [x] `__reversed__`
    - [x] `__sizeof__`
    - [x] `__str__`
    - [x] `cbind`
    - [x] `colindex`
    - [x] `copy`
    - ~`countna`~, ~`countna1`~
    - [x] `export_names`
    - [x] `head`
    - [x] `key`
    - [x] `keys`
    - [x] `ltypes`
    - [x] `materialize`
    - ~`max`~, ~`max1`~
    - ~`mean`~, ~`mean1`~
    - ~`min`~, ~`min1`~
    - ~`mode`~, ~`mode1`~
    - [x] `names`
    - [x] `ncols`
    - ~`nmodal`~, ~`nmodal1`~
    - [x] `nrows`
    - ~`nunique`~, ~`nunique1`~
    - [x] `rbind`
    - [x] `replace`
    - ~`sd`~, ~`sd1`~
    - [x] `shape`
    - [x] `sort`
    - [x] `source`
    - [x] `stype`
    - [x] `stypes`
    - ~`sum`~, ~`sum1`~
    - [x] `tail`
    - [x] `to_csv`
    - [x] `to_dict`
    - [x] `to_jay`
    - [x] `to_list`
    - [x] `to_numpy`
    - [x] `to_pandas`
    - [x] `to_tuples`
    - [x] `view`
- [x] `FExpr`
    - [x] `__add__`
    - [x] `__and__`
    - [x] `__bool__`
    - [x] `__eq__`
    - [x] `__floordiv__`
    - [x] `__ge__`
    - [x] `__gt__`
    - [x] `__init__`
    - [x] `__invert__`
    - [x] `__le__`
    - [x] `__lshift__`
    - [x] `__lt__`
    - [x] `__mod__`
    - [x] `__mul__`
    - [x] `__ne__`
    - [x] `__neg__`
    - [x] `__or__`
    - [x] `__pos__`
    - [x] `__pow__`
[...]

PR #2594, opened 2020-08-30
Convert FTRL docs to xfunction/xclass format

oleksiyskononenko:
- create FTRL reference under the API section;
- overall improvement of FTRL docs, though it may still lack some info on exceptions.

Also, with the new system I'm not sure what is the best way to document the named tuples `FtrlParams` and `FtrlFitOutput`, that we use in `Ftrl` class.

WIP for #2586

### DIFF
--- a/docs/api/models/Ftrl/__init__.rst
+++ b/docs/api/models/Ftrl/__init__.rst
@@ -1,0 +1,4 @@
+
+.. xmethod:: datatable.models.Ftrl.__init__
+    :src: src/core/models/py_ftrl.cc Ftrl::m__init__
+    :doc: src/core/models/py_ftrl.cc doc___init__
--- a/docs/api/models/Ftrl/alpha.rst
+++ b/docs/api/models/Ftrl/alpha.rst
@@ -1,0 +1,5 @@
+
+.. xdata:: datatable.models.Ftrl.alpha
+    :src: src/core/models/py_ftrl.cc Ftrl::get_alpha Ftrl::set_alpha
+    :doc: src/core/models/py_ftrl.cc doc_alpha
+    :settable: newalpha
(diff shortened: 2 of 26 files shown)


oleksiyskononenko:
Yeah, that's just a starting point and definitely will require more work.


>make an overview list of methods/attributes in the Ftrl class;

Yep, I thought that may be we could have this as a standard feature. I guess `Frame` class will require the same kind of info displayed. 

>add general description of what FTRL is in the class description;
>alpha/beta: add the "ftrl formula" where those coefficients appear (note that you can use :math: role to add formulas);

Yes, this section should be expanded including the formula, etc.

>colnames: is it List[str]? also, note that a list is mutable, so make sure that nothing breaks if the user decides to modify that list (say, delete some elements);
>column_hashes: is it List[int]? not clear what "hashing trick" is;

I've noticed that and not sure either why both of them are lists. Will look through this and may be convert it to tuples.

>double_precision: I think this controls whether float64s or float32s are used for internal computations, but again not clear;

Yes, float32/float64 arithmetic, will clarify it.

>feature_importances: missing details, like what are the names/stypes of two columns, also what is "normalized" here? does it mean that all values are within [0;1], or that there will be a feature with importance 0, another with importance 1, and all others are in between?

It is normalized to the maximum importance, so there will be a feature with importance of 1 and all the values will be within [0; 1]. This should be added to the docs, you're right.

>nepochs : what does float (fractional) nepochs even mean? this is something rather hard to guess...

This is explained on the `__init__()` page, but not on the getter/setter page. It seems that sometimes we need to duplicate quite a lot of information...

st-pasha:
> It seems that sometimes we need to duplicate quite a lot of information...

You can always create a link from one place to another, for example:
```
nepochs : float
    The number of training epochs, see :attr:`.nepochs <Ftrl.nepochs>` for details.
```

oleksiyskononenko:
Yes, but following the link changes the context.

PR #2600, opened 2020-09-03
Promote result to float for the power operator

### LINKED ISSUE #2587: Zero raised to the negative power results in -2147483648
```
>>> from datatable import dt, f, math
>>> DT = dt.Frame([[0], [-1]])
>>> DT[:, f[0] ** f[1]]
   |          C0
-- + -----------
 0 | -2147483648

[1 row x 1 column]
```

At the same time the result is correct when a scalar or `math` module are used
```
>>> DT[:, f[0] ** -1]
   |  C0
-- + ---
 0 | inf

[1 row x 1 column]
>>> DT[:, math.pow(f[0], f[1])]
   |  C0
-- + ---
 0 | inf

[1 row x 1 column]
```

pradkrish:
Closes #2587

### DIFF
--- a/src/core/expr/fbinary/fexpr__pow__.cc
+++ b/src/core/expr/fbinary/fexpr__pow__.cc
@@ -80,7 +80,7 @@ class FExpr__pow__ : public FExpr_BinaryOp {
       xassert(lcol.nrows() == rcol.nrows());
       auto stype1 = lcol.stype();
       auto stype2 = rcol.stype();
-      auto stype0 = SType::FLOAT64;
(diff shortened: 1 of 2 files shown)


pradkrish:
I haven't fixed all the tests. If someone can comment on whether this is a good approach, I can go on to fix the rest of the failing tests.

st-pasha:
Yes, this is quite reasonable. Because raising to a power would quickly cause an overflow if the result is kept as integer. 

I believe this would fix issue #2587, so please add "Closes #2587" to your PR message.

oleksiyskononenko (approved):

PR #2601, opened 2020-09-03
Dividing by zero gives inf instead of NA

### LINKED ISSUE #2589: Dividing by zero could be possibly an inf instead of NA
```
>>> DT = dt.Frame([1])
>>> DT[:, f[0]/0]
   | C0
-- + --
 0 | NA

[1 row x 1 column]
```

The result could be possibly an `inf`, though in Python `1/0` is an exception anyways.

pradkrish:
Closes #2589

### DIFF
--- a/tests/expr/fbinary/test-truediv.py
+++ b/tests/expr/fbinary/test-truediv.py
@@ -23,6 +23,7 @@
 #-------------------------------------------------------------------------------
 import pytest
 import random
+import math
 from datatable import dt, f
 from tests import assert_equals
 
@@ -70,7 +71,7 @@ def test_div_boolean_columns():
     DT = dt.Frame(x=[True, True, True, False, False, False, None, None, None],
                   y=[True, False, None] * 3)
     RES = DT[:, f.x / f.y]
-    EXP = dt.Frame([1, None, None, 0, None, None, None, None, None]/dt.float64)
+    EXP = dt.Frame([1, math.inf, None, 0, None, None, None, None, None]/dt.float64)
     assert_equals(RES, EXP)
 
 
@@ -78,7 +79,7 @@ def test_div_boolean_column_scalar():
     DT = dt.Frame(x=[True, False, None])
     RES = DT[:, [f.x / True, f.x / False, f.x / None]]
     EXP = dt.Frame([[1, 0, None]/dt.float64,
-                   [None, None, None]/dt.float64,
+                   [math.inf, None, None]/dt.float64,
                    [None, None, None]/dt.bool8])
     assert_equals(RES, EXP)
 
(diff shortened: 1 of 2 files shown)


tests/expr/fbinary/test-truediv.py
```diff
@@ -116,21 +126,21 @@ def test_div_floats_random(seed):
 def test_mul_booleans_integers_floats_random(seed):
     random.seed(seed)
     n = 1000
-    src1 = [random.randint(-100, 100) for _ in range(n)]
+    src1 = [random.randint(1, 100) for _ in range(n)]
     src2 = [random.randint(0, 1) for _ in range(n)]
-    src3 = [random.random() * 1000 - 500 for _ in range(n)]
+    src3 = [random.random() * 1000 + 1 for _ in range(n)]
```
oleksiyskononenko, line 131: Perhaps, instead of narrowing this test down to positive numbers only, we simply create a helper function `dt_div(num, denom)` that will implement the datatable logic in terms of division and will return one of the following `num/denom` `math.inf`, `-math.inf` or `None`. Then, this function could be used to fill out the `EXP` frame below. It could also be used in the `test_div_integers()` test.

tests/expr/fbinary/test-truediv.py
```diff
@@ -92,8 +93,17 @@ def test_div_integers(seed):
...
-    EXP = dt.Frame([None if src2[i] == 0 else src1[i] / src2[i]
-                   for i in range(n)])
+    EXP_list = []
+    for i in range(n):
+        if src1[i] == 0 and src2[i] == 0:
+    	    EXP_list.append(None)
+        elif src1[i] > 0 and src2[i] == 0:
+            EXP_list.append(math.inf)
+        elif src1[i] < 0 and src2[i] == 0:
+            EXP_list.append(-math.inf)
```
oleksiyskononenko, line 103: This could be simplified to 
```
elif src2[i] == 0:
  EXP_list.append(math.copysign(math.inf, src1[i]))
```

st-pasha: I would have written
```
if src2[i] == 0:
    EXP_list.append(+math.inf if src1[i] > 0 else
                    -math.inf if src1[i] < 0 else
                    None)
else:
    EXP_list.append(src1[i] / src2[i])
```
but then it's really up to whatever each person finds most readable.

tests/expr/fbinary/test-truediv.py
```diff
@@ -116,21 +126,21 @@ def test_div_floats_random(seed):
 def test_mul_booleans_integers_floats_random(seed):
```
st-pasha, line 126: Hmm, shouldn't the test name be test_**div**_booleans_integers_floats_random?

oleksiyskononenko: Yeah, probably I had a typo in the test name when created. @pradkrish if in this PR you could rename it to `test_div_booleans_integers_floats_random` I would greatly appreciate it. Thanks.

tests/expr/fbinary/test-truediv.py
pradkrish: Done

tests/expr/fbinary/test-truediv.py
pradkrish: I will toss a coin for it. ;)

tests/expr/fbinary/test-truediv.py
pradkrish: Done

st-pasha (approved):

Issue #792, opened 2018-03-01
Implement sorting in reverse order


st-pasha:
Expected API:
```
# sort by A column in descending order
df.sort("A", descending=True)
df.sort(-f.A)
# sort by A descending, and B ascending
df.sort("A", "B", descending=[True, False])
df.sort(-f.A, f.B)
```
Note that `-f.A` should work even if `A` is a string column; and even if it is numeric, the intermediate column `-f.A` should not be computed.

jangorecki:
removed `low` tag as it is probably required to compute q8 in groupby benchmark, which is also compactly described in https://stackoverflow.com/questions/54128229/top-n-rows-by-group-using-python-datatable

st-pasha:
Tag `[low]` means something that is nice to have but nobody really wants it just yet. Since now you do need this feature, the tag ought to be removed.

PR #2610, opened 2020-09-08
Fix reverse flag in sort

### LINKED ISSUE #2505: reverse= parameter in sort not working
``Example : ``

                       'Gender': ['F', 'M', 'F', 'M', 'F',
                                       'M', 'F', 'M', 'M', 'M'],
                                         'algebra', 'stats', 'stats', 'algebra',
                                         'bio', 'bio'],
                            'Grade': grades}

          df = dt.Frame(data)

            df[:5,:, sort('Grade')]

`

                   ID | Gender | ExamYear | Class | Participated | Passed | Employed | Grade
              0 | x8 | M       | 2009     | bio     | yes        | no     | 1        | 36
              1 | x4 | F       | 2008     | algebra | no         | no     | 0        | 42
              2 | x0 | F       | 2007     | algebra | yes        | no     | 1        | 48
              3 | x7 | M       | 2009     | algebra | yes        | yes    | 1        | 68
              4 | x6 | F       | 2008     | stats   | yes        | yes    | 0        | 72

             #reverse=True ... no difference from above
             df[:5,:, sort('Grade', reverse=True)]

I had a look at the docs and could not find an explanation for this parameter, even though it was used in an example.

pradkrish:
Closes #2505

### DIFF
--- a/src/core/expr/fexpr_list.cc
+++ b/src/core/expr/fexpr_list.cc
@@ -323,7 +323,7 @@ void FExpr_List::prepare_by(
         outflags.push_back(SortFlag::DESCENDING);
       } else {
         outwf.cbind( arg->evaluate_n(ctx) );
-        outflags.push_back(SortFlag::NONE);
+        outflags.push_back(ctx.get_sort_direction());
       }
     }
   }
--- a/src/core/expr/fexpr_list.cc
+++ b/src/core/expr/fexpr_list.cc
@@ -313,7 +313,7 @@ void FExpr_List::prepare_by(
   if (kind == Kind::Str || kind == Kind::Int) {
     for (const auto& arg : args_) {
       outwf.cbind( arg->evaluate_f(ctx, 0) );
-      reverse ? outflags.push_back(SortFlag::DESCENDING) : outflags.push_back(SortFlag::NONE);
+      outflags.push_back(reverse ? SortFlag::DESCENDING : SortFlag::NONE);
     }
   }
   else if (kind == Kind::Func) {
@@ -321,10 +321,10 @@ void FExpr_List::prepare_by(
       auto negcol = arg->unnegate_column();
       if (negcol) {
         outwf.cbind( negcol->evaluate_n(ctx) );
-        reverse ? outflags.push_back(SortFlag::NONE) : outflags.push_back(SortFlag::DESCENDING);
+        outflags.push_back(reverse ? SortFlag::NONE : SortFlag::DESCENDING);
       } else {
         outwf.cbind( arg->evaluate_n(ctx) );
-        reverse ? outflags.push_back(SortFlag::DESCENDING) : outflags.push_back(SortFlag::NONE);
+        outflags.push_back(reverse ? SortFlag::DESCENDING : SortFlag::NONE);
       }
     }
   }
--- a/src/core/expr/fexpr_list.cc
+++ b/src/core/expr/fexpr_list.cc
@@ -309,10 +309,11 @@ void FExpr_List::prepare_by(
   if (args_.empty()) return;
 
   auto kind = _resolve_list_kind(args_);
+  bool reverse = ctx.reverse_sort();
   if (kind == Kind::Str || kind == Kind::Int) {
     for (const auto& arg : args_) {
       outwf.cbind( arg->evaluate_f(ctx, 0) );
-      outflags.push_back(SortFlag::NONE);
+      reverse ? outflags.push_back(SortFlag::DESCENDING) : outflags.push_back(SortFlag::NONE);
     }
   }
   else if (kind == Kind::Func) {
@@ -320,10 +321,10 @@ void FExpr_List::prepare_by(
       auto negcol = arg->unnegate_column();
       if (negcol) {
         outwf.cbind( negcol->evaluate_n(ctx) );
-        outflags.push_back(SortFlag::DESCENDING);
+        reverse ? outflags.push_back(SortFlag::NONE) : outflags.push_back(SortFlag::DESCENDING);
       } else {
         outwf.cbind( arg->evaluate_n(ctx) );
-        outflags.push_back(ctx.get_sort_direction());
+        reverse ? outflags.push_back(SortFlag::DESCENDING) : outflags.push_back(SortFlag::NONE);
       }
     }
   }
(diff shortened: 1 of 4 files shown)


pradkrish:
All the tests passed on my machine. Even on the cluster, the same tests are passing on some but not on others. Even those that fail, they are failing on functions that are not even part of this commit, any idea?

oleksiyskononenko:
@pradkrish The tests that are failing are the `groupby` tests. In datatable the `groupby` operation is based upon the `sort`, because it first sorts the data and then produces groups of equal values.

I didn't look at all the failing tests, but [some of them](https://travis-ci.org/github/h2oai/datatable/jobs/725500590) are failing because of the wrong order of the data, see below

```
E                   AssertionError: The frames have different data in column 0 `id2` at index 0: LHS has 2, and RHS has 1
E                     LHS = [2, 2, 2, 1, 1, 1]
E                     RHS = [1, 1, 1, 2, 2, 2]
```

So this is directly related to the issue you are trying to address on this PR. To reproduce this locally you may want to check your python version, note that the tests only fail for python 3.7 and 3.8, though on my mac they also fail for python 3.6 

Let me know if you have any other questions.

pradkrish:
@oleksiyskononenko Thanks for the pointers

I do all the development in python3.8 environment and that particular test, just to take an example, runs just fine on a build with this commit. Could you kindly run the following code on your mac with python 3.6 (with this commit, of course) and check whether EXP and RES are equal. Thanks and sorry for the trouble.

```python
import random
from datatable import dt, f

DT = dt.Frame(id2=[1, 2] * 3, id4=[1] * 6, v3=[1, 3, 2, 3, 3, 3])
RES = DT[:, :, dt.by(f.id2, f.id4)]
EXP = dt.Frame(id2=[1, 1, 1, 2, 2, 2], id4=[1] * 6, v3=[1, 2, 3, 3, 3, 3])
```

oleksiyskononenko:
@pradkrish Yeah, thanks, it seems that it now passes all the tests, both on Travis/Jenkins/Appveyor and for me locally. I will review it once get some more time.

src/core/expr/fexpr_list.cc
```diff
@@ -323,7 +323,7 @@ void FExpr_List::prepare_by(
         outflags.push_back(SortFlag::DESCENDING);
       } else {
         outwf.cbind( arg->evaluate_n(ctx) );
-        outflags.push_back(SortFlag::NONE);
+        outflags.push_back(ctx.get_sort_direction());
```
oleksiyskononenko, line 326: I guess this logic may have a problem. For example, if both `negcol` and `ctx.get_sort_direction()` are `true`, the order will be descending, though it should be ascending:

```python
>>> DT = dt.Frame(range(5))
>>> DT[:, :, sort(-f[0])]
   | C0
-- + --
 0 |  4
 1 |  3
 2 |  2
 3 |  1
 4 |  0

[5 rows x 1 column]
>>> DT[:, :, sort(-f[0], reverse=True)] # nothing changes, but it should
   | C0
-- + --
 0 |  4
 1 |  3
 2 |  2
 3 |  1
 4 |  0

[5 rows x 1 column]
```

pradkrish: so if we have double negatives (`negcol` and `reverse=True`), the result should be positive (that is ascending), is that how it should be? is that how it's done in R-datatable as well?

samukweku:
just as an aside, if we have the negation to mean reverse, why keep the `reverse` argument? why not deprecate it?

src/core/expr/fexpr_list.cc
pradkrish: Samuel has raised a good point (see below). Having double negatives does not seem like a good design. I would rather go with either deprecating the reverse or the behaviour to be descending (like in your example above) despite double negatives.

oleksiyskononenko: Yes, double negative normally results in a positive, see below

```python
>>> DT[:, :, sort(-(-f[0]))]
   | C0
-- + --
 0 |  0
 1 |  1
 2 |  2
 3 |  3
 4 |  4

[5 rows x 1 column]
```

Why it should be descending? Minus sign just specifies what value should be used for sorting, in the above case it is sorted by `-f[0]`, and the `reverse` parameter controls the overall sorting direction: whether it should be done asc or desc. So from the user point of view the result of `DT[:, :, sort(-f[0], reverse=True)]` should be asc sorting.

With respect to deprecating the `reverse` parameter, I would say it can be pretty useful in some cases. For instance, sometimes one doesn't know in advance if the sort should be done asc or desc, it may be an input boolean parameter or whatever else, then it is easier and more readable to write something like `DT[:, :, sort(f[0], reverse=MY_INPUT_PARAMETER)` than converting `MY_INPUT_PARAMETER` to +-1 and then multiply it by `f[0]`. 

May be @st-pasha also had some other use-cases in mind when [proposed](https://github.com/h2oai/datatable/issues/792) the API.

oleksiyskononenko:
@pradkrish @samukweku See this use-case https://github.com/h2oai/datatable/blob/master/tests/ijby/test-sort.py#L100-L105 The sort clause also supports the first parameter to be the column index, such that one can right something like `DT[:, :, sort(0)]` instead of `DT[:, :, sort(f[0])]`. In such a case `DT[:, :, sort(-0)]` will have no effect, and the `reverse` parameter is needed.

src/core/expr/fexpr_list.cc
oleksiyskononenko: Btw, when I use the column index, instead of the `f[]`,  the `reverse` parameter still doesn't work

```python
>>> DT = dt.Frame(range(5))
>>> DT[:, :, dt.sort(0, reverse=True)]
   | C0
-- + --
 0 |  0
 1 |  1
 2 |  2
 3 |  3
 4 |  4

[5 rows x 1 column]

>>> DT[:, :, dt.sort(f[0], reverse=True)]
   | C0
-- + --
 0 |  4
 1 |  3
 2 |  2
 3 |  1
 4 |  0

[5 rows x 1 column]
```

pradkrish: `Btw, when I use the column index, instead of the f[], the reverse parameter still doesn't work`

Good observation. This doesn't work on master either and I think (correct me if I am wrong), this is not directly related to how we deal with the reverse flag. Shall we create a new issue for this or combine it in the same issue? I vote for the former.

samukweku:
@oleksiyskononenko it would be consistent then that if the user wants to sort on a column index, the `f` proxy is used. ``DT[:, :, sort(-f[0])]`` as you posted works fine. I feel it is one less thing the user needs to worry about, and less maintenance of codebase. My opinion of course ...

st-pasha (approved):

PR #2611, opened 2020-09-10
Add tutorial on API comparison of R's data.table vs datatable

samukweku:
This documentation should help users from R's ``data.table`` execute similar basic operations in ``datatable``.

It covers :
- row subsetting and filtering, 
- sorting,
- selecting columns, 
- row subsetting and selection/computation in columns,
- grouping operations
- joins


docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
+.. py:currentmodule:: datatable
+
+Comparison with Rdatatable
+==========================
+
+``Datatable`` strives to mimic `Rdatatable's <https://rdatatable.gitlab.io/data.table/index.html>`_ API as closely as possible; however, there are differences due to language constraints.
```
oleksiyskononenko, line 6: I would say we strive to implement the same functionality as there is in now in `data.table`, as for the API we normally try to get the best of `data.table`, `pandas`, `numpy`, etc.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+
+This page shows how to perform similar basic operations in `Rdatatable <https://rdatatable.gitlab.io/data.table/index.html>`_  in ``datatable``.
+
+Subsetting Rows
+----------------
```
oleksiyskononenko, line 11: We normally use as many underscores as there is characters in the title, here it seems that you have one extra. This is not a big deal though.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+    import numpy as np
+
+    # data is from the example data in Rdatatable
+
+    df = dt.Frame({"x":np.repeat(["b","a","c"],3),
+                   "y": [1,3,6]*3,
+                   "v": range(1,10)}) # datatable
+
+    DT = data.table(x=rep(c("b","a","c"),each=3),
+                    y=c(1,3,6), v=1:9) # rdatatable
```
oleksiyskononenko, line 24: I would split this into two blocks of code: one for R and one for Python, because now it looks like this is a Python code that won't really work when executed. Then we could also use the same frame name `DT` in the both languages, so that it will look more consistent in the table below.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+                                    | ``df[:, :, sort(x, -f.y)]``
+===============================   ============================================ ==================================================
+
+Note the use of the ``f`` symbol when performing computations or sorting in descending order. You can read more about :ref:`f-expressions`.
+
+Selecting Columns
+------------------
+
+=========================================== ============================================ ==============================================
+RDatatable                                           Datatable                                                  Action
```
oleksiyskononenko, line 51: The Python package name is better to spell as `datatable` everywhere on this tutorial, because this is how it is named on [pypi](https://pypi.org/project/datatable/).

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+``DT[, .(m = x)]``                           ``df[:. {"m" : f.x}]``                      Rename and select column
+``DT[, .(sv=sum(v))]``                       ``df[:, {"sv": dt.sum(f.v)}]``              Sum column ``v`` and rename as ``sv``
+``DT[, .(v, v*2)]``                          ``df[:, [f.v, f.v*2]]``                     Return two columns, ``v`` and ``v`` doubled
+``DT[, 2]``                                  ``df[:, 1]``                                Select the second column
+``DT[, ncol(DT), with=FALSE]``               ``df[:, -1]``                               Select last column
+``DT[, .SD, .SDcols=x:y]``                   ``df[:, f["x":"y"]]``                       Select columns ``x`` through ``y``
+``DT[ , .SD, .SDcols = !x:y]``               | ``df[:, [name not in ("x","y")``          Exclude columns ``x`` and ``y``
+                                             |          ``for name in df.names]]`` or
+                                             | ``df[:, f[:].remove(f['x':'y'])]``
+``DT[ , .SD, .SDcols = patterns('^[xv]')]``  | ``df[:, [name.startswith(("x","v"))``     Select columns that start with x or v
```
oleksiyskononenko, line 64: Backticks around `x` and `v` seem to be missing.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+                                             |          ``for name in df.names]]``
```
oleksiyskononenko, line 65: When I render this table it seems that there is a horizontal scroll bar that appears. We should try to avoid it, but I'm not sure if the page width is a hard constraint on readthedocs. @st-pasha any ideas?

Screen Shot 2020-09-11 at 11 18 18 AM

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+=========================================== ============================================ ==============================================
+
+In ``datatable``, a single column can also be selected as ``df['v']``.
+
+Subset rows and Select/Compute
```
oleksiyskononenko, line 70: I would change "Compute" to "Aggregate".

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+-------------------------------
+
+====================================             ==========================================          ==============================================
+RDatatable                                           Datatable                                              Action
```
oleksiyskononenko, line 74: "RDatatable" -> "R's data.table", "Datatable" -> "datatable".

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+====================================             ==========================================          ==============================================
+``DT[2:3, .(sum(v))]``                            ``df[1:3, dt.sum(f.v)]``                             Sum column ``v`` over rows 2 and 3
+``DT[2:3, .(sv=sum(v))]``                         ``df[1:3, {"sv": dt.sum(f.v)}]``                     Same as above, new column name
+``DT[x=="b", .(sum(v*y))]``                       ``df[f.x=="b", dt.sum(f.v * f.y)]``                  Filter in ``i`` and aggregate in ``j``
+``DT[x=="b", sum(v*y)]``                          ``df[f.x=="b", dt.sum(f.v * f.y)][0,0]``             Same as above, return as scalar
```
oleksiyskononenko, line 79: Minor, to keep the same style everywhere: `[0,0]` -> `[0, 0]`

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+====================================             ==========================================          ==============================================
+
+In `R <https://www.r-project.org/about.html>`_, indexing starts at 1 and when slicing, the first and last items are included. However, in `Python <https://www.python.org/>`_, indexing starts at 0, and when slicing, only the first item is included.
+
+Grouping Operations - j and :func:`by()`
```
oleksiyskononenko, line 84: I would call it simply "Grouping with by" just like we already do [here](https://datatable.readthedocs.io/en/latest/manual/groupby_examples.html).

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+----------------------------------------
+
+========================================         ============================================     ============================================================
+RDatatable                                           Datatable                                              Action
```
oleksiyskononenko, line 88: "RDatatable" -> "R's data.table", "Datatable" -> "datatable".

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+                                                 | ``df[:, 'z'] = 42`` or
+                                                 | ``df = df[:, f[:].extend({"z":42})]``
+``DT[, z:=NULL]``                                | ``del df['z']`` or                               Remove column
+                                                 | ``del df[:, 'z']`` or
+                                                 | ``df = df[:,f[:].remove(f.z)]``
+``DT["a", v:=42L, on="x"]``                      | ``df[f.x=="a", update(v=42)]`` or                Subassign to exiting ``v`` column
+                                                 | ``df[f.x=="a", 'v'] = 42``
+``DT["b", v2:=84L, on="x"]``                     | ``df[f.x=="b", update(v2=84)]`` or               Subassign to new column (NA padded)
+                                                 | ``df[f.x=='b', 'v2'] = 84``
+``DT[, m:=mean(v), by=x]``                       | ``df[:, update(m=dt.mean(f.v)), by("x")]``       Add new column by group
```
oleksiyskononenko, line 141: "Add new column by group" -> "Add new column assigning values group-wise"?

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+
+       x y v i.v foo
+    1: c 1 7   8   4
+    2: c 3 8   8   4
+    3: c 6 9   8   4
+    4: b 1 1   7   2
+    5: b 3 2   7   2
+    6: b 6 3   7   2
+
+``Datatable``::
```
oleksiyskononenko, line 209: `Datatable` -> `datatable`

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+
+    df[g[-1]!=None, :, join(X)][:, dt.sum(f.v), by("x")]
+
+        x	v
+    0	b	6
+    1	c	24
+
+- Compute on columns from both dataframes in ``j``
```
oleksiyskononenko, line 294: I would again call it "Aggregate".

oleksiyskononenko:
Thanks for the PR, @samukweku, looks awesome. I left some comments, mostly cosmetical, and would also wait for @st-pasha to review the R's part as I'm not an expert on R's data.table.

docs/manual/comparison_with_rdatatable.rst
samukweku: Changed the description here with text from the readme section of the library

docs/manual/comparison_with_rdatatable.rst
samukweku: I'll take your suggestion, since it is more explicit

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,330 @@
...
+    1	b	3	2	7	2
+    2	b	6	3	7	2
+    3	a	1	4	NA	NA
+    4	a	3	5	NA	NA
+    5	a	6	6	NA	NA
+    6	c	1	7	8	4
+    7	c	3	8	8	4
+    8	c	6	9	8	4
+
+- An inner join could be simulated by removing the nulls. Again, this only works if the keyed column has unique values.
```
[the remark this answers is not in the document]
samukweku: poor sentence construction for me.

docs/manual/comparison_with_rdatatable.rst
samukweku: @oleksiyskononenko not sure how I would go about affecting the width. maybe @st-pasha has some ideas?

samukweku:
@oleksiyskononenko made some more updates for your review

docs/manual/comparison_with_rdatatable.rst
oleksiyskononenko: Yeah, that's something that is out of your control here and should be addressed globally. I guess you have a reasonable amount of content presented.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,338 @@
...
+
+``datatable`` is closely related to R's `data.table <https://data.table.gitlab.io/data.table/index.html>`__ attempts to mimic its core algorithms and API; however, there are differences due to language constraints.
+
+This page shows how to perform similar basic operations in R's `data.table <https://data.table.gitlab.io/data.table/index.html>`__  versus ``datatable``.
+
+Subsetting Rows
+---------------
+The sample data is from the example data in R's `data.table <https://data.table.gitlab.io/data.table/index.html>`__.
+
+``data.table``::
```
oleksiyskononenko, line 14: Guess you can do here

```
.. code-block:: R
```

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,338 @@
...
+
+    library(data.table)
+
+    DT = data.table(x=rep(c("b","a","c"),each=3),
+                    y=c(1,3,6), v=1:9)
+
+
+``datatable``::
```
oleksiyskononenko, line 22: and  here
```
.. code-block:: python
```
Though I'm not sure how much different the rendering will be.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,338 @@
...
+
+    DT[:, dt.sum(f.v*g.foo), join(X), by(f.x)][f[-1]!=0, :]
+
+        x	C0
+    0	b	12
+    1	c	96
+
+- Compute on columns with same name from both dataframes in ``j``
```
oleksiyskononenko, line 320: "Compute" -> "Aggregate"?

oleksiyskononenko (commented):
Looks awesome, @samukweku. I just left several minor comments, and let's see if we can do something to better render the wise tables on our side.

docs/manual/comparison_with_rdatatable.rst
oleksiyskononenko: @samukweku While we're trying to solve the problem with rendering of this table, I thought it could look more logical if we  move the "Action" column to the first place, i.e. the column order of "Action/data.table/datatable". What do you think?

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,354 @@
...
+=========================================== ============================================ ==============================================
+``DT[, .(v)]``                               ``DT[:, 'v']``                              Select column ``v``
+``DT[, .(x,v)]``                             ``DT[:, ['x', 'v']]``                       Select multiple columns
+``DT[, .(m = x)]``                           ``DT[:. {"m" : f.x}]``                      Rename and select column
+``DT[, .(sv=sum(v))]``                       ``DT[:, {"sv": dt.sum(f.v)}]``              Sum column ``v`` and rename as ``sv``
+``DT[, .(v, v*2)]``                          ``DT[:, [f.v, f.v*2]]``                     Return two columns, ``v`` and ``v`` doubled
+``DT[, 2]``                                  ``DT[:, 1]``                                Select the second column
+``DT[, ncol(DT), with=FALSE]``               ``DT[:, -1]``                               Select last column
+``DT[, .SD, .SDcols=x:y]``                   ``DT[:, f["x":"y"]]``                       Select columns ``x`` through ``y``
```
st-pasha, line 67: or `DT[:, 'x':'y']`

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,354 @@
...
+              MyMin=min(v),
+              MyMax=max(v)),
+       by=.(x, y%%2)]
+
+The same can be replicated in ``datatable`` by using a dictionary ::
+
+    DT[:, {MySum: dt.sum(f.v),
+           MyMin: dt.min(f.v),
+           MyMax: dt.max(f.v)},
+       by(f.x, f.y%2)]
```
st-pasha, line 128: If using a dictionary, the keys must be quoted: `{"MySum": dt.sum(f.v), ...}`. Alternatively, you can use python's `dict()` constructor to get the same syntax as in R:
```
DT[:, dict(MySum=dt.sum(f.v),
           MyMin=dt.min(f.v),
           MyMax=dt.max(f.v)),
   by(f.x, f.y % 2)]
```

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,354 @@
...
+Note the use of the ``f`` symbol when performing computations or sorting in descending order. You can read more about :ref:`f-expressions`.
+
+Selecting Columns
+-----------------
+
+=========================================== ============================================ ==============================================
+data.table                                           datatable                                                  Action
+=========================================== ============================================ ==============================================
+``DT[, .(v)]``                               ``DT[:, 'v']``                              Select column ``v``
+``DT[, .(x,v)]``                             ``DT[:, ['x', 'v']]``                       Select multiple columns
```
st-pasha, line 61: I think there should also be an example of selecting a column whose name is in a variable, for example if `var='v'` then it's `DT[var]` in Python, and `DT[, ..var]` in R.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,354 @@
...
+Add/Update/Delete Columns
+-------------------------
+
+========================================         ===============================================  ============================================================
+data.table                                           datatable                                              Action
+========================================         ===============================================  ============================================================
+``DT[, z:=42L]``                                 | ``DT[:, update(z=42)]`` or                       Add new column
+                                                 | ``DT['z'] = 42`` or
+                                                 | ``DT[:, 'z'] = 42`` or
+                                                 | ``DT = DT[:, f[:].extend({"z":42})]``
```
st-pasha, line 140: Also: create a new column whose name is given in a variable `var`: in python it's `DT[var] = 4242`.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,354 @@
...
+                                                 | ``DT = DT[:, f[:].remove(f.z)]``
+``DT["a", v:=42L, on="x"]``                      | ``DT[f.x=="a", update(v=42)]`` or                Subassign to existing ``v`` column
+                                                 | ``DT[f.x=="a", 'v'] = 42``
+``DT["b", v2:=84L, on="x"]``                     | ``DT[f.x=="b", update(v2=84)]`` or               Subassign to new column (NA padded)
+                                                 | ``DT[f.x=='b', 'v2'] = 84``
+``DT[, m:=mean(v), by=x]``                       | ``DT[:, update(m=dt.mean(f.v)), by("x")]``       Add new column, assigning values group-wise
+========================================         ===============================================  ============================================================
+
+Note that the :func:`update` function, as well as the ``del`` function operates in-place; there is no need for reassignment. Another advantage of the :func:`update` method is that the row order of the dataframe is not changed, even in a groupby; this comes in handy in a lot of transformation operations.
```
st-pasha, line 152: `del` is a python keyword (not function)

st-pasha (commented):
Thanks Samuel, this is a great tutorial! I've sprinkled a few comments here and there, where there were small errors, or the syntax could be made more concise, or some easy examples added.

docs/manual/comparison_with_rdatatable.rst
```diff
@@ -0,0 +1,354 @@
+.. py:currentmodule:: datatable
+
+Comparison with R's data.table
```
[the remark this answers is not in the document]
samukweku: didnt change the outcome. 

```
.. py:currentmodule:: datatable

.. raw:: html
    <style>
    .wy-nav-content { max-width: 100%; }
    </style>

Comparison with R's data.table
```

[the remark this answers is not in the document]
st-pasha: (there should be a blank line after `.. raw:: html`)

docs/manual/comparison_with_rdatatable.rst
samukweku: hazards of copy and paste. Thanks

docs/manual/comparison_with_rdatatable.rst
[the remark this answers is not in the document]
samukweku: works fine. Thanks.

docs/manual/comparison_with_rdatatable.rst
samukweku: I put the `Action` column at the end, because it follows a read from left to right approach; the user familiar with R sees the data.table code, sees the python code, and then sees the action ... in this scenario, the `Action` column is more like comments without the `#`

samukweku: I could probably find a better word than `Action`, ... maybe `Explanation` ?

oleksiyskononenko: Yes, may be "Operation" or may be even just remove the caption. It is pretty obvious just reading the text, that it is a description of the particular operation on data.

Regarding the column order, I agree that for an experienced R user it is ok to start with the R's code right away, but for all the others it will require some time to understand what's going on. If we do it as "Action/data.table/datatable", then, reading it from left to right, a person will immediately get a feeling of what kind of functionality we're going to present and then see the examples.

st-pasha (approved):

st-pasha:
So, I've been trying to experiment with how we can improve the layout to avoid those overly-wide tables, and here's what i've got:
Screen Shot 2020-09-15 at 2 11 39 PM

Generally, the idea is to have a layout which would allow longer formulas/descriptions without blowing through the page width.

What do you think?

oleksiyskononenko:
Yes, I've been also looking through the similar tutorials and most of them use this kind of layout. This seems to be more flexible.

st-pasha:
@oleksiyskononenko Can you post some links to the tutorials you've seen?

oleksiyskononenko:
@st-pasha 

https://datascience-enthusiast.com/R/pandas_datatable.html
https://towardsdatascience.com/python-pandas-vs-scala-how-to-handle-dataframes-part-ii-d3e5efe8287d

There was also one on comparing pandas vs numpy, but I can find it now for some reason. In most of the cases the content was divided by sections, each section corresponding to one particular operation, let's say "Adding a column"
and then followed by the two-column comparison. Something similar to what you just did, but your design looks more neat and compact.

st-pasha:
I also found these pandas tutorials: https://pandas.pydata.org/docs/getting_started/comparison/comparison_with_r.html

samukweku:
I had checked out the Pandas tutorials, but didnt go with it, since it had just two columns, and I wanted to include an `Action` column that described what was happening for each example.

st-pasha:
Another tutorial, this time compares pandas vs dplyr: https://jarvmiller.github.io/2018/02/28/r-pandas/

vjtrost88:
@st-pasha I used that table you made on subsetting and aggregation to figure out how to filter rows by group -- many thanks, but I did spend a significant amount of time trying to figure out why the following didn't work:
```
DT[f.x=="b"]  
#TypeError: Column selector must be an integer or a string, not <class 'datatable.FExpr'>
```
This wouldn't just return all the rows where `x` had a value of `b`, until I realized I needed to specify that I wanted to return all the columns as well.  
#### Solution:
```
DT[f.x=="b",:]


x | y | v
-- | -- | --
0 | b | 1 | 1
1 | b | 3 | 2
2 | b | 6 | 3

```

Would you consider adding this small example to your table? It might help someone else. I'd also love to suggest this as an improvement to make in a future datatable version if you can point me in the right direction. I think needing to put that extra colon will throw a bunch of R data.table users off, as it's a pythonic trait.

samukweku:
@vjtrost88 , thanks for the feedback. I think it would be better to add a note or an explanation somewhere that highlights this difference so that users wont get stomped. I'll have to look for a convenient place to introduce this. If you have more suggestions, or wish to make changes to the docs, please feel free to submit a PR.

PR #2612, opened 2020-09-11
Make empty sort API consistent

### LINKED ISSUE #2220: Unexpected behavior when using empty sort()
When sort() is used in the by() position. Given a table like:
```
   | C0  C1
-- + --  --
 0 |  1   6
 1 |  1   5
 2 |  0   4
 3 |  0   3
 4 |  2   1
 5 |  2   2
```
If we use `DT.sort()` we get: 
```
   | C0  C1
-- + --  --
 0 |  0   3
 1 |  0   4
 2 |  1   5
 3 |  1   6
 4 |  2   1
 5 |  2   2
```
The default behavior is to sort by both columns. However, if we use `DT[:,:,sort()]` it is treated as an empty group by (it returns the same Frame). This isn't really a bug, but it seems inconsistent. `DT[:,:,sort(0)]` has the same behavior as `DT.sort(0)`, so it would make sense for `DT[:,:,sort()]` to have the same behavior as `DT.sort()`.

pradkrish:
Closes #2220

### DIFF
--- a/docs/releases/v0.11.0.rst
+++ b/docs/releases/v0.11.0.rst
@@ -157,8 +157,9 @@
     previously it could return a dict of Frames if multiple sources were
     detected. Use :func:`iread()` if you need to read multi-source input.
 
-  -[api] Function: The behaviour of DT.sort() is made consistent with
-    DT[:, :, sort()] when the list of columns to sort is empty.
+  -[api] The behaviour of :meth:`.sort()`  is made consistent with
+    func:`sort()`. When the list of columns to sort is empty, both will not
+    sort any columns.
 
   General
   -------
@@ -255,4 +256,3 @@
     1   @mmalohlava
     1   @Viktor-Demin
     1   @chrinide            <Chrinide>
-    1   @pradkrish           <Pradeep Krishnamurthy>
--- a/docs/releases/v0.11.0.rst
+++ b/docs/releases/v0.11.0.rst
@@ -157,6 +157,8 @@
     previously it could return a dict of Frames if multiple sources were
     detected. Use :func:`iread()` if you need to read multi-source input.
 
+  -[api] Function: The behaviour of DT.sort() is made consistent with
+    DT[:, :, sort()] when the list of columns to sort is empty.
 
   General
   -------
@@ -253,3 +255,4 @@
     1   @mmalohlava
     1   @Viktor-Demin
     1   @chrinide            <Chrinide>
+    1   @pradkrish           <Pradeep Krishnamurthy>
--- a/src/core/sort.cc
+++ b/src/core/sort.cc
@@ -1469,10 +1469,8 @@ py::oobj py::Frame::sort(const PKArgs& args) {
   dt::expr::EvalContext ctx(dt);
 
   if (args.num_vararg_args() == 0) {
-    py::otuple all_cols(dt->ncols());
-    for (size_t i = 0; i < dt->ncols(); ++i) {
-      all_cols.set(i, py::oint(i));
-    }
+    size_t n_cols = 0;
+    py::otuple all_cols(n_cols);
     ctx.add_sortby(py::osort(all_cols));
   }
   else {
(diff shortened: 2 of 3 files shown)


src/core/sort.cc
```diff
@@ -1469,10 +1469,8 @@ py::oobj py::Frame::sort(const PKArgs& args) {
   dt::expr::EvalContext ctx(dt);
 
   if (args.num_vararg_args() == 0) {
-    py::otuple all_cols(dt->ncols());
-    for (size_t i = 0; i < dt->ncols(); ++i) {
-      all_cols.set(i, py::oint(i));
-    }
+    size_t n_cols = 0;
+    py::otuple all_cols(n_cols);
```
oleksiyskononenko, line 1473: I guess now a better name for this variable is `no_cols`, so you can just do `py::otuple no_cols(0);` here.

oleksiyskononenko (approved):

oleksiyskononenko:
@pradkrish It looks great, thanks for the PR!

src/core/sort.cc
pradkrish: Done

docs/releases/v0.11.0.rst
```diff
@@ -157,6 +157,8 @@
     previously it could return a dict of Frames if multiple sources were
     detected. Use :func:`iread()` if you need to read multi-source input.
 
+  -[api] Function: The behaviour of DT.sort() is made consistent with
+    DT[:, :, sort()] when the list of columns to sort is empty.
```
oleksiyskononenko, line 161: Guess it is better to say

```
The behaviour of  :meth:`.sort()`  is made consistent with the :func:`sort()` clause. When the list of columns to sort is empty, both will not sort any columns.
```

Note the usage of the `:meth` and `func:`, this will automatically create links to documentation.

docs/releases/v0.11.0.rst
```diff
@@ -253,3 +255,4 @@
     1   @mmalohlava
     1   @Viktor-Demin
     1   @chrinide            <Chrinide>
+    1   @pradkrish           <Pradeep Krishnamurthy>
```
oleksiyskononenko, line 258: No need to worry about this, we have a tool that will automatically extract all the contribution details from github. Once released, this information will be included in this changelog.

docs/releases/v0.11.0.rst
pradkrish: Thanks very much.

src/core/sort.cc
```diff
@@ -1469,10 +1469,8 @@ py::oobj py::Frame::sort(const PKArgs& args) {
   dt::expr::EvalContext ctx(dt);
 
   if (args.num_vararg_args() == 0) {
-    py::otuple all_cols(dt->ncols());
-    for (size_t i = 0; i < dt->ncols(); ++i) {
-      all_cols.set(i, py::oint(i));
-    }
+    size_t no_cols = 0;
+    py::otuple all_cols(no_cols);
```
oleksiyskononenko, line 1473: Was it by accident that `no_cols` changed back to `all_cols`?

pradkrish: The previous commit with 
```python
py::otuple no_cols(0)
ctx.add_sortby(py::osort(no_cols))
```
failed during the test. I couldn't open to find out where the error was, the server failed to load the page after many attempts. Since some compilers could be very strict, I assumed `0` could have been interpreted as `int` and failed since `otuple(size_t n)` expects `size_t`. Changing it back to the way it is now worked.

oleksiyskononenko: I see, I don't think it would fail due to this reason, but you can even omit `0` and just say `py::otuple no_cols;`. By default the tuple will have zero columns.

pradkrish: well, I changed it back and it worked this time. :)

oleksiyskononenko: Yeah, no reason to work. Looks good, will merge now.

oleksiyskononenko (approved):

PR #2614, opened 2020-09-12
Improve documentation on models

oleksiyskononenko:
- improve FTRL docs, add more validation and tests including a very minor API change;
- add `kfold()` and `kfold_random()` docs;
- add `aggregate()` docs, a minor API change and file renaming for consistence.

WIP for #2586


src/core/models/aggregate.cc
```diff
@@ -35,60 +35,106 @@
...
 
-Returns
--------
-A list `[frame_exemplars, frame_members]`, where
-- `frame_exemplars` is the aggregated `frame` with an additional
-  `members_count` column, that specifies number of members for each exemplar.
-- `frame_members` is a one-column datatable that contains `exemplar_id` for
-  each row from the original `frame`.
-)"
+return: List[Frame, Frame]
```
st-pasha, line 103: This is not a valid type. You can have either `Tuple[Frame, Frame]` or `List[Frame]`. I guess in this case it would make more sense to change the return type to be a tuple.

src/core/models/py_ftrl.cc
```diff
@@ -69,58 +69,71 @@ mantissa_nbits=10, nepochs=1, double_precision=False, negative_class=False,
...
 
-Create a new :class:`FTRL <datatable.models.Ftrl>` model object.
+Create a new :class:`Ftrl <datatable.models.Ftrl>` object.
 
 Parameters
 ----------
-alpha: float
-    :math:`\alpha` in per-coordinate learning rate formula, should be
-    positive. See :attr:`.alpha <models.Ftrl.alpha>` for details.
+alpha: float (optional)
```
st-pasha, line 76: I'm not sure we need these "(optional)" marks. Any parameter that has a default value in the signature is going to be optional.

st-pasha (approved):

src/core/models/py_ftrl.cc
oleksiyskononenko: Yes, that's true. I just saw we had it somewhere in the docs and thought we're going to have it everywhere... though it seems now that it was only `kfold` that has that mark.

src/core/models/aggregate.cc
oleksiyskononenko: Why it is not valid?

Issue #2605, opened 2020-09-06
Documentation rendering issues


oleksiyskononenko:
- [x] `**cols` is parsed as `*, *, cols`. This issue seems to be introduced just recently, because it only shows up in the most recent `master`, but not on [readthedocs](https://datatable.readthedocs.io/en/latest/api/frame/__init__.html). <img width="735" alt="Screen Shot 2020-09-06 at 1 30 47 AM" src="https://user-images.githubusercontent.com/35204136/92321671-98978100-efe0-11ea-9342-143c756e9935.png">

- [x] A [glitch](https://datatable.readthedocs.io/en/latest/api/fexpr/__add__.html) in background, when multiple parameters are specified on the same line, i.e `x, y: FExpr` <img width="748" alt="Screen Shot 2020-09-06 at 1 32 34 AM" src="https://user-images.githubusercontent.com/35204136/92321695-dac0c280-efe0-11ea-88b7-d5ac936c2ead.png">

- [x] See if we can slightly increase the page width to avoid horizontal scroll bar appearing here: https://github.com/h2oai/datatable/pull/2611#discussion_r487213006

- [x] [Sometimes](https://datatable.readthedocs.io/en/latest/api/models/Ftrl/model_type.html) the line spacing on the bulleted lists doesn't look good <img width="732" alt="Screen Shot 2020-09-12 at 3 52 57 AM" src="https://user-images.githubusercontent.com/35204136/92994006-7358b600-f4ab-11ea-8363-28622ec801ac.png">

- [x] Fix styles in the third-level menu in the sidebar: 
  <img width="302" alt="Screen Shot 2020-09-15 at 4 19 28 PM" src="https://user-images.githubusercontent.com/4231472/93274447-40dedf80-f76f-11ea-8d28-6c278c47167f.png">

PR #2620, opened 2020-09-15
Handle multi-params in function 'Parameters' section

st-pasha:
This PR allows to declare multiple parameters at once in the "parameters" section of documentation, e.g.
```
Parameters
----------
x, y, z: float
    (description of parameters)
```

WIP for #2605

### DIFF
--- a/docs/_static/xfunction.css
+++ b/docs/_static/xfunction.css
@@ -182,7 +182,9 @@
 .xparam-head {
     align-items: baseline;
     background: #f5f5f5;
+    color: #ddd;  /* affects only the separators */
     display: flex;
+    font-family: "SF Mono", Menlo, monospace;
 }
 
 .xparam-head .param {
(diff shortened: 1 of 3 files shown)


oleksiyskononenko:
What about making the grey color that is used for `,` and parameter type a little bit darker? Or at least the `,` to use the same style as the parameter type?

Screen Shot 2020-09-15 at 11 21 23 AM

st-pasha:
Right now it is the same color as the separator `|` between types on the RHS. I agree that it looks kinda weird, but not sure whether it's the problem with the color, or maybe we need to do something else.

But I'd say it's a minor issue that we can address after the release.

oleksiyskononenko:
Yeah, if the reason is not obvious it is, of course, not a high priority issue...

oleksiyskononenko (approved):

oleksiyskononenko:
May be we can just use the blue background for the whole `x, y` stuff in this case. Will probably look way better than `x`, `y`.

PR #2637, opened 2020-09-17
Documentation for internal. submodule

st-pasha:
- Create documentation for the function in the `datatable.internal.` module;

- Adjusted permalink URLs so that they would be based off the HEAD commit instead of master. This means that in order for the links to work one would need to commit and push the local changes to GitHub first.

WIP for #2586

### DIFF
--- a/src/core/datatablemodule.cc
+++ b/src/core/datatablemodule.cc
@@ -115,7 +115,7 @@ static const char* doc_frame_columns_virtual =
 R"(frame_columns_virtual(frame)
 --
 
-Return the list indicating which columns in the Frame are virtual.
+Return the list indicating which columns in the `frame` are virtual.
 
 Parameters
 ----------
@@ -186,17 +186,17 @@ R"(frame_integrity_check(frame)
 
 This function performs a range of tests on the `frame` to verify
 that its internal state is consistent. It returns None on success,
-or throws an AssertionError if any problems were found.
+or throws an :exc:`AssertionError` if any problems were found.
 
 Parameters
 ----------
 frame: Frame
-    A :class:`Frame` that needs to be checked for internal consistency.
+    A :class:`Frame` object that needs to be checked for internal consistency.
 
 return: None
 
 except: AssertionError
-    An error is raised if there were any issues with the `frame`.
+    An exception is raised if there were any issues with the `frame`.
 )";
 
 static py::PKArgs args_frame_integrity_check(
@@ -215,7 +215,7 @@ static void frame_integrity_check(const py::PKArgs& args) {
 
 static const char* doc_in_debug_mode =
 R"(
-Return True if datatable was compiled in debug mode.
+Return `True` if :mod:`datatable` was compiled in debug mode.
 
 .. deprecated:: 0.11.0
 )";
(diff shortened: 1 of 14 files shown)


src/core/datatablemodule.cc
```diff
@@ -153,15 +180,28 @@ static py::oobj frame_column_data_r(const py::PKArgs& args) {
...
+
+Parameters
+----------
+frame: Frame
+    A :class:`Frame` that needs to be checked for internal consistency.
+
+return: None
+
+except: AssertionError
+    An error is raised if there were any issues with the `frame`.
```
oleksiyskononenko, line 199: For consistency with all the other pages I would write it as "The exception is raised".

src/core/datatablemodule.cc
```diff
@@ -173,10 +213,16 @@ static void frame_integrity_check(const py::PKArgs& args) {
 }
 
 
+static const char* doc_in_debug_mode =
+R"(
+Return True if datatable was compiled in debug mode.
```
oleksiyskononenko, line 218: I would add back-ticks around `True`.

src/core/datatablemodule.cc
```diff
@@ -187,10 +233,22 @@ static py::oobj in_debug_mode(const py::PKArgs&) {
...
+
+Calling this function will cause the threads to spawn if they
+haven't done already. (This behavior may change in the future).
+
+Parameters
+----------
+return: List[str]
+    The list of thread ids used by the datatable. The first element
+    in the list is the id of the main thread.
+)";
```
oleksiyskononenko, line 248: May be add a "See also" section with a link to `dt.options.nthreads`? Though we first need to merge my PR then to test.

oleksiyskononenko (approved):

src/core/datatablemodule.cc
st-pasha: ok

src/core/datatablemodule.cc
st-pasha: done

src/core/datatablemodule.cc
st-pasha: I added the see also link, hopefully it'll resolve once your PR is merged

src/core/datatablemodule.cc
```diff
@@ -186,17 +186,17 @@ R"(frame_integrity_check(frame)
...
 ----------
 frame: Frame
-    A :class:`Frame` that needs to be checked for internal consistency.
+    A :class:`Frame` object that needs to be checked for internal consistency.
 
 return: None
 
 except: AssertionError
-    An error is raised if there were any issues with the `frame`.
+    An exception is raised if there were any issues with the `frame`.
```
oleksiyskononenko, line 199: It seems that in other places we're saying "the exception", probably meaning that, like in this case, it is the `AssertionError` exception. But at the same time it is "an exception", because it is just one of the exceptions. Not sure what is better.

st-pasha: I would leave "an" here, because there could really be over a dozen different exceptions here (though all of them are AssertionErrors).

PR #2638, opened 2020-09-17
Fix FTRL algo image on readthedocs

oleksiyskononenko:
Use raw html instead of the rst `figure::`, minor improvement to the image itself.


st-pasha (approved):

st-pasha:
Yep, seems to work well now.

PR #2643, opened 2020-09-20
Update version number after the release

st-pasha:


ci/Jenkinsfile.groovy
```diff
@@ -127,7 +127,7 @@ ansiColor('xterm') {
                         }
 
                         isMasterJob = (env.CHANGE_BRANCH == null || env.CHANGE_BRANCH == '')
-                        isRelease   = env.CHANGE_BRANCH.startsWith("rel-")
+                        isRelease   = !isMasterJob && env.CHANGE_BRANCH.startsWith("rel-")
```
oleksiyskononenko, line 130: Does adding `!isMasterJob` in this condition really changes anything? If `env.CHANGE_BRANCH.startsWith("rel-") == true` then `!isMasterJob` is automatically `true`...

oleksiyskononenko (approved):

ci/Jenkinsfile.groovy
st-pasha: but if `CHANGE_BRANCH` is `null`, then `.startsWith()` will throw an error, which is really bad.

oleksiyskononenko: Yeah, if Jenkins throws exceptions, that's not good :)

Issue #2642, opened 2020-09-19
Drop support for python 3.5


st-pasha:
Python 3.5 reached the end of its life on Sep 13, 2020 (https://endoflife.date/python). 
In particular, python 3.5 support will be dropped by pip in January 2021 (around the time we're planning the next release), so even if we created binary wheels for py3.5, they wouldn't be installable with the latest pip.

Dropping support for python 3.5 gives us certain minor advantages when writing tests: we no longer have problems with random keys in a dictionary, and will be able to use f-strings freely.

st-pasha:
Note that pandas already doesn't support python 3.5: https://pandas.pydata.org/pandas-docs/stable/getting_started/install.html#python-version-support

Issue #2639, opened 2020-09-18
Release 0.11.0


st-pasha:
## GitHub

- [x] Open a "Release X.Y.0" issue, and copy the text of this issue, clearing all checkboxes. 

- [x] Visit the [Milestone](https://github.com/h2oai/datatable/milestones) corresponding to the version being released, and remove all issues that are still pending (except for those that will be closed during the release).

- [x] Closed issues / merged PRs must be appropriately assigned to a Milestone, in order for our system of counting contributors to work properly. In particular, any closed PR with a milestone will be attributed to its author. If a PR has no milestone, it will not be counted. Likewise, closed issues will attributed to their authors if and only if they are marked with a Milestone. 

  - [x] Check that all PRs [merged since last release](https://github.com/h2oai/datatable/pulls?q=is:pr+is:merged+no:milestone) have a milestone. 

  - [x] Check that all PRs that were [closed without merging](https://github.com/h2oai/datatable/pulls?q=is:pr+is:unmerged+is:closed+milestone:*) do not have a milestone. 

  - [x] Check that all issues that were [closed via a PR](https://github.com/h2oai/datatable/issues?q=is:issue+is:closed+no:milestone+linked:pr) have a milestone. 

  - [x] Check that all issues that were [closed without a PR](https://github.com/h2oai/datatable/issues?q=is:issue+is:closed+-linked:pr+milestone:*) have no milestone. 

## Local preparation

- [x] Create local branch `rel-X.Y` (based on latest `master`) corresponding to the release number. Do not push to GitHub yet.

- [x] Run `python ci/headers.py src/core` to check that there are no missing headers in the C++ source files.

- [x] The current version of the project is stored in the [VERSION.txt](https://github.com/h2oai/datatable/blob/master/VERSION.txt) file. For the release, the version must be a plain dotted number such as `0.11.0`, without any extra suffixes. Update the file and commit changes to your local branch.

## Documentation

- [x] Find the file "release notes" file corresponding to the version being released. It should be in `docs/releases/vX.Y.0.rst`. Update the `.. changelog::` directive to include the version number and the release date. (Note: you would not be able to specify the wheels yet). 

- [x] Open file `docs/releases/index-releases.rst` and update the entry for the version being released.

- [x] In the `docs/` folder and run `make clean && make html`. Ensure that the documentation builds without errors/warnings. 

- [x] Open the locally built documentation in the browser and check that it rendered properly. In particular, pay attention to broken links. If necessary, make changes and commit.

[...]

PR #2657, opened 2020-09-23
Convert appveyor.yml to the PowerShell format

### LINKED ISSUE #2653: Convert appveyor.yml to PowerShell
It seems that the PowerShell syntax is much more powerful than the cmd one, at least it allowed us to solve the `if ... else` problem when building the main branch https://github.com/h2oai/datatable/pull/2651/. So there is no reason to keep both `ps` and `cmd` formats in the `appveyor.yml` file, and it should just use `ps`.

oleksiyskononenko:
Improve AppVeyor building script and convert it to the PowerShell format.

Closes #2653

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,21 +20,21 @@ build_script:
 
     $env:DT_HARNESS = "AppVeyor"
 
-    $env:MAIN_BUILD_ID = git rev-list --count main
+    $MAIN_BUILD_ID = git rev-list --count main
 
 
     # Do not take into account the AppVeyor's additional merge commit
 
-    $env:BRANCH_BUILD_ID = (git rev-list --count main..) - 1
+    $BRANCH_BUILD_ID = (git rev-list --count main..) - 1
 
     if ($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH -match "^rel-") {
       $env:DT_RELEASE="True"
     } elseif ($env:APPVEYOR_PULL_REQUEST_NUMBER -ne $null) {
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -13,112 +13,116 @@ artifacts:
...
-    pip install pytest docutils pandas
+    echo "DT_BUILD_NUMBER = $env:DT_BUILD_NUMBER"
 
-    pytest .\tests\
+    echo "DT_BUILD_SUFFIX = $env:DT_BUILD_SUFFIX"
 
 
 
-    rem Python 3.6
+    # Build and test wheel for Python 3.5
```
st-pasha, line 68: Note that we're dropping python 3.5 support in the next release (#2642)

appveyor.yml
```diff
@@ -1,4 +1,4 @@
-version: 0.10.{build}
+version: 1.0.{build}
```
st-pasha, line 1: I wonder what does this version do? Should we mention in the release docs that it should be updated with each release? (if yes, then please add a line into issue #2639)

st-pasha (approved):

appveyor.yml
oleksiyskononenko: Yeah, we can remove it at any time here.

appveyor.yml
oleksiyskononenko: Frankly, I have no idea :) It was there and I just made it look closer to what we call the version. However, this is not the only place where one can set up the version (there is also a field in the project settings). It would be nice to understand what it means, but in reality we don't need any of those... the ideal solution is just to drop it everywhere, where it makes no sense for us.

PR #2658, opened 2020-09-24
Expanded ifelse() function to allow multiple conditions

### LINKED ISSUE #2656: CASE-WHEN functionality
This is a request for function that is equivalent to SQL's [CASE WHEN](https://www.postgresql.org/docs/current/functions-conditional.html). Implementations in other libraries include:
  - R data.table's [`fcase()`](https://rdatatable.gitlab.io/data.table/reference/fcase.html);
  - dplyr's [`case_when()`](https://dplyr.tidyverse.org/reference/case_when.html);
  - numpy's [`select()`](https://numpy.org/doc/stable/reference/generated/numpy.select.html);
  - pandas N/A.

I'm thinking that the most straightforward API is to extend the [ifelse()](https://datatable.readthedocs.io/en/latest/api/dt/ifelse.html) function:
```
ifelse(cond1, value1, cond2, value2, ..., default)
```
For example:
```
ifelse(f.X > 5,  f.X - 5,
       f.X < -5, -f.X - 5,
       0)
```

st-pasha:
New ifelse() function allows chained evaluation of several conditions, making it the equivalent of SQL's CASE WHEN.

 - The evaluation is "lazy": only the smallest amount of expressions needed to get the result is computed for each row;
 - Both `condition_i` and `value_i` can have unequal but compatible shapes: some of them can be full-sized, others grouped, yet others simple scalars;
  - The value columns will be upcasted into the smallest common stype as necessary.

Closes #2656

### DIFF
--- a/docs/manual/comparison_with_rdatatable.rst
+++ b/docs/manual/comparison_with_rdatatable.rst
@@ -517,9 +517,6 @@ This is a list of some functions in ``data.table`` that do not have an equivalen
    - `between <https://rdatatable.gitlab.io/data.table/reference/between.html>`__
    - `%chin% <https://rdatatable.gitlab.io/data.table/reference/chmatch.html>`__
 
-- Conditional functions
-   - `fcase <https://rdatatable.gitlab.io/data.table/reference/fcase.html>`__
-
 - Duplicate functions
    - `duplicated <https://rdatatable.gitlab.io/data.table/reference/duplicated.html>`__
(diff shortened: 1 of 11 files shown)


samukweku:
This is great! I will have to redo my documentation on conditional expressions

st-pasha:
(The only failing tests are on py35, so I'm going to ignore them)

PR #2659, opened 2020-09-24
Na position sort

### LINKED ISSUE #793: Implement `na_position` argument for `sort()`
Possible values: `"first"` (default), `"last"`, `"remove"`.

pradkrish:
Closes #793


src/core/expr/py_sort.cc
```diff
@@ -114,6 +118,9 @@ const std::vector<bool>& osort::osort_pyobject::get_reverse() const {
   return *reverse_;
 }
 
+std::string osort::osort_pyobject::get_na_position() const {
```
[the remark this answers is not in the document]
pradkrish: Okay, let me give it a go.

pradkrish:
@st-pasha Alas, the test suite keeps failing here but they pass at my end. I have tested it on python 3.6, 3.7 and 3.8 on my ubuntu machine. Can you quickly check whether the test suite on this branch pass locally on your system? Thanks.

st-pasha:
@pradkrish 
1) I'd recommend merging branch `main` into your branch, so that it won't have to run tests for python 3.5;

2) Out of all github checks you should be able to see two: Travis and AppVeyor jobs. The latter is currently showing 8 failing tests, and also there are some build warnings (there is a small chance that they might be related).

3) Some of the tests in Jenkins fail with segfaults:
```
tests/test-dt.py::test_issue1728 Fatal Python error: Segmentation fault
```
However, while it says "(core file dumped)", there is no core file in the job artifacts. I'm not exactly sure why that is the case, perhaps the job configuration script needs to be tweaked...

pradkrish:
@st-pasha I didn't really combine reverse and na_position into a single SortFlag, I have kept them separate for now. I suggest we  create a separate task for it (I can do it). In that way, it will be self-contained and much smaller in scope.

src/core/sort.cc
```diff
@@ -756,15 +786,15 @@ class SortContext {
       dt::parallel_for_static(n,
         [&](size_t j) {
           TI t = xi[o[j]];
-          xo[j] = t == una? 0 :
+          xo[j] = t == una? replace_una :
```
st-pasha, line 789: Previously we had a map
```
NA -> 0
min -> 1
...
max -> (max-min + 1)
```
However, right now when NA is in last position the mapping is
```
min -> 1
...
max -> (max-min + 1)
NA -> (max-min + 2)
```
which can potentially overflow for a sufficiently wide range min..max. For example, try sorting an array that contains both `-MAX`, `MAX` and `NA`, where `MAX=2**32-1`.

src/core/expr/py_sort.h
```diff
@@ -34,13 +35,14 @@ class osort : public oobj
     private:
       oobj cols_;
       std::vector<bool>* reverse_;
+      std::vector<NaPosition>* na_position_;
```
st-pasha, line 38: Since it's stored as a naked pointer, don't forget to free it in `m__dealloc__()`

src/core/expr/py_sort.cc
```diff
@@ -83,6 +91,33 @@ void osort::osort_pyobject::m__init__(const PKArgs& args)
         "of booleans, instead got " << arg_reverse.typeobj();
   }
 
+  if (arg_na_position.is_none_or_undefined()) {
+    auto it = py::NaPositionValue.find("first");
+    na_position_ = new std::vector<NaPosition>(1, it->second);
```
st-pasha, line 96: it's easier to say `new std::vector<NaPosition>(1, NaPosition::FIRST);`

src/core/expr/py_sort.cc
```diff
@@ -25,7 +25,14 @@
 #include "utils/exceptions.h"
 namespace py {
 
-
+/**
+ *  Model type names and their corresponding dt::FtrlModelType's
+ */
```
st-pasha, line 30: I believe this comment came here by accident.

src/core/expr/py_sort.cc
```diff
@@ -25,7 +25,14 @@
...
+static const std::unordered_map<std::string, NaPosition> NaPositionValue {
+   {"first", NaPosition::FIRST},
+   {"last", NaPosition::LAST},
+   {"remove", NaPosition::REMOVE},
+};
```
st-pasha, line 35: wouldn't it be easier to have a simple static function
```
static NaPosition  get_na_position_from_string(const std::string& str) {
   return (str == "first")? NaPosition::FIRST :
          (str == "last")? NaPosition::LAST :
          (str == "remove")? NaPosition::REMOVE : NaPosition::INVALID;
}
```

src/core/expr/py_sort.cc
```diff
@@ -83,6 +91,33 @@ void osort::osort_pyobject::m__init__(const PKArgs& args)
...
+      auto it = py::NaPositionValue.find(na_position_list[i].to_string());
+      if (it == py::NaPositionValue.end()) {
+        throw ValueError() << "na position value `" << arg_na_position.to_string() << "` is not supported";
+      }
+      (*na_position_)[i] = it->second;
+    }
+  }
+  else {
+    throw TypeError() << arg_na_position.name() << " should be a boolean or a list "
+        "of booleans, instead got " << arg_na_position.typeobj();
```
st-pasha, line 118: ... should be one of 'first', 'last' or 'remove' ...

src/core/sort.cc
```diff
@@ -505,9 +505,12 @@ class SortContext {
     bool use_order;
     bool descending;
     int : 8;
+    NaPosition na_pos;
+    int count_nas;
```
st-pasha, line 509: this `count_nas` seem to be unused

tests/ijby/test-sort.py
```diff
@@ -1016,6 +1016,157 @@ def test_sort_double_negation():
...
+#-------------------------------------------------------------------------------
+
+def key_func(x, rev, na_pos):
+    if na_pos == "first":
+        return not rev if x is not None else rev
+    else:
+        return not rev if x is None else rev
+
+def sort_func(src, rev, na_pos):
+    res = sorted(src, key=lambda x: (key_func(x, rev, na_pos), x), reverse=rev)
```
st-pasha, line 1030: If you handle `na_pos="remove"` here, then the tests can be simplified (making them parametrized on the `na_position` argument):
```
if na_pos == "remove":
  src = [s for s in src if s is not None]
```

src/core/sort.cc
pradkrish: Good point, Wouldn't overflow happen when we map `max -> (max-min+1)` as well?

Just experimented with the following
```python
MAX = pow(2,32)-1
dt_int = dt.Frame(x=[0,1,None,MAX,None,-MAX,None])
dt_int[:, :, dt.sort(0, reverse=False, na_position="last")]
```
and got 
```code
  |           x
-- + -----------
 0 |  4294967295
 1 |           1
 2 |           0
 3 | -4294967297
 4 |          NA
 5 |          NA
 6 |          NA
```
Obviously, the column.stypes changed from `int32` to `int64` and It also happens when we map `max -> (max-min+1)` and `NA -> 0`

st-pasha: Ah, my bad. The largest value that can fit into int32 is `MAX = 2**31 - 1`. With this value, I'm getting (with your branch)
```
>>> dt_int[:, :, dt.sort(0, reverse=False, na_position="last")]
   |           x
-- + -----------
 0 |          NA
 1 |          NA
 2 |          NA
 3 | -2147483647
 4 |           0
 5 |           1
 6 |  2147483647

[7 rows x 1 column]
```

pradkrish: Yeah. So `Max` ends up getting mapped to `std::numeric_limits<unsigned int>::max()`. Anything above that is an overflow. do you have any suggestions? :worried:

st-pasha: Well, the formula that is computed here is basically
```
xo[j] = (x == NA)? 0 : 
          ASC? (x - min + 1) :
               (max - x + 1);
```
That "+1" makes it so that `x==min` maps into the value 1, which is right after the map value of NA (which is 0). But with the alternative placement of NAs, you'd have to modify this formula so that it says
```
xo[j] = (x == NA)? (max - min + 1) : 
          ASC? (x - min) :
               (max - x);
```
So you can either make a bigger nested if-condition, or maybe try to combine both cases into one. For example, write `(x - min + NAFIRST)` where `NAFIRST=(na_pos=='first')?1:0`.

tests/ijby/test-sort.py
```diff
@@ -1016,6 +1016,177 @@ def test_sort_double_negation():
...
+    EXP2 = dt.Frame(A=sort_func(src, rev, "last"))
+    EXP3 = dt.Frame(A=sort_func(src, rev, "remove"))
+
+    RES1 = DT[:, :, dt.sort(0, reverse=rev, na_position="first")]
+    RES2 = DT[:, :, dt.sort(0, reverse=rev, na_position="last")]
+    RES3 = DT[:, :, dt.sort(0, reverse=rev, na_position="remove")]
+
+    assert_equals(EXP1, RES1)
+    assert_equals(EXP2, RES2)
+    assert_equals(EXP3, RES3)
```
st-pasha, line 1189: I think all these tests can be combined into a single one that uses parametrization:

```suggestion
@pytest.mark.parametrize('rev', [True, False])
@pytest.mark.parametrize('napos', ['first', 'last', 'remove'])
@pytest.mark.parametrize('src', [[-5,-8,None,None,11,2,8,None,4]*1000,
                                 [-5.9,None,-8.3,11.5576,2.2,8.9,None,4.1]*1000,
                                 [True,None,False,None,False,True]*1000,
                                 ['',None,'pr',None,'','rww','auy','dfuy']*1000,
                                 [0,1,None,2**31-1,None,-(2**31-1),None]*1000,
                                 [0,1,None,2**63-1,None,-(2**63-1),None]*1000])
def test_sort_na_position(rev, napos, src):
    DT = dt.Frame(A=src)
    RES = DT[:, :, dt.sort(0, reverse=rev, na_position=napos)]
    EXP = dt.Frame(A=sort_func(src, rev, napos))
    assert_equals(RES, EXP)
```

st-pasha, line 1189: I think all these tests can be combined into a single one that uses parametrization:

```suggestion
@pytest.mark.parametrize('rev', [True, False])
@pytest.mark.parametrize('napos', ['first', 'last', 'remove'])
@pytest.mark.parametrize('src', [[-5,-8,None,None,11,2,8,None,4]*1000,
                                 [-5.9,None,-8.3,11.5576,2.2,8.9,None,4.1]*1000,
                                 [True,None,False,None,False,True]*1000,
                                 ['',None,'pr',None,'','rww','auy','dfuy']*1000,
                                 [0,1,None,2**31-1,None,-(2**31-1),None]*1000,
                                 [0,1,None,2**63-1,None,-(2**63-1),None]*1000])
def test_sort_na_position(rev, napos, src):
    DT = dt.Frame(A=src)
    RES = DT[:, :, dt.sort(0, reverse=rev, na_position=napos)]
    EXP = dt.Frame(A=sort_func(src, rev, napos))
    assert_equals(RES, EXP)
```

src/core/expr/py_sort.cc
```diff
@@ -102,6 +133,7 @@ void osort::osort_pyobject::m__dealloc__() {
   delete reverse_;
   reverse_ = nullptr;
   cols_ = nullptr;  // Releases the stored oobj
+  na_position_ = nullptr;
```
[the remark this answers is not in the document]
pradkrish: oh dear, very sloppy of me, thanks.

pradkrish:
Since this is also a user-visible change, should this be added to the release notes?

st-pasha:
Ah, absolutely! We only skip release notes for internal refactoring changes.

pradkrish:
lovely, in the `FExpr` section?

tests/ijby/test-sort.py
```diff
@@ -1016,6 +1016,36 @@ def test_sort_double_negation():
...
 
+#-------------------------------------------------------------------------------
+# Sort with positional value for NAs
+#-------------------------------------------------------------------------------
+
+def key_func(x, rev, na_pos):
+    if na_pos == "first":
+        return not rev if x is not None else rev
+    else:
+        return not rev if x is None else rev
```
st-pasha, line 1027: Simplifies into
```suggestion
    return (x is None) ^ (rev) ^ (na_pos == "first")
```

It's not a big deal of course, but such a rare opportunity to use operator XOR in real life, I thought it would be a shame to pass it up :)

pradkrish: I will grant your wish :)

st-pasha:
Yeah, FExpr section sounds appropriate

tests/ijby/test-sort.py
```diff
@@ -1016,6 +1016,33 @@ def test_sort_double_negation():
...
+    else:
+        return sorted(src, key=lambda x: (key_func(x, rev, na_pos), x), reverse=rev)
+
+
+@pytest.mark.parametrize('rev', [True, False])
+@pytest.mark.parametrize('napos', ['first', 'last', 'remove'])
+@pytest.mark.parametrize('src', [[-5,-8,None,None,11,2,8,None,4]*1000,
+                                 [-5.9,None,-8.3,11.5576,2.2,8.9,None,4.1]*1000,
+                                 [True,None,False,None,False,True]*1000,
+                                 ['',None,'pr',None,'','rww','auy','dfuy']*1000,
```
st-pasha, line 1038: Let's add the following src too: `['', None, '\x00', '\x01', '\x00'*5, None, '']*987`, just so we know NUL chars are sorted properly

st-pasha (changes requested):
Hi Pradeep,
Sorry that I didn't spot all the issues on my first review, but it looks like there may be some outstanding problems with string sorting. Luckily, they are easy to fix :)

A larger remaining issue is the treatment of list-valued `na_position=` - you can either try to make it work, or, if you prefer, comment-out the corresponding part in `py_sort.cc` and leave the support of this functionality for a future time.

pradkrish:
> Hi Pradeep,
> Sorry that I didn't spot all the issues on my first review, but it looks like there may be some outstanding problems with string sorting. Luckily, they are easy to fix :)

Thank you for taking the time to go through this PR even at an advanced stage. I really appreciate it, it only improves the quality of the PR.

> A larger remaining issue is the treatment of list-valued `na_position=` - you can either try to make it work, or, if you prefer, comment-out the corresponding part in `py_sort.cc` and leave the support of this functionality for a future time.

Since it is concrete and self-contained, I will leave the support for that functionality for a future time. Related to that, shouldn't the same thing be applicable for reverse as well, that is list-valued `reverse=`? is there a support for that currently?

src/core/expr/py_sort.cc
```diff
@@ -83,6 +88,32 @@ void osort::osort_pyobject::m__init__(const PKArgs& args)
...
+    if (na_pos == NaPosition::INVALID) {
+      throw ValueError() << "na position value `" << arg_na_position.to_string() << "` is not supported";
+    }
+    na_position_ = new std::vector<NaPosition>(1, na_pos);
+  }
+  else if (arg_na_position.is_list_or_tuple()) {
+    auto na_position_list  = arg_na_position.to_pylist();
+    na_position_ = new std::vector<NaPosition>(na_position_list.size());
+    for (size_t i = 0; i < na_position_->size(); ++i) {
+      NaPosition na_pos = get_na_position_from_string(arg_na_position.to_string());
```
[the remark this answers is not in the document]
pradkrish: Changed but commented-out this block as explained in the comment section.

src/core/sort.cc
```diff
@@ -873,10 +901,10 @@ class SortContext {
                            : 0xFE - static_cast<uint8_t>(*value.data());
```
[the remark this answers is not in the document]
pradkrish: Good catch, changed.

tests/ijby/test-sort.py
pradkrish: Yeah, added.

tests/ijby/test-sort.py
```diff
@@ -1016,6 +1016,33 @@ def test_sort_double_negation():
...
+                                 [0,1,None,2**31-1,None,-(2**31-1),None]*1000,
+                                 [0,1,None,2**63-1,None,-(2**63-1),None]*1000])
+def test_sort_na_position(rev, napos, src):
+    DT = dt.Frame(A=src)
+    RES = DT[:, :, dt.sort(0, reverse=rev, na_position=napos)]
+    EXP = dt.Frame(A=sort_func(src, rev, napos))
+    assert_equals(RES, EXP)
```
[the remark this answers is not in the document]
pradkrish: Added the first one. Perhaps, test 2 and test 3 can be added when we have list-valued na_position implemented?

src/core/sort.cc
```diff
@@ -809,6 +833,10 @@ class SortContext {
     allocate_x();
     TO* xo = x.data<TO>();
 
+    TO replace_na = na_pos == NaPosition::LAST ?
+                    static_cast<TO>(sizeof(TO) == 8 ? 0xFFFFFFFFFFFFFFFFULL : 0xFFFFFFFF)
```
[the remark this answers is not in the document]
pradkrish: I guess this comment is only a suggestion. If so, I will leave it the way it is.

st-pasha (approved):
Good work, Pradeep! I'll merge it now.

PR #2662, opened 2020-09-25
Fix git_revision/git_branch during a release

### LINKED ISSUE #2645: Make sure build_info has correct git_revision and git_branch during the release
With our current approach, the release ends up with `git_branch="HEAD"`  and `git_revision=<non-existing SHA>`. 
This is because the repository is checked out using 
```
git fetch origin +refs/pull/${env.CHANGE_ID}/merge
git checkout -qf FETCH_HEAD
```
which creates a new local "merge" commit that is detached from any branch.

We need to ensure that during the release we use proper `git checkout <branch>` instead.

In addition, the contents of `build_info` should be printed into the logs both in Jenkins and in AppVeyor.

st-pasha:
- Both in Jenkins & AppVeyor we now checkout the actual branch, so that during the release build scripts can report the correct branch / git commit.

- The contents of `_build_info.py` are printed to log.

- This branch's name starts with `rel-`, which should trigger the release process. However, since our release is fully manual right now, it should be sufficient to simply not continue with the release, and just delete this branch in the end.

Closes #2645


oleksiyskononenko (approved):

appveyor.yml
```diff
@@ -31,6 +31,7 @@ build_script:
 
     if ($env:APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH -match "^rel-") {
       $env:DT_RELEASE = "True"
+      git checkout $env:APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH$
```
oleksiyskononenko, line 34: It should be just `$env:APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH`, no need for the final `$` character.

appveyor.yml
```diff
@@ -90,6 +75,10 @@ build_script:
 
     $datatable_wheel = ls dist\*-cp36-*.whl
 
+    echo "----- build_info 3.6 ---------------------------------------------"
+    cat src/datatable/_build_info.py
+    echo "------------------------------------------------------------------"
```
oleksiyskononenko, line 80: Also, here, you might need to add new lines between every command, otherwise they will all stick together and produce an error.

appveyor.yml
oleksiyskononenko, line 34:
```suggestion
      git checkout $env:APPVEYOR_PULL_REQUEST_HEAD_REPO_BRANCH

```

appveyor.yml
oleksiyskononenko, line 80:
```suggestion
    echo "----- build_info 3.6 ---------------------------------------------"
    
    cat src/datatable/_build_info.py
    
    echo "------------------------------------------------------------------"

```

appveyor.yml
```diff
@@ -106,6 +95,10 @@ build_script:
 
     $datatable_wheel = ls dist\*-cp37-*.whl
 
+    echo "----- build_info 3.7 ---------------------------------------------"
+    cat src/datatable/_build_info.py
+    echo "------------------------------------------------------------------"
```
oleksiyskononenko, line 100:
```suggestion
    echo "----- build_info 3.7 ---------------------------------------------"
    
    cat src/datatable/_build_info.py
    
    echo "------------------------------------------------------------------"
    
```

appveyor.yml
```diff
@@ -122,6 +115,10 @@ build_script:
 
     $datatable_wheel = ls dist\*-cp38-*.whl
 
+    echo "----- build_info 3.8 ---------------------------------------------"
+    cat src/datatable/_build_info.py
+    echo "------------------------------------------------------------------"
```
oleksiyskononenko, line 120:
```suggestion
    echo "----- build_info 3.8 ---------------------------------------------"
    
    cat src/datatable/_build_info.py
    
    echo "------------------------------------------------------------------"
```

appveyor.yml
oleksiyskononenko, line 100:
```suggestion
    echo "----- build_info 3.7 ---------------------------------------------"
    
    cat src/datatable/_build_info.py
    
    echo "------------------------------------------------------------------"
```

appveyor.yml
st-pasha: you're right, they did end up stuck together:
```
# Build and test wheel for Python 3.6
# =======================================================================
$env:PATH = "C:\Python36-x64;C:\Python36-x64\Scripts;$DEFAULT_PATH"
python ci\ext.py wheel
$datatable_wheel = ls dist\*-cp36-*.whl
echo "----- build_info 3.6 ---------------------------------------------" cat src/datatable/_build_info.py echo "------------------------------------------------------------------"
pip --disable-pip-version-check install $datatable_wheel pytest docutils pandas
pytest .\tests\
```

appveyor.yml
st-pasha: fixed

appveyor.yml
oleksiyskononenko: Yeah, this only happens with multiline commands for some reason...

Issue #2604, opened 2020-09-04
Documentation checklist


samukweku:
The aim here is to have a checklist of documentation suggestions, and also to keep track, in case others have suggestions or are already working on one, they can reference from here : 

- [ ]  pandas vs datatable syntax comparison, showing how to replicate common pandas operations

- [ ]  rdatatable vs datatable syntax comparison, for users wishing to replicate familiar operations in R

- [ ]  sql vs datatable comparison, for sql users transiting to datatable

- [ ]  add examples to functions in docstrings; this will help users understand how to use the functions quickly, 

- [ ]  transformation documentation

- [ ]  documentation on joins

- [ ]  documentation on conditional operations

- [ ]  string operations in datatable

- [ ] cookbook

- [ ] updates to existing documentation (e.g fread ... more information on fread and excel (named ranges, ...)

Contributions/suggestions on what should be included as documentation are welcome

st-pasha:
Hi Samuel,
thanks for taking up this initiative. As you probably noticed, we've spent a significant effort lately to improve our documentation. We're currently mostly focused on documenting the API, but various tutorials / user guides might be just as useful if not more useful.
I've sent you an invite to join the project, so that you can create PRs directly (without going through the trouble of forking), and also better manage issues: assign labels, projects, milestones, etc.

samukweku:
Yes, I noticed documentation on the API. I was particularly happy to see the guide on creating and adding a function to datatable. Thanks for the opportunity @st-pasha

myamullaciencia:
Great Initiative @samukweku, I would choose one or two sections from the checklist and get back to yours here sooner. and i will let yours know if i need any help.

samukweku:
@myamullaciencia cool! Will be on the lookout for your PR

st-pasha:
Samuel, have you seen #2400 -- do you think it's something that can be incorporated inside your data.table/datatable comparison page?

samukweku:
@st-pasha It is a good idea. I will make a PR to cover that.

samukweku:
@st-pasha I noticed that datatable docs does not get updated when new docs are added. Same goes for the changelog. Any reasons for this? Maybe a design decision? I believe users (me included) would love to get access to new information.

st-pasha:
I've set up a webhook just now, so hopefully it'll start building automatically now.

samukweku:
@st-pasha No worries; I'll keep checking the site. At the moment, it has not updated.

[...]

PR #2665, opened 2020-09-28
Comparison between SQL and datatable

samukweku:
- Doc highlighting similar operations in SQL and `datatable`
- Also highlights some operations in SQL that are currently not possible in `datatable`

References #2604


st-pasha (commented):
Great work Samuel!
There were just few formatting issues related to underscores in table headers.

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
...
+
+Most of the examples will be based on the famous `iris <https://en.wikipedia.org/wiki/Iris_flower_data_set>`__ dataset. The assumption here is that the data is already loaded into a database; we'll simply show the relevant ``SQL`` queries side by side with ``datatable`` constructs.
+
+Let's import ``datatable`` and read in the data:
+
+.. code-block:: python
+
+    from datatable import dt, f, g, by, join, sort, update, fread
+
+    url = 'https://bit.ly/3kTNxM4'
```
oleksiyskononenko, line 16: I wouldn't use `bit.ly` here, because by looking at the link it is impossible to guess where it points to (without clicking on it) and also, if `bit.ly` goes down temporarily or permanently we might have a problem.

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
...
+    	4.9,	3,	1.4,	0.2,	setosa
+    	4.7,	3.2,	1.3,	0.2,	setosa
+    	4.6,	3.1,	1.5,	0.2,	setosa
+    	5,	3.6,	1.4,	0.2,	setosa
+
+
+SELECT
+-------
+
+In ``SQL``, you can select a subset of the columns with the ``SELECT`` clause:
```
oleksiyskononenko, line 36: What about mentioning a single-column selector `DT[j]` here?

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
...
+FILTER
+-------
+
+Filtering in ``SQL`` is done via the ``WHERE`` clause.
+
+.. code-block:: SQL
+
+    SELECT *
+    FROM iris
+    WHERE species = 'viriginica'
```
oleksiyskononenko, line 131: `viriginica` -> `virginica`?

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
...
+    SELECT *
+    FROM iris
+    WHERE species = 'setosa'
+    AND sepal_length = 5;
+
+In ``datatable`` each condition is wrapped in parentheses; the ``&`` operator is the equivalent of ``AND``, while ``|`` is the equivalent of ``OR``.
+
+.. code-block:: python
+
+    iris[(f.species=="setosa")&(f.sepal_length==5), :]
```
oleksiyskononenko, line 164: For better readability I would add some spaces at least around `&`, something like this may also work
```
iris[(f.species == "setosa") & (f.sepal_length == 5), :]
```

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
...
+
+    	0,	5.95263
+    	1,	5.77634
+
+At the moment, names cannot be assigned in the ``by`` section.
+
+LEFT OUTER JOIN
+----------------
+
+We will compare the left outer join, as that is the only join currently implemented in ``datatable``. Another aspect is that the dataframe being joined must be keyed, the column or columns to be keyed must not have duplicates, and the joining column has to have the same name in both dataframes. You can read more about the :func:`join()` API and have a look at the `Tutorial on the join operator <https://datatable.readthedocs.io/en/latest/start/quick-start.html#join>`_
```
oleksiyskononenko, line 512: I guess you can just say `frame` instead of a `dataframe` to be consistent with the rest of the docs we have now. It will also be more intuitive for the user to establish a relation `frame` <-> `dt.Frame()`, and not do something like `dt.dataframe()` that won' work...

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
...
+    	c,	9
+    	b,	7
+    	c,	8
+
+SQL's ``UNION`` removes duplicate rows after combining the results of the individual queries; there is no built-in function in ``datatable`` yet that handles duplicates.
+
+SQL's WINDOW FUNCTIONS
+----------------------
+
+Some SQL window functions can be replicated in datatable (`rank` is one of the windows function not currently implemented in datatable) :
```
oleksiyskononenko, line 597: I would add back-quotes to `datatable` here to be consistent with the rest of the content.

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,662 @@
+.. py:currentmodule:: datatable
+
+Comparison with SQL
+====================
+
+This page provides some examples of how various `SQL <https://en.wikipedia.org/wiki/SQL>`__ operations can be performed in ``datatable``. The ``datatable`` library is still growing; as such, not all functions in ``SQL`` can be replicated yet. If there is a feature you would love to have in ``datatable``, please make a feature request on the `github issues <https://github.com/h2oai/datatable/issues>`__ page.
```
oleksiyskononenko, line 6: Do we want to mention which particular SQL implementation we're discussing in this tutorial?

samukweku: The code I used is pretty generic and applies to all. I used SQLite by the way

oleksiyskononenko: Nice, if it works in SQLite it should work almost everywhere...

docs/manual/comparison_with_sql.rst
samukweku: Ok. I would rather download the file and store on the library's github page; however, I do not know the policy on file storage for datatable.

oleksiyskononenko: Also good idea, we can just put it under the `docs/_static`.

docs/manual/comparison_with_sql.rst
```diff
@@ -0,0 +1,782 @@
...
+
+Most of the examples will be based on the famous `iris <https://en.wikipedia.org/wiki/Iris_flower_data_set>`__ dataset. `SQLite <https://www.sqlite.org/>`__ will be the flavour of SQL used in the comparison.
+
+Let's import ``datatable`` and read in the data:
+
+.. code-block:: python
+
+    from datatable import dt, f, g, by, join, sort, update, fread
+
+    file_path = '/datatable/docs/_static/iris.csv'
```
oleksiyskononenko, line 16: I guess you can even do file_path = 'http://github.com/h2oai/datatable/docs/_static/iris.csv', then it will work even if there is no local `datatable` folder.

st-pasha (approved):

Issue #922, opened 2018-04-06
fread should not detect sep within quoted fields


st-pasha:
```
>>> dt.fread('"abc"\n"1,000"\n"2,000"\n')
     "1  000"
---  --  ----
 0   "2  000"

[1 row x 2 columns]
```

PR #2668, opened 2020-09-30
Double quoted strings

### LINKED ISSUE #491: Fread fails to parse file with a single unquoted string column
```
>>> dt.fread("../h2o-3/smalldata/parser/column.csv")
RuntimeError: Line 3 from sampling jump 0 starting "Rating (ou" has more than the 
expected 2 fields. Separator 2 occurs at position 12 which is character 5 of the last 
field: "(out of 10)". Consider setting 'comment.char=' if there is a trailing comment 
to be ignored.
```
This is the contents of the file:
```
ID
HELLO
Hello
Hello Kitty
Hello World
"Hercule Poriot"
Rating (out of 10)
Hello Goodbye by the Beatles
"Say hello, to my little friend"
Goodbye cruel world
"Ain't a thing like a chicken wing"
```

pradkrish:
"partially" Closes #491 

This branch allows treating entries with a double-quote as a single entity. For example, `fread`ing [apa.txt](https://github.com/Rdatatable/data.table/files/3478066/apa.txt) previously will end up as 2 columns because the double quoted string does not prevent the action of separator (space in this case).
```python
12 | M5EN3U5SW4  NA
 13 | LCUIX02FBP  NA
 14 | 0U5N3PO5R5  NA
  … | …           … 
 96 | "QXI04X6F1  " 
 97 | MD9167JW4F  NA
 98 | LCUIX02FBP  NA
 99 | "QXI04X6F1  " 
100 | 0U5N3PO5R5  NA
```
So logic has been added to prevent using separator until a matching double quote is found. It defaults to true for parsing unquoted strings. `fread`ing [apa.txt](https://github.com/Rdatatable/data.table/files/3478066/apa.txt) will now be
```python
12 | LCUIX02FBP
13 | 0U5N3PO5R5
14 | LCUIX02FBP
 … | …         
95 | QXI04X6F1 
96 | MD9167JW4F
97 | LCUIX02FBP
98 | QXI04X6F1 
99 | 0U5N3PO5R5
```
Tested also on [transformer_mojo.csv.zip](https://github.com/h2oai/datatable/files/3479240/transformer_mojo_input.csv.zip)

This logic broke a couple of test cases `test_runaway_quote()` and `test_unmatched_quotes()` in `tests/fread/test-fread-small.py`
They are both examples of `fread`ing strings with a single double quote. More logic is added to deal with this. After parsing a string that does not contain a matching double quote, parse the string from the beginning again, only this time using separator  and by skipping all double quotes.

### DIFF
--- a/src/core/read/parsers/parse_string.cc
+++ b/src/core/read/parsers/parse_string.cc
@@ -193,12 +193,17 @@ static void parse_string_unquoted(const ParseContext& ctx) {
   const char quote = ctx.quote;
   const char sep = ctx.sep;
   bool found_matching_quote = true;
+  bool single_quouted_string = false;
   if (ctx.strip_whitespace) {
     while (ch < end && *ch == ' ') ch++;
   }
   const char* field_start = ch;
   while (ch < end) {
     char c = *ch;
+    if (c == '\"' && single_quouted_string) {
+      ch++;
+      continue;
+    }
     if (c == '\"' && found_matching_quote) {
       found_matching_quote = false;
     }
@@ -217,6 +222,12 @@ static void parse_string_unquoted(const ParseContext& ctx) {
     else if (c == quote && QUOTES_FORBIDDEN) {
       ctx.target->str32.setna();
       return;
+    }
+    if (*(ch+1) == '\n' && !found_matching_quote) {
+      ch = ctx.ch;
+      found_matching_quote = true;
+      single_quouted_string = true;
+      continue;
     }
     ch++;
   }
--- a/src/core/read/parsers/parse_string.cc
+++ b/src/core/read/parsers/parse_string.cc
@@ -193,7 +193,6 @@ static void parse_string_unquoted(const ParseContext& ctx) {
   const char quote = ctx.quote;
   const char sep = ctx.sep;
   bool found_matching_quote = true;
-  bool use_sep = true;
   if (ctx.strip_whitespace) {
     while (ch < end && *ch == ' ') ch++;
   }
@@ -202,13 +201,11 @@ static void parse_string_unquoted(const ParseContext& ctx) {
     char c = *ch;
     if (c == '\"' && found_matching_quote) {
       found_matching_quote = false;
-      use_sep = false;
     }
     else if (c == '\"' && !found_matching_quote) {
       found_matching_quote = true;
-      use_sep = true;
     }
-    if (c == sep && use_sep) break;  // end of field
+    if (c == sep && found_matching_quote) break;  // end of field
     if (static_cast<uint8_t>(c) <= 13) {  // probably a newline
       if (c == '\n') {
         // Move back to the beginning of \r+\n sequence
--- a/src/core/read/parsers/parse_string.cc
+++ b/src/core/read/parsers/parse_string.cc
@@ -192,14 +192,23 @@ static void parse_string_unquoted(const ParseContext& ctx) {
   const char* end = ctx.eof;
   const char quote = ctx.quote;
   const char sep = ctx.sep;
-
+  bool found_matching_quote = true;
+  bool use_sep = true;
   if (ctx.strip_whitespace) {
     while (ch < end && *ch == ' ') ch++;
   }
   const char* field_start = ch;
   while (ch < end) {
     char c = *ch;
-    if (c == sep) break;  // end of field
+    if (c == '\"' && found_matching_quote) {
+      found_matching_quote = false;
+      use_sep = false;
+    }
+    else if (c == '\"' && !found_matching_quote) {
+      found_matching_quote = true;
+      use_sep = true;
+    }
+    if (c == sep && use_sep) break;  // end of field
     if (static_cast<uint8_t>(c) <= 13) {  // probably a newline
       if (c == '\n') {
         // Move back to the beginning of \r+\n sequence
--- a/src/core/read/parsers/parse_string.cc
+++ b/src/core/read/parsers/parse_string.cc
@@ -181,6 +181,13 @@ static void save_unescaped_string(const ParseContext& ctx,
   * advanced); if the flag is false then the quote chars are treated
   * as any other regular character.
   *
+  * found_matching_quote prevents using sep until a matching double quote
+  * is found. Defaults to true for parsing unquoted strings.
+  *
+  * After parsing a string that does not contain a matching double quote,
+  * parse the string from the beginning again using sep skipping all
+  * (unmatched) double quotes.
+  *
   * This function
   *   - WILL NOT check for NA strings;
   *   - WILL NOT check for UTF8 validity;
@@ -200,6 +207,7 @@ static void parse_string_unquoted(const ParseContext& ctx) {
   const char* field_start = ch;
   while (ch < end) {
     char c = *ch;
+    // skip all double quotes if it is a single quoted string
     if (c == '\"' && single_quouted_string) {
       ch++;
       continue;
@@ -210,6 +218,7 @@ static void parse_string_unquoted(const ParseContext& ctx) {
     else if (c == '\"' && !found_matching_quote) {
       found_matching_quote = true;
     }
+    // use sep only if found_matching_quote is true
     if (c == sep && found_matching_quote) break;  // end of field
     if (static_cast<uint8_t>(c) <= 13) {  // probably a newline
       if (c == '\n') {
@@ -223,6 +232,8 @@ static void parse_string_unquoted(const ParseContext& ctx) {
       ctx.target->str32.setna();
       return;
     }
+    // resume parsing the string from the beginning if matching
+    // double quote is not found
     if (*(ch+1) == '\n' && !found_matching_quote) {
       ch = ctx.ch;
       found_matching_quote = true;


pradkrish:
@st-pasha I figured out a way to get through the single double quote case I mentioned above. For example, I have a solution for
```python
d0 = dt.fread('"A,B,C\n1,2,3\n4,5,6')
```
But my logic fails for expressions with \r in it
```python
DT = dt.fread('A,"B\nC"\r5,7')
```
it is expected to give 
```python
dt.Frame({"A": [5], "B.C": [7]})
```
I only know \r means carriage return but I am struggling to understand what \r means in this context (I guess chat function could be very useful for these kind of discussions).

st-pasha:
Could you please explain the logic that you are trying to implement here? I.e. what kinds of inputs will now be parsed differently from before.

pradkrish:
@st-pasha I have updated the comment above with a better description of the PR. Let me know if it's unclear.

st-pasha:
I mean, we currently have 4 modes for string parsing:
- QR=0 is called "doubled". It accepts quoted strings where inner quote chars are doubled, such as `"abc"` or `"abc ""def"" gh"` or bare-word strings such as `hello` which cannot have quotes/separators inside them;
- QR=1 is called "escaped". It is similar to QR=0, but inner quotes are escaped with a backslash: `"abc \"def\" gh"`;
- QR=2 is "naive" where inner quotes are not escaped at all, so you'd see `"abc "def" gh"`. This one is an ugly duckling, but basically it will terminate a field when it sees a quote followed by a sep (or newline);
- QR=3 is "unquoted": the quotes are treated as if they were regular characters, the field always ends when a sep is encountered.

So, which of these rules are you modifying and how? Or do you want to add a new rule?

Looking at apa.txt, it appears to be perfectly parseable using current QR=0, so the issue is not really with the parser per se, but with the detection logic that tries to figure out which parse settings to use. This is what's reported in the verbose mode:
```
[2] Detect parse settings
  sep=',' with 2 lines of 5 fields using quote rule 3
  Detected 5 columns
  Quote rule = 3
  sep = ','
```

pradkrish:
That was my initial approach, given there were 4 parser settings, I thought something must be wrong with the detection logic. 

To work it out, I started by `fread`ing a csv file with a single entry `"Hello,there"` and printed some outputs at `src/core/csv/reader_fread.cc:295` and this is what I saw. 
```python
thisncol = 1, quoteRule = 0, sep = ,
thisncol = 1, quoteRule = 0, sep = |
thisncol = 1, quoteRule = 0, sep = ;
thisncol = 1, quoteRule = 0, sep = 	
thisncol = 1, quoteRule = 0, sep =  
thisncol = 1, quoteRule = 1, sep = ,
thisncol = 1, quoteRule = 1, sep = |
thisncol = 1, quoteRule = 1, sep = ;
thisncol = 1, quoteRule = 1, sep = 	
thisncol = 1, quoteRule = 1, sep =  
thisncol = 1, quoteRule = 2, sep = ,
thisncol = 1, quoteRule = 2, sep = |
thisncol = 1, quoteRule = 2, sep = ;
thisncol = 1, quoteRule = 2, sep = 	
thisncol = 1, quoteRule = 2, sep =  
thisncol = 2, quoteRule = 3, sep = ,  <------
thisncol = 1, quoteRule = 3, sep = |
thisncol = 1, quoteRule = 3, sep = ;
thisncol = 1, quoteRule = 3, sep = 	
thisncol = 1, quoteRule = 3, sep = 
```
The result should have been `(0 rows * 1 column)` but it was `(0 rows * 2 columns).` From this and a few more examples and also stepping through the code, I ended up thinking that every line ends up getting parsed by all quote rules and the number of columns present are inferred by taking into account the parsed result from every quote rule. 

This is why the logic in this branch was added in `parse_string_unquoted` (QR=3) so that it can handle even if a quoted string "sneaks in" (unbeknownst to me whether parsing every line through every quote rule is really a bug or a feature). This is how I understood, do correct me if I am wrong.

st-pasha:
So the problem is not that existing parsers cannot parse this input, but rather that we select incorrect parse settings among all possible choices.

Frankly, I don't quite like the current approach for parse setting detection anyways -- it's doing way too much work trying all the different combinations, and if we ever wanted to expand the parse settings to include new separators or some other parameters (like quote char), then the amount of work will increase exponentially.
A better way would be to treat it more the way humans do. After all, we don't go through all possible combinations in our heads.

pradkrish:
so how do you suggest we fix this particular issue? by improving the detection logic so that single- or double-quoted string never end up getting parsed by QR=3 at all? :thinking:

pradkrish:
How do you think humans parse through? Categorise by accumulating evidence cumulatively?

st-pasha:
One possible approach would be to scan the line, and tally the counts of special characters in each line. Then a true separator character will be the one that appears the same number of times in each line. However, this also needs to account for quoted fields (and different quote rules). And the `\s+` separator will need to be treated carefully.

This does not produce an explosion of parse attempts. For example, as you parse the line if you never encounter any `"` then the separator must be one of the special symbols you've seen so far (it's also possible to have a single-column input, in which case there should be no separator). If there was a `"` symbol, then it could either start a field (QR 0,1,2) or it could be a regular symbol (QR 3). In the latter case we parse as if there were no quoted fields. In the former case it's even better, because the character immediately before the quote must be the separator. The character after the field must also be the separator. So if those don't match then you know that that QR is invalid. If one of them is missing (because the field is first/last on the line), then you still know what the separator is. If both are missing (the field both starts and ends the line) then it must be a single-column file.

I think this logic is pretty close to how we (people) would try to parse the file if we were constrained to look at it in a sequential manner.

I even tried to implement something like this before (search for `allowedseps` in the code), but never finished.

pradkrish:
That way of parsing is definitely a good idea, I agree. It's a bigger task, is there an Epic for that ? 
In the meanwhile the bug remains and how do you wish to go about this PR?

st-pasha:
Regarding this PR, I'm worried about modifying the parse procedure that is used for parsing almost every field in every dataset. While the tests pass, it could be that it breaks some other datasets that are not in the test suit, particularly if the data contains quote marks. Consider the following dataset of TV sizes for example:
```
Size    Width   Height  Area(in²)
32"     27.9"   15.7"   438
40"     34.9"   19.6"   684
43"     37.5"   21.1"   791
50"     43.6"   24.5"   1068
55"     47.9"   27.0"   1293
60"     52.3"   29.4"   1538
```
(you could also check what would happen if spaces are replaced with commas or tabs).

I agree that the approach that I described may be a lot of work. A simpler approach would be to go into the current logic that does all combinations of parse parameters and tweak it so that a QR would be more important factor in choosing which parse parameters to select. For example, we can declare that if the input is consistently parseable under QR=0/1, then we don't even try any other QR rules.

I think that would fix a lot of these issues, which arise from the practice of using QR=3 where a smaller QR would work better.

pradkrish:
I agree that it should not break any existing dataset whether it's in test suit or not. I ran the example you gave by replacing spaces  with commas or tabs, it worked fine and I got a 6*4 column. In what way did you expect it to fail?

st-pasha:
Hmm, what if you try
```
dt.fread("""
  Size    Width   Height  Area
  "       "       "       sq.in. 
  32"     27.9"   15.7"   438
  40"     34.9"   19.6"   684
  43"     37.5"   21.1"   791
  50"     43.6"   24.5"   1068
  55"     47.9"   27.0"   1293
  60"     52.3"   29.4"   1538
""", verbose=True)
```

It seems that the original example selected QR=2, and your changes affect mostly QR=3.

pradkrish:
I get 
```code
   | Size  Width  Height  Area  
-- + ----  -----  ------  ------
 0 | "     "      "       sq.in.
 1 | 32"   27.9"  15.7"   438   
 2 | 40"   34.9"  19.6"   684   
 3 | 43"   37.5"  21.1"   791   
 4 | 50"   43.6"  24.5"   1068  
 5 | 55"   47.9"  27.0"   1293  
 6 | 60"   52.3"  29.4"   1538  

[7 rows x 4 columns]
```
The changes in this branch exclusively solves #922 leaving all else unaffected. That is why I think that this PR will do no more harm than what we already have on master.

st-pasha:
Weird, here's what i'm getting from your branch:
```
>>> import datatable as dt
>>> dt.fread("""
...   Size    Width   Height  Area
...   "       "       "       sq.in. 
...   32"     27.9"   15.7"   438
...   40"     34.9"   19.6"   684
...   43"     37.5"   21.1"   791
...   50"     43.6"   24.5"   1068
...   55"     47.9"   27.0"   1293
...   60"     52.3"   29.4"   1538
... """, verbose=True)
[*] Process input parameters
  Using default 12 thread(s)
  na_strings = ["NA"]
  strip_whitespace = True
  skip_blank_lines = False
Input contains '\n', treating it as raw text
[1] Prepare for reading
  ==== file sample ====
  
    Size    Width   Height  Area
    "       "       "       sq.in. 
    32"     27.9"   15.7"   438
    40"     34.9"   19.6"   684
  =====================
Skipped 3 initial whitespace character(s)
LF character (\n) found in input, \r-only newlines will be prohibited
[2] Detect parse settings
  sep=' ' with 6 lines of 4 fields using quote rule 2
  sep=' ' with 7 lines of 3 fields using quote rule 3
  sep=' ' detected, setting fill to True
  Detected 4 columns
  Quote rule = 3
  sep = ' '
[3] Detect column types and header
  Number of sampling jump points = 1 because input is less than 100 lines
  Type codes (jump 0): sss?
  Type codes  (final): sss?
  `header` determined to be True because all inputs columns are strings and better guess is not possible
  =====
  Sampled 7 rows at 1 jump point(s)
  Bytes from first data row to the end of last row: 246
  Line length: mean=31.00 sd=1.41 min=30 max=34
  Estimated number of rows: 8
  Initial alloc = 8 rows (using bytes/max(mean-2*sd,min) clamped between [1.1*estn, 2.0*estn])
  All rows were sampled since file is small so we know nrows=6 exactly
[4] Assign column names
[5] Apply user overrides on column types
  Allocating 4 column slots with 6 rows
[6] Read the data
  Number of threads reduced to 1 because data is small
  The input will be read in 1 chunk of size 217 each
  Too few rows allocated, reallocating to 7 rows
[7] Finalizing the frame
=============================
Read 7 rows x 4 columns from 246 bytes input in 00:00.002s
 = 0.000s ( 0%) memory-mapping input file
 + 0.000s ( 8%) detecting parse parameters
 + 0.000s (11%) detecting column types using 6 sample rows
 + 0.001s (67%) allocating [6 x 4] frame (845 bytes) of which 7 (117%) rows used
 + 0.000s ( 9%) reading data
    = 0.000s ( 1%) reading into row-major buffers
    = 0.000s ( 0%) saving into the output frame
    = 0.000s ( 8%) waiting
    = 0.000s ( 4%) creating the final Frame
   | Size           Width  Height  Area
-- + -------------  -----  ------  ----
 0 | "       "      "      sq.in.    NA
 1 | 32"     27.9"  15.7"  438       NA
 2 | 40"     34.9"  19.6"  684       NA
 3 | 43"     37.5"  21.1"  791       NA
 4 | 50"     43.6"  24.5"  1068      NA
 5 | 55"     47.9"  27.0"  1293      NA
 6 | 60"     52.3"  29.4"  1538      NA

[7 rows x 4 columns]
```

Can you check what your verbose log shows to see where the discrepancy comes from?

pradkrish:
my bad, I may not have checked properly before, I get the same as you now. I see now why it fails and I understand why you are not on board with this approach. It's like you have got a secret validation dataset that can break PRs. :) 

Quoting you from above *For example, we can declare that if the input is consistently parseable under QR=0/1, then we don't even try any other QR rules*, I have created an experimental branch `sep_inside_quotes_922`, can you check whether the changes on that branch is along the lines of what you suggested?

PR #2670, opened 2020-10-01
New sphinx domain: xpy

### LINKED ISSUE #2627: Improve rendering of links in docs
Currently a simple code such as
```
:class:`Frame`
```
will generate the link https://datatable.readthedocs.io/en/latest/api/frame.html#datatable.Frame (note the anchor), which will go to the correct page, but will scroll down past the page header.

~~In order to fix this problem we'd have to hook up the [event-missing-reference](https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-missing-reference) Sphinx event and return the correct references ourselves.~~

st-pasha:
This domain is designed for all our python functions/classes/etc. that we want to document as our API. In particular, the `..xfunction` directive registers targets within this domain, and then reference roles such as `:func:` link to objects in this domain. This domain is declared as default in project settings.

The reason for this new domain is because the original python domain was designed for documenting python, where they attempt to document multiple entities per page. In contrast, we follow the same approach as pandas / numpy / other large projects: dedicate a separate page for each documented object. As such, a reference target should be the page itself, not an anchor within the page.

The second problem is that with its lax rules on targets, it's too easy to create a target that points to an unexpected place (especially when using Intersphinx). With `xpy:` domain the roles for referencing external targets have `ext-` in their name, e.g. `:ext-func:'repr()'`, making it clear that we are referencing an external object there.

The rules for forming the links in this domain are as follows:
- there is a notion of "current context" which determines where the non-explicit links belong to;
- certain directives set up current context automatically:
    - `.. xpy-module` puts the page into the context of the declared module;
    - `.. xclass` sets the context to the name of the class;
    - `.. xfunction` and `.. xdata` sets the context to the function's module;
    - `.. xattr` and `.. xmethod` set the context to the name of their class;
- within the class context a dot-link will refer to that class. For example, `:attr:'.nrows'` refers to `attr:datatable.Frame.nrows` within the context of class `datatatable.Frame`;
- within the module context a bareword name refers to the member of the class. For example, `:func:'abs()'` refers to `func:datatable.math.abs` within the context of module `datatable.math`;
- all other types of links must be fully-qualified. For example, if you are in the context of class `dt.Frame` and want to refer to top-level function `rbind()` you have to write `:func:'dt.rbind()'`;
- anywhere you need to write a fully-qualified name, using `dt.` instead of `datatable.` should work.

Also, if any reference is missing or somehow malformed, it will now complain in loud red letters in the build log (but won't stop the build process).

Closes #2627


oleksiyskononenko:
I have noticed, that on the attribute pages we have some problems with rendering the variables. For instance, on the screenshot below `nthreads` variable is not highlighted, so does the `new_nthreads`. Do you think this PR will fix this?

Screen Shot 2020-10-01 at 9 41 14 PM

st-pasha:
You can merge this PR into your branch and see if it changes anything. However, this problem that you're describing can be caused by something else.

oleksiyskononenko:
@st-pasha merged, it didn't help... seems like we just don't parse attribute pages correctly.

Issue #2671, opened 2020-10-01
Improve options documentation


oleksiyskononenko:
We need to add all the default values, links to the source code for getters/setters, exceptions, etc.

PR #2673, opened 2020-10-01
Improve options documentation

oleksiyskononenko:
In this PR we

- create separate getters/setters and the corresponding links, improve documentation, include information about exceptions, default values, etc;
- remove `FreadWarning` replacing it with `IOWarning`;
- add a page on datatable exceptions, that we should expand in future.

Closes https://github.com/h2oai/datatable/issues/2671


src/core/frame/names.cc
```diff
@@ -353,11 +353,22 @@ void Frame::_init_names(XTypeMaker& xt) {
...
+Parameters
+----------
+return: int
+    Current `names_auto_index` value. Initially, this option is set to `0`.
+
+new_names_auto_index: int
+    New `names_auto_index` value.
+
+except: TypeError
+    The exception is raised when the type of `new_names_auto_index` is not `int`.
```
st-pasha, line 371: This exception is raised for every single argument of every single function if you pass the incorrect type. So, I'm not sure this is something even worth documenting.

oleksiyskononenko: That's true, not sure if it hurts though... especially that we already have it in many places.

st-pasha: Well, it creates a source of inconsistency: if we document this for one function but not for other, then it looks like the other function does not check the types of its arguments?

oleksiyskononenko: Well, if it looks like that then it's not because it is documented here, as here we do not report anything wrong, but because it is not documented for the other functions or for datatable in general. I feel like we need to create a dedicated page for all the possible types of exceptions/warnings datatable may produce and then remove all the individual `TypeError`s of this type everywhere in the docs.

docs/api/exceptions.rst
```diff
@@ -0,0 +1,39 @@
+
+.. xdata:: datatable.exceptions
+    :src: --
+
+    This namespace contains warnings and exceptions that ``datatable`` may
+    throw during runtime.
+
+    Warnings are issued when it is helpful to inform the user of some condition
+    in a program, that doesn't result in an exception and the program termination.
+    Along with the built-in Python `warnings <https://docs.python.org/3/library/warnings.html>`_
```
st-pasha, line 10: The only builtin warning that we issue is `FutureWarning` about deprecated features.

docs/api/exceptions.rst
```diff
@@ -0,0 +1,39 @@
...
+    `exceptions <https://docs.python.org/3/library/exceptions.html>`_
+    ``datatable`` may also throw:
+
+    .. list-table::
+        :widths: auto
+        :class: api-table
+
+        * - `InvalidOperationError`
+          - Thrown when the operation requested is illegal
+            for the given combination of parameters.
```
st-pasha, line 38: There are many more exceptions, see `exceptions.py`.

docs/api/exceptions.rst
```diff
@@ -0,0 +1,39 @@
+
+.. xdata:: datatable.exceptions
```
st-pasha, line 2: `datatable.exceptions` is a module, not a variable. Similar to `datatable.math` or `datatable.models`.

st-pasha (commented):
Generally, we'd want the exceptions to be one docpage per object, so that cross-references could work properly.

docs/api/exceptions.rst
oleksiyskononenko: Yes, but they're all the same as the built-in Python exceptions.

docs/api/exceptions.rst
oleksiyskononenko: Yes, but what I mean here is that we either issue a warning that is built-in, so no need to add it to the docs, or we issue something specific to datatable, then we add it to the docs.

docs/api/exceptions.rst
```diff
@@ -0,0 +1,39 @@
...
+
+    .. list-table::
+        :widths: auto
+        :class: api-table
+
+        * - ``DatatableWarning``
+          - Generic ``datatable`` warnings;
+
+        * - ``FreadWarning``
+          - Warnings related to :func:`fread`;
```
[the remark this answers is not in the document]
oleksiyskononenko: I also didn't realize why we have these two separate...

oleksiyskononenko:
Well, having a page per a warning/exception seems not reasonable as it is just a tiny piece of information. For a cross-reference it seems that it is better to just use the same URL with the anchor specifying a particular warning/exception.

docs/api/exceptions.rst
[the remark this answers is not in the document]
st-pasha: probably just a leftover from some ancient times

docs/api/exceptions.rst
oleksiyskononenko: Yeah, typo. So how do we define modules now? `.. py:module:: datatable.exceptions` or `.. xpy:module:: datatable.exceptions`. For some reason we have both now for math and models.

docs/api/exceptions.rst
st-pasha: They are not exactly the same: they inherit from the builtin exception with the same name, and also from the common `DtException`, which means that you can catch any `datatable` error by catching `DtException`.

Plus, there's still advantage in listing all exceptions that datatable can throw, and describing general circumstances when different exceptions may occur. For example, you mentioned that we should document the fact that argument types for each function are strictly enforced. The perfect place for that is the docpage for our `TypeError`.

oleksiyskononenko: >which means that you can catch any datatable error by catching DtException

This is definitely worth to mention. But listing all the exceptions with the same names as the built-in ones seems like just reiteration of the standard python docs. For instance, `TypeError` has exactly the same meaning in `datatable` and in python as explained here https://docs.python.org/3/library/exceptions.html#TypeError

PR #2694, opened 2020-10-13
Proofread documentation

### LINKED ISSUE #2655: Proofread documentation
Go thru documentation and proofread.

"This is the link to the datatable source code: https://github.com/h2oai/datatable/tree/main/ and this is the documentation as it is rendered now on readthedocs: https://datatable.readthedocs.io/en/latest/

We have recently added quite a lot of information to the API section and expanded the others. It is kept in the RST format (reStructuredText) in the docs folder here: https://github.com/h2oai/datatable/tree/main/docs

Some of the .rst files have links to the C++ source code that is stored under the src/core folder: https://github.com/h2oai/datatable/tree/main/src/core It has a lot of function/class/option-specific information in the doc_ variables. Here is an example https://github.com/h2oai/datatable/blob/main/src/core/set_funcs.cc#L152-L193

The RST format is readable enough even for those who are not familiar with the format itself. So I would propose just to go through the text on readthedocs, once you see something that is wrong and needs a correction — you simply search for this particular phrase in the code and fix it in-place. If it is something more complicated than just a text or the search cannot find the relevant place in the code, you just contact me or pasha for a help."

hannah-tillman:
Primarily minor syntax updates. Combed through all of the docs. Please let me know if you need me to update anything!

Closes #2655

### DIFF
--- a/src/core/models/aggregate.cc
+++ b/src/core/models/aggregate.cc
@@ -51,7 +51,7 @@ grouping for string columns.
 
 When the input frame has more columns than two,
 a parallel one-pass Ad-Hoc algorithm is employed, see description of
-`Aggregator<T>::group_nd() <https://github.com/h2oai/datatable/blob/main/src/core/models/aggregate.cc#L879>`_
+`Aggregator<T>::group_nd() <https://github.com/h2oai/datatable/blob/00185546f033cc9d2fca202b516a1720b85581f5/src/core/models/aggregate.cc#L878-L1024>`_
 method for more details. This algorithm takes into account
 the numeric columns only, and all the string columns are ignored.
 
@@ -111,7 +111,7 @@ return: Tuple[Frame, Frame]
 
     The second element in the tuple is the members frame with the shape of
     `(frame.nrows, 1)`. Each row in this frame corresponds to the
-    row with the same id in the input `frame`. Only the column `exemplar_id`
+    row with the same id in the input `frame`. The single column `exemplar_id`
     has an stype of `int32` and contains the exemplar ids that a particular
     member belongs to. These ids are effectively the ids of
     the exemplar's rows from the input frame.
--- a/docs/manual/f-expressions.rst
+++ b/docs/manual/f-expressions.rst
@@ -168,7 +168,7 @@ columns into the columnset as well::
 When a columnset is extended, the order of the elements is preserved. Thus, a
 columnset is closer in functionality to a python list than to a set. In
 addition, some of the elements in a columnset can have names if the columnset
-is created from a dictionary. The names may be non-unique, too.
+is created from a dictionary. The names may be non-unique too.
 
 The ``.remove()`` method is the opposite of ``.extend()``: it takes an existing
 columnset and then removes all columns that are passed as the argument::
--- a/docs/start/quick-start.rst
+++ b/docs/start/quick-start.rst
@@ -127,7 +127,7 @@ expression can even be a python type (such as ``int`` or ``dt.float32``),
 selecting all columns matching that type.
 
 In addition to the selector expression shown above, we support the update and
-delete statements, too:
+delete statements too:
 
 ::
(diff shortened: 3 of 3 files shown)


st-pasha (approved):
Thanks, Hannah!

I was only wondering about commas before "too"s. According to [this blog post](https://www.grammarly.com/blog/comma-before-too/) they aren't really needed. Do you have a different perspective on this?

hannah-tillman:
> Thanks, Hannah!
> 
> I was only wondering about commas before "too"s. According to [this blog post](https://www.grammarly.com/blog/comma-before-too/) they aren't really needed. Do you have a different perspective on this?

Ah, it was the way I was taught in grade school and kind of just stuck to it--I can undo those changes if you'd prefer :)

src/core/models/aggregate.cc
```diff
@@ -51,7 +51,7 @@ grouping for string columns.
 
 When the input frame has more columns than two,
 a parallel one-pass Ad-Hoc algorithm is employed, see description of
-`Aggregator<T>::group_nd() <https://github.com/h2oai/datatable/blob/main/src/core/models/aggregator.cc>`_
+`Aggregator<T>::group_nd() <https://github.com/h2oai/datatable/blob/main/src/core/models/aggregate.cc#L879>`_
```
oleksiyskononenko, line 54: Nice catch, I guess we can even make it a link like to a particular commit (the last one in this case) `https://github.com/h2oai/datatable/blob/00185546f033cc9d2fca202b516a1720b85581f5/src/core/models/aggregate.cc#L878-L1024` so that if we have a minor change in the source code, we don't need to worry about the docs. If we have a major change, I will update this link.

src/core/models/aggregate.cc
```diff
@@ -110,9 +110,9 @@ return: Tuple[Frame, Frame]
     number of members per exemplar.
 
     The second element in the tuple is the members frame with the shape of
-    `(frame.nrows, 1)`, each row in this frame corresponds to the
-    row with the same id in the input `frame`. The only column `exemplar_id`
-    has an stype of `int32` and contains the exemplar ids a particular
+    `(frame.nrows, 1)`. Each row in this frame corresponds to the
+    row with the same id in the input `frame`. Only the column `exemplar_id`
+    has an stype of `int32` and contains the exemplar ids that a particular
```
oleksiyskononenko, line 115: Here I wanted to say that we only have one column in this frame. I guess it is better to write it as "The single column `exemplar_id`...".

src/core/models/aggregate.cc
st-pasha: Why not just transfer the description from code into the docs?

st-pasha:
> Ah, it was the way I was taught in grade school 

Could it be that your school teacher was mistaken? The article on Grammarly refers to the _Chicago Manual of Style_, which sounds like a more authoritative source.

hannah-tillman:
> > Ah, it was the way I was taught in grade school
> 
> Could it be that your school teacher was mistaken? The article on Grammarly refers to the _Chicago Manual of Style_, which sounds like a more authoritative source.

Perhaps--I'll go back thru and undo those changes

src/core/models/aggregate.cc
oleksiyskononenko: That description was for the developer use only, so I would prefer to revisit it and make the algo more clear for the general user first.

oleksiyskononenko (approved):

src/core/models/aggregate.cc
st-pasha: Yeah, I understand that. But since we are mentioning it explicitly in the public docs, then it kinda no longer is "developer only".

oleksiyskononenko: Yep, that link was a temporary solution... I will make a separate PR to improve the aggragator docs at some point.

Issue #2625, opened 2020-09-16
Improve style of doc page on R/Python comparison


st-pasha:
The documentation page  https://datatable.readthedocs.io/en/latest/manual/comparison_with_rdatatable.html may need custom table style / formatting, in order to allow multi-line code samples and. Also the table could be converted into 2-column style so that it would fit on the page easier. 

This issue stems from the discussion in #2611. The following "syntax comparison" pages were noted as examples:
  - https://datascience-enthusiast.com/R/pandas_datatable.html
  - https://towardsdatascience.com/python-pandas-vs-scala-how-to-handle-dataframes-part-ii-d3e5efe8287d
  - https://pandas.pydata.org/docs/getting_started/comparison/comparison_with_r.html
  - https://atrebas.github.io/post/2019-03-03-datatable-dplyr/ (:star:)

And this was suggested as a possible style: 
<img width="736" alt="Screen Shot 2020-09-15 at 2 11 39 PM" src="https://user-images.githubusercontent.com/4231472/93265630-bf328600-f75d-11ea-81e2-89c9191b1602.png">

samukweku:
@st-pasha how do I get the style suggested above, so I can use it for documentation on comparison with Pandas. thanks

samukweku:
@st-pasha @oleksiyskononenko We could use [sphinx-panels](https://sphinx-panels.readthedocs.io/en/latest/) to improve the documentation here; make it look like a card layout.  Used in [jupyter-book](https://jupyterbook.org/content/content-blocks.html#panels) as well; might be helpful.

PR #2696, opened 2020-10-15
New theme for documentation

st-pasha:
Trying to create a new theme for documentation, which would provide a cleaner and more modern look.


oleksiyskononenko:
The "Table of Contents" menu seems to be buggy. When I click an item there it shows me that item, but on the menu it selects the previous one.

oleksiyskononenko:
In some case when I click on the item on the left, the right menu jumps to the last item immediately:
Screen Shot 2020-10-20 at 4 16 21 PM

oleksiyskononenko:
Submodule functions, like for math or models submodules, are rendered again black for some reason and are not clickable anymore:

Screen Shot 2020-10-20 at 4 18 20 PM

st-pasha:
Re 1: fixed
Re 2: the side menu shows current section based on the scroll position of the page. If the sections are too short so that all of them are on the same page, then there is no really wrong answer here.

oleksiyskononenko:
Re 2: Yes, but it is more intuitive to show the first section that is available on the page rather than the last one.

oleksiyskononenko:
Btw, the module link is still broken and black:

Screen Shot 2020-10-26 at 11 38 33 AM

oleksiyskononenko:
Also, I'm not sure if this is only my local problem, but the `csv-table` is not rendered well in the new theme:

Screen Shot 2020-10-26 at 1 00 54 PM

st-pasha:
> Btw, the module link is still broken and black:

This is broken in our current documentation too, so not related to this PR.

> csv-table is not rendered well in the new theme:

Indeed, the styles for csv table are missing. However, we already have an issue to fix styling of these "syntax comparison" pages (#2625), so we can consider this problem as part of that issue.

oleksiyskononenko:
>This is broken in our current documentation too, so not related to this PR.

Yes, I thought it was `1.1` that you said was fixed.

PR #2697, opened 2020-10-15
Delete rows from an unsorted array

### LINKED ISSUE #805: Allow deleting rows from an unsorted array
Currently we can delete rows, using an array of integers as a selector. However that array must be sorted, otherwise an exception will be thrown. This issue asks to support deleting an unsorted list of rows too -- probably by sorting and de-duplicating them.

pradkrish:
Closes #805 

The rows to be deleted can be an unsorted array.
The array can contain duplicate values. 
Size of the array to be deleted can be larger than the size of the frame as long as the index to be deleted does not exceed the number of rows on the frame.

### DIFF
--- a/src/core/rowindex_array.cc
+++ b/src/core/rowindex_array.cc
@@ -31,7 +31,6 @@
 #include "utils/assert.h"
 #include "utils/macros.h"
 #include "stype.h"
-#include <algorithm>
 #if DT_DEBUG
   inline static void test(ArrayRowIndexImpl* o) {
     o->refcount++;
@@ -346,7 +345,7 @@ RowIndexImpl* ArrayRowIndexImpl::negate_impl(size_t nrows) const
       outputs[k++] = i;
     }
   }
-
+  xassert(nrows >= count_indices_to_skip);
   int flags = RowIndex::SORTED;
   flags |= (sizeof(TO) == sizeof(int32_t))? RowIndex::ARR32 : RowIndex::ARR64;
   outbuf.resize((newsize - count_indices_to_skip) * sizeof(TO));
@@ -355,7 +354,6 @@ RowIndexImpl* ArrayRowIndexImpl::negate_impl(size_t nrows) const
 
 
 RowIndexImpl* ArrayRowIndexImpl::negate(size_t nrows) const {
-  xassert(nrows >= length);
   if (type == RowIndexType::ARR32) {
     if (nrows <= INT32_MAX)
       return negate_impl<int32_t, int32_t>(nrows);
--- a/src/core/rowindex_array.cc
+++ b/src/core/rowindex_array.cc
@@ -31,7 +31,7 @@
 #include "utils/assert.h"
 #include "utils/macros.h"
 #include "stype.h"
-
+#include <algorithm>
 #if DT_DEBUG
   inline static void test(ArrayRowIndexImpl* o) {
     o->refcount++;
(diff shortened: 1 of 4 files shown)


src/core/rowindex_array.cc
```diff
@@ -314,40 +313,47 @@ void ArrayRowIndexImpl::compactify()
...
-  size_t newsize = nrows - length;
+  size_t newsize = nrows;
   size_t inpsize = length;
   Buffer outbuf = Buffer::mem(newsize * sizeof(TO));
   auto outputs = static_cast<TO*>(outbuf.xptr());
   TO orows = static_cast<TO>(nrows);
 
   TO next_index_to_skip = static_cast<TO>(inputs[0]);
-  size_t j = 1;  // next index to read from the `inputs` array
+  size_t count_indices_to_skip = 0;
```
st-pasha, line 336: This seems more like `count_indices_skipped`

src/core/rowindex_array.cc
```diff
@@ -314,40 +313,47 @@ void ArrayRowIndexImpl::compactify()
...
   size_t k = 0;  // next index to write into the `outputs` array
   for (TO i = 0; i < orows; ++i) {
     if (i == next_index_to_skip) {
-      next_index_to_skip =
-        j < inpsize? static_cast<TO>(inputs[j++]) : orows;
-      if (next_index_to_skip <= i) {
-        throw ValueError() << "Cannot invert RowIndex which is not sorted";
-      }
+      ++count_indices_to_skip;
+      get_next_index_to_skip(inputs, j);
```
st-pasha, line 342: The name of this function does not reflect what it does. `get_smth()` ought to be returning the value `smth`. However, in your case the function is about advancing `j` to the correct position. So i'd suggest to either (1) rename the function, or (2) inline it, or (3) move calculation of `next_index_to_skip` into the function and return that value.

src/core/rowindex_array.cc
```diff
@@ -314,40 +313,47 @@ void ArrayRowIndexImpl::compactify()
   _resize_data();
 }
 
+template<typename TI>
+void get_next_index_to_skip(const TI* inputs, size_t& j)
+{
+  while (inputs[j] == inputs[j+1]) {
```
st-pasha, line 319: This doesn't check that `inputs[j+1]` is within the array, so it may segfault for certain inputs.

I'd suggest to add a test where the user deletes a list of all identical indices (e.g. `[3]*1000`). This could be parametrized for lists of various lengths.

tests/munging/test-delete.py
```diff
@@ -316,14 +316,27 @@ def test_del_rows_array():
...
+    assert d0.to_list() == [[2, 5, 6, 7, 8]]
+
+
+def test_del_rows_array_unordered_larger_than_frame():
+    d0 = dt.Frame(range(10))
+    del d0[[5, 6, 5, 8, 5, -1, 3, 0, 0, 3, -1, 4, 1, 1], :]
+    frame_integrity_check(d0)
+    assert d0.to_list() == [[2, 7]]
+
+
```
st-pasha, line 339: I would also suggest to add the following tests: 
- delete a one-element list of rows (e.g. `[3]`);
- delete an empty list of rows (`[]`);
- delete a list of rows, possibly with repetitions, such that all rows are removed from the frame (e.g. `[3, 2, 0, 2, 0, 1, 4, 4, 0]` with a 5-rows frame);
- delete a long list of identical rows (e.g. `[3]*1000`), see my comment above.

st-pasha (changes requested):
Thanks, Pradeep. Since this is a user-visible change, please also add a mention in the Release Notes file `docs/releases/v1.0.0.rst`, probably in the "Frame" section.

src/core/rowindex_array.cc
pradkrish: good point, I agree.

src/core/rowindex_array.cc
pradkrish: Agree

src/core/rowindex_array.cc
pradkrish: I agree, no need to move it into a separate function. I have brought the computation back into the calling function.

tests/munging/test-delete.py
pradkrish: I agree

st-pasha (approved):

PR #2706, opened 2020-10-24
Allow str64 in a call to fread

### LINKED ISSUE #2704: [bug?] can't asign column type in fread function using at 0.11.0
dttype = [dt.str32, dt.str32, dt.int64, dt.str32, dt.str64, dt.str32]
ddf = dt.fread('/data/zxdata/mon_zx20201022.txt',columns = dttype)

**in 0.10.1:** 
It's correct.
**in 0.11.0:** 
There is a Error!
RuntimeError: Unknown column SType in fread

st-pasha:
I made it so str64 falls back into str32, because implementing a new stype for fread seems excessive at this moment.

Closes #2704

### DIFF
--- a/src/core/read/input_column.cc
+++ b/src/core/read/input_column.cc
@@ -105,7 +105,7 @@ void InputColumn::set_rtype(int64_t it) {
     case RFloat64: parse_type_ = PT::Float64Plain; break;
     case RStr:     parse_type_ = PT::Str32; break;
     case RStr32:   parse_type_ = PT::Str32; break;
-    case RStr64:   parse_type_ = PT::Str64; break;
+    case RStr64:   parse_type_ = PT::Str32; break;  // ?
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/core/read/input_column.cc
```diff
@@ -105,7 +105,7 @@ void InputColumn::set_rtype(int64_t it) {
     case RFloat64: parse_type_ = PT::Float64Plain; break;
     case RStr:     parse_type_ = PT::Str32; break;
     case RStr32:   parse_type_ = PT::Str32; break;
-    case RStr64:   parse_type_ = PT::Str64; break;
+    case RStr64:   parse_type_ = PT::Str32; break;  // ?
```
pradkrish, line 108: Shouldn't a note like (see #2704 ) be added here? In this way, anyone who is looking at the code will have a context for this.

st-pasha: done

PR #2707, opened 2020-10-24
Docs style small fixes

RaffaeleMorganti:
I think these small modification can improve doc usability

- Added a small scrollbar in the left menu

old | new
---|---
|

- Index styles modified to prevent elements overlay

old | new
---|---
|

### DIFF
--- a/docs/_static/code.css
+++ b/docs/_static/code.css
@@ -59,6 +59,12 @@ div.rst-content .admonition-title {
 
 /* ------- Side menu adjustments -------------------------------------------- */
 
+@media (min-width:768px) {
+  .wy-side-scroll {
+    width: 310px;
+  }
+}
+
 .wy-menu-vertical {
   margin-bottom: 48pt;
 }
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,36 +1,29 @@
 
+|logo|Datatable
+================
+.. |logo| image:: _static/py_datatable_logo.png
+          :width: 80
+          :class: no-scaled-link
+
+**Datatable** is a python library for manipulating tabular data.
+It supports out-of-memory datasets, multi-threaded data processing,
+and flexible API.
+
 .. raw:: html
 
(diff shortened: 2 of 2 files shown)


pradkrish:
I like the introduction of scroll bar.  The overlay does happen sometimes when I resize the window but sometimes it doesn't. It seems to be inconsistent. is that part of this PR? (sorry, I have no knowledge of html and css, that's why I am asking).

RaffaeleMorganti:
For the scrollbar is simply css (actually is behind the page main content so it's hidden), I simply moved it 10px on left to let it be visible.

In the index page actually there are
* logo and badges aligned to keep a fixed distance to the right of the page
*  buttons menu aligned to keep a fixed distance to the left of the page

This cause them to overlay with certain window sizes

With the modification:
* logo and badges are positioned inline with text, no more with absolute position, so they cannot overlay anything (if window is too small they are shown into multiple lines)
* buttons menu is aligned to stay centered in the page

This make overlays impossible on any window size

Hope this clarify

pradkrish:
yeah, it does. I will let others with web design knowledge decide.

oleksiyskononenko:
@RaffaeleMorganti have you seen this PR https://github.com/h2oai/datatable/pull/2696 ? The new theme seems to already have a scroll bar at the left:

Screen Shot 2020-10-26 at 11 19 28 AM

RaffaeleMorganti:
@oleksiyskononenko oh I didn't know that. The version I pulled is simply the one I use as local reference. 
Nice to see the new theme, it looks pretty good, so I close this PR.

Since I looked at the new theme, please notice that in my opinion the lack of screen responsivity make it actually not really comfortable to use when not in full screen mode.

st-pasha:
Thanks for your feedback, Raffaele. I've tried to make some style adjustments to make it look better at small screen widths. Still, there may be some room for improvement there, so suggestions/PRs welcome!

PR #2708, opened 2020-10-25
fread parse error

pradkrish:
Closes #2680 

At `src/core/read/fread/fread_thread_context.cc:210,` If `j<ncols` and if `*tch` reaches the end of line with the current quote rule, the while loop exits ending up in IOError. 

With the code added in this PR, one way to fix this is to decrement tch`(tch--)` to the previous `sep` and try parsing again with the next quote rule. If the number of columns in the current row is only one, decrementing tch will never hit the sep in the current row and end up running into the previous row, hence the `if` logic inside the `while` loop to prevent this.

Issue #1036, opened 2018-05-11
File containing a single unescaped " out-of-sample is read incorrectly


st-pasha:
When `fill=False`, this example raises an error:
```
>>> src = "A,B,C\n" + "q,f,r\n" * 100 + "foo,\"bar,bza\n" + "a,bb,ccc\n" * 200
>>> dt.fread(src, fill=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/pasha/github/datatable/datatable/fread.py", line 69, in fread
    return freader.read()
  File "/Users/pasha/github/datatable/datatable/fread.py", line 715, in read
    _dt = core.gread(self)
RuntimeError: Too few fields on line 102: expected 3 but found only 2 (with sep=','). Set fill=True to ignore this error.  <<foo,"bar,bza>>
```
even though the file can be read with QR=3 (and in fact if we move problematic line to the beginning of the file, it will be parsed correctly).

When `fill=True` the file does not throw an error, but returns a 101x3 Frame where the last row contains value `"bar,bza\na,bb,ccc\n...` with the content of the rest of the file.

In addition, verbose mode shows
```
  Estimated number of rows: 2419 / 6.00 = 404
  Initial alloc = 444 rows (404 + 9%) using bytes/max(mean-2*sd,min) clamped between [1.1*estn, 2.0*estn]
  All rows were sampled since file is small so we know nrows=99 exactly
```
Obviously not all rows were sampled, and therefore `nrows=99` is a gross underestimation (there should be 301 rows).

pradkrish:
When I run this example on main branch with `fill=True,` it does return a 301*3 frame. 

By the way, how did you run in it verbose mode? I did `python -v filename.py` and did not see the info you mentioned above.

st-pasha:
"verbose mode" refers to supplying parameter `verbose=True` to fread:
```
dt.fread(src, fill=False, verbose=True)
```

st-pasha:
Indeed, this example seem to be working correctly now. So closing the issue.

st-pasha:
Ok, I just tried creating a test example for this, and it turns out it's not quite working.
See `test-fread-issues.py::test_issue1036`.

pradkrish:
@st-pasha I don't see `test-fread-issues.py::test_issue1036` on the main branch, are you sure you have added that example?

st-pasha:
You're right, I forgot to push that branch earlier. See #2730

[...]

PR #2709, opened 2020-10-26
fread should not detect sep within quoted fields

### LINKED ISSUE #922: fread should not detect sep within quoted fields

pradkrish:
Closes #922 

The idea is that if the input can be consistently parsed using QR=0/1, there is no need to use QR>1.


st-pasha (commented):
This is good, but we definitely need some tests to  verify  that it's actually working  as  intended. For example, #922 already has an example that can  be converted into a test.

pradkrish:
> This is good, but we definitely need some tests to verify that it's actually working as intended. For example, #922 already has an example that can be converted into a test.

Done. Should it be mentioned in the release notes?

st-pasha:
Yeah, the more release notes the merrier. There  is a dedicated "fread" section in the release notes for that.

src/core/csv/reader_fread.cc
```diff
@@ -345,6 +347,7 @@ void FreadReader::detect_sep_and_qr() {
             << topNumFields << " fields using quote rule " << topQuoteRule;
       }
     }
+    if (!check_all_quoteRules && quoteRule == 1) break; else continue;
```
st-pasha, line 350: i think `else  continue` is superfluous  here

pradkrish: my bad, of course, it is redundant, thanks for that :)

pradkrish:
This PR has been out for a couple weeks now. Do you think this is an easy one to close?

tests/fread/test-fread-small.py
```diff
@@ -1207,3 +1207,11 @@ def test_qr_bump_out_of_sample(capsys):
     pysrc[0][137] = -1
     pysrc[1][137] = '"This" is not funny'
     assert d0.to_list() == pysrc
+
+def test_sep_within_quoted_fields():
+    src1 = ['"abc"\n"1,000"\n"2,000"\n']
+    src2 = ['movies\n"encore, once more encore!"\n"great gatsby"']
+    d1 = dt.fread(src1)
+    d2 = dt.fread(src2)
+    assert d1.to_list() == [[1000, 2000]]
```
st-pasha, line 1216: it's better to use `assert_equals` (which checks also column names and types):
```
assert_equals(d1, dt.Frame(abc=[1000, 2000]))
```

tests/fread/test-fread-small.py
```diff
@@ -1207,3 +1207,11 @@ def test_qr_bump_out_of_sample(capsys):
     pysrc[0][137] = -1
     pysrc[1][137] = '"This" is not funny'
     assert d0.to_list() == pysrc
+
+def test_sep_within_quoted_fields():
+    src1 = ['"abc"\n"1,000"\n"2,000"\n']
+    src2 = ['movies\n"encore, once more encore!"\n"great gatsby"']
```
st-pasha, line 1213: `src1` and `src2` should be separated into 2 distinct tests, because they are unrelated to each other (and therefore can fail or succeed independently from each other).

Also, they don't have to be lists - plain strings will do.

src/core/csv/reader_fread.cc
```diff
@@ -296,6 +297,7 @@ void FreadReader::detect_sep_and_qr() {
         if (thisncol < 0) {
           // invalid file with this sep and quote rule; abort
           numFields[0] = -1;
+          if (quoteRule <= 1) check_all_quoteRules = true;
```
st-pasha, line 300: If parse fails at QR=0 but succeeds at QR=1, then we'll still go into checking all other QRs.
Consider the following test case:
```python
def test_sep_within_quoted_fields3():
    assert_equals(dt.fread('\n"1,\\"2"\n"3,\\"4"\n'),
                  dt.Frame(['1,"2', '3,"4']))
```

src/core/csv/reader_fread.cc
```diff
@@ -279,6 +279,7 @@ void FreadReader::detect_sep_and_qr() {
   // `numLines` has the number of lines in each group.
   int numFields[JUMPLINES+1];
   int numLines[JUMPLINES+1];
+  bool check_all_quoteRules = false;
```
st-pasha, line 282: Maybe it's better to have variable `max_quoteRule`, which starts at 2 but then gets changed to 4 (or vice versa)?
Then  the for loop below could be  `for (quoteRule=0; quoteRule<max_quoteRule; quoteRule++)`.

st-pasha (commented):
It's good, but there is still one case where the logic fails, see my comments below

src/core/csv/reader_fread.cc
pradkrish: That's definitely more descriptive, done.

src/core/csv/reader_fread.cc
pradkrish: Changed the logic a little and added this test as well.

tests/fread/test-fread-small.py
pradkrish: Done

tests/fread/test-fread-small.py
pradkrish: Done

src/core/csv/reader_fread.cc
pradkrish: After submitting the new patch, I realised that the new test I ended up adding, which by the way passes,
```python
def test_sep_within_quoted_fields3():
    assert_equals(dt.fread('cols\n"1,\\"2"\n"3,\\"4"\n'), dt.Frame(cols=['1,"2', '3,"4']))
```
is not the same as what you suggested. From your example, dt.fread('\n"1,\\"2"\n"3,\\"4"\n') gives
```
   | 1,"2
-- + ----
 0 | 3,"4

```
and dt.Frame(['1,"2', '3,"4']) gives
```
   | C0  
-- + ----
 0 | 1,"2
 1 | 3,"4
```
The columns names and the shape of the table don't add up and hence this test would fail on this branch. In fact, this fails on the master branch as well. do you think it's a bug?

st-pasha: How about  `dt.fread('"1,\\"2"\n"3,\\"4"\n', header=False)`?
I just don't  want fread to be able to infer that there is only one column solely based on the fact that the first row has only 1 token.

src/core/csv/reader_fread.cc
```diff
@@ -296,6 +297,7 @@ void FreadReader::detect_sep_and_qr() {
         if (thisncol < 0) {
           // invalid file with this sep and quote rule; abort
           numFields[0] = -1;
+          if (quoteRule <= 1) max_quoteRule = 2+(quoteRule*2);
```
st-pasha, line 300: so if parse fails (`thisncol<0`) at `quoteRule=0` (but not at `quoteRule=1`), then the new  `max_quoteRule` will be 2, which is the same as the initial  value of `max_quoteRule`. I'm not sure i fully understand the logic  here.

src/core/csv/reader_fread.cc
st-pasha: We could similarly have a test case with the qr=0 -encoded  quotes:
```
dt.fread('"1,""2"\n"3,""4"\n', header=False)
```

src/core/csv/reader_fread.cc
pradkrish: My apologies for leaving this issue open for a long time. This logic is obviously sloppy and I have pushed another commit, kindly have a look. 

Also, the tests seems to fail on py36 and 37 but it passes on my Linux machine when I tested it on py36, py37 and py38. I have no permission to check which/why tests are failing, can you have a look? Thanks.

pradkrish: @st-pasha can you kindly look into why this is failing?

st-pasha: Ah yes, it fails for the same reason that #2708 was failing. So if you merge `main` into this branch, the errors should (hopefully) go away.

st-pasha:
Also, can you please remove "xfail" mark from `tests/fread/test-fread-issues.py::test_issue1036` (and mark #1036 as closed) -- it seems to have been fixed by #2708

src/core/csv/reader_fread.cc
pradkrish: Done

pradkrish:
> Also, can you please remove "xfail" mark from `tests/fread/test-fread-issues.py::test_issue1036` (and mark #1036 as closed) -- it seems to have been fixed by #2708

Done

st-pasha (approved):

PR #2713, opened 2020-10-28
Cast functionality is now performed via the FExpr framework

### LINKED ISSUE #2712: TypeError: descriptor '__sub__' requires a 'datatable.FExpr' object but received a 'Expr'
```
from datatable import dt, f,  FExpr
fn = FExpr.__sub__
col1 = dt.float64(f.A)
col2 = dt.float64(f.B)
DT = dt.Frame(A=[3], B=[1])
DT[:, fn(col1, col2)]
```
produces  an exception
```
TypeError: descriptor '__sub__' requires a 'datatable.FExpr' object but received a 'Expr'
```

st-pasha:
- Added new function  `as_type()` for type-casting. This function is used from the `stype()` call, but is also available directly to the end-user.
- Using FExpr infrastructure  for  casting fixes the problem in issue #2712.

Closes #2712 
WIP for #2562


samukweku:
@st-pasha A suggestion here, to offer a dictionary option here, where different dtypes can be assigned to different columns. At the moment, only one dtype can be assigned to multiple columns.

PR #2722, opened 2020-11-01
[DOC] Examples for update() method

samukweku:
- Add examples for the `update` method

WIP #2604


st-pasha:
Same here: the examples are good, but if you moved them into the source file where the rest of the help string is declared, that would be even better.

samukweku:
@st-pasha No worries I'll work on it

st-pasha (approved):
But please convert tabs to spaces here too.

PR #2728, opened 2020-11-04
Join examples

samukweku:
WIP #2604 

- Add examples for joins


docs/api/dt/join.rst
```diff
@@ -3,3 +3,47 @@
...
+
+    df1[:, :, join(df2)]
+
+        date	        X1	X2	X3
+    0	01-01-2020	H	10	5
+    1	01-02-2020	H	30	5
+    2	01-03-2020	Y	15	10
+    3	01-04-2020	Y	20	10
+
+You can access the values in the right frame in an expression, using the secondary namespace object `g  <https://datatable.readthedocs.io/en/latest/api/dt/g.html#>`__::
```
oleksiyskononenko, line 37: I guess there is no need for the final `#` in the URL.

src/core/expr/py_join.cc
```diff
@@ -62,7 +62,51 @@ except: ValueError
...
+
+    df1 = dt.Frame("""date	       X1	X2
+            	01-01-2020	H	10
+            	01-02-2020	H	30
+            	01-03-2020	Y	15
+            	01-04-2020	Y	20""")
+
+    df2 = dt.Frame("""X1	X3
+            	   H	5
+                       Y	10""")
```
st-pasha, line 81: Could you please reformat with "convert Tabs to Spaces", i think that's what's causing everything to look misaligned.

src/core/expr/py_join.cc
```diff
@@ -62,7 +62,51 @@ except: ValueError
...
+
+    df1[:, :, join(df2)]
+
+        date	        X1	X2	X3
+    0	01-01-2020	H	10	5
+    1	01-02-2020	H	30	5
+    2	01-03-2020	Y	15	10
+    3	01-04-2020	Y	20	10
+
+You can access the values in the right frame in an expression, using the secondary namespace object `g  <https://datatable.readthedocs.io/en/latest/api/dt/g.html>`__::
```
st-pasha, line 98:
```suggestion
You can refer to columns of the joined frame using prefix :data:`g. <dt.g>`, similar to how columns of the left frame can be accessed using prefix :data:`f. <dt.f>`.
```

st-pasha (approved):

src/core/expr/py_join.cc
samukweku: If you don't mind @st-pasha explaining what you mean ? The misalignment is bugging me as well

st-pasha: In most IDEs there is a function "convert all tabs to spaces", and also a setting to "always insert tabs as spaces". I don't know which editor you're using, but I think googling this function for your particular editor will tell you where to find it.

src/core/expr/py_join.cc
```diff
@@ -62,7 +62,59 @@ except: ValueError
 
 See Also
 --------
-- :ref:`Tutorial on joins <join tutorial>`
+- `Tutorial on joins <../../start/quick-start.html#join>`_
+
+Examples
+--------
```
st-pasha, line 68: extra header?

src/core/expr/py_join.cc
st-pasha: Please note that internal links within the documentation should use dedicated syntax; in this case the link  to `g.` is 
```
:data:`g. <dt.g>`
```
This has several  advantages:
- the link is not rendered  as external (with a tiny arrow at the end);
- the link works properly when rendered locally;
- the link will work correctly with different versions of the documentation (i.e. v0.11, v1.0, latest, etc).

st-pasha (approved):

PR #2731, opened 2020-11-05
[DOC] first() examples

samukweku:
WIP #2604 

- Examples for the `first` function


oleksiyskononenko:
Looks good to me. @st-pasha I wonder, what is the best way to go with the examples:
- adding them directly to the `.rst` file;
- adding them to the corresponding `doc_` variable, `doc_first` in this case.

st-pasha:
I would argue in favor of including the examples in the `doc_first` variable declared in C++.  This would allow the same help to be available from Python too.

@samukweku If you look at the `:doc:` (or `:src:`) parameter of the `xfunction` directive, it will tell you the name of the C++ file where that variable is located.

oleksiyskononenko:
I would also say it is better to keep it in the `doc_` vars, one more point is that then we will keep everything in the single space. Easier to find, to fix, less surprises.

st-pasha (approved):

Issue #2729, opened 2020-11-04
Allow a frame to carry meta information


oleksiyskononenko:
Sometimes it is useful to attach meta information to a frame, that would describe the nature of data or provide additional insights. This could be implemented as a python dictionary under a new `.meta` frame property. Any datatable function that returns a frame could then fill this property with the relevant additional information, if needed. It should also be accessible in the r/w mode to the user.

PR #2734, opened 2020-11-09
Allow frame to carry meta information

oleksiyskononenko:
In this PR we add a new frame `.meta` property that could be used by datatable functions/methods or by the user to store any frame relevant meta information.

WIP for https://github.com/h2oai/datatable/issues/2358
Closes https://github.com/h2oai/datatable/issues/2729


st-pasha:
Should  `.meta` be  copied when doing `Frame.copy()`? Should it be preserved when saving  to Jay?

oleksiyskononenko:
That's a good question, right now this is what I put in the docs

```
it is not guaranteed that the existing meta information will be preserved
by the functions and methods called on the frame
```

I guess we can gradually start handling it in the functions/methods where we believe it is necessary. We may want to be consistent with how we handle the `.source` for example. Do we keep it now when making a copy / saving to jay?

st-pasha:
`.source` is preserved in `copy()`, but not in Jay, because the jay file becomes the new source.

st-pasha:
Maybe we should clarify  this moment in the docs: "The meta information is not preserved when saving the data into a Jay file. This behavior may change in the future".

Because I think there could be an expectation that this information will be stored (for example Arrow format stores any meta information).

oleksiyskononenko:
Yes, it is definitely a good idea to pickle/save to jay all the meta data related to a particular frame, but we need to think it through properly. For the moment I've improved the docs incorporating the changes you propose.

st-pasha (approved):

oleksiyskononenko:
Travis seems to be slow, but all the tests are passed, so merging.

Issue #2725, opened 2020-11-03
Cannot fread files 3+ Gb on Windows 


MichaelMoroz1993:
Hello!

I've faced the following issue:
The fread() function used to read 3+ Gb files returns the error message. At the same moment pandas reads those files without any issues.

[Link ](https://drive.google.com/drive/folders/1ijk-cqAleuXfzIkmNGyDNvPFmZ8b3u32?usp=sharing)to file.

The code and error message are below:

**Code:**
`import datatable as dt` 
`data = dt.fread(file='LF.csv')`

**Error:**
`Traceback (most recent call last):`
 `D:/.../datatable test.py:8 in <module>  data = dt.fread(file='LF.csv')`
`IOError: Unable to obtain size of LF.csv: [errno 132] value too large`

**Environment**
Datatable version - 0.11.0 
Python version - 3.7.2544.0
OS - Windows 10 Pro

Thank you!

samukweku:
I was able to read the data : 

```python
from datatable import fread
df = fread("LF.zip")
df.head()

C0	Row	Date	Timestamp	Sum	ID	Issue	Level	Support	Extras	Left	Skipped	Attempt	Active_support	Platform
▪▪▪▪	▪▪▪▪	▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪	▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪▪▪▪▪	▪▪▪▪	▪▪▪▪
0	0	0	2020-08-01	1596151327000	0	1100357	Fail	27	0	NA	2	0	4		iOS
1	1	1	2020-08-01	1596234522000	0	1101065	Fail	28	0	NA	1	0	2		iOS
2	2	2	2020-08-01	1595571719000	0	1075926	Fail	142	0	NA	1	0	1		iOS
3	3	3	2020-08-01	1582080163000	0	545616	Fail	457	0	NA	1	0	1		iOS
4	4	4	2020-08-01	1594868430000	0	1052669	Fail	262	1	NA	1	1	1		iOS
5	5	5	2020-08-01	1595446708000	0	1070398	Fail	338	0	NA	1	0	3		iOS
[...]

PR #2753, opened 2020-11-20
[WIP] Allow reading/writing >4Gb files on Windows

oleksiyskononenko:


docs/releases/v1.0.0.rst
```diff
@@ -42,6 +45,8 @@
     -[fix] Fread no longer throws an exception when the list of column types
       passed to parameter ``columns=`` contains ``str64``. [#2704]
 
+    -[fix] Fix an error when reading files larger than 4Gb on Windows. [#]
```
st-pasha, line 48: #2725

src/core/lib/mman/mman.cc
```diff
@@ -113,12 +113,9 @@
...
 
-      const off_t maxSize = off + (off_t)len;
-
-      const DWORD dwMaxSizeLow = (sizeof(off_t) <= sizeof(DWORD)) ?
-                      (DWORD)maxSize : (DWORD)(maxSize & 0xFFFFFFFFL);
-      const DWORD dwMaxSizeHigh = (sizeof(off_t) <= sizeof(DWORD)) ?
-                      (DWORD)0 : (DWORD)((maxSize >> 32) & 0xFFFFFFFFL);
+      const size_t maxSize = (size_t)off + len;
+      const DWORD dwMaxSizeLow = (DWORD)(maxSize & 0xFFFFFFFFL);
+      const DWORD dwMaxSizeHigh = (DWORD)((maxSize >> 32) & 0xFFFFFFFFL);
```
st-pasha, line 118: We should probably have a `static_assert` about the size of `DWORD` somewhere

docs/releases/v1.0.0.rst
```diff
@@ -13,6 +13,9 @@
       can now be an unsorted list. The list can also contain duplicate
       values.
 
+    -[fix] Fix a crash when saving frames larger than 4Gb to jay or csv
+      on Windows. [#]
```
st-pasha, line 17: If there is no corresponding issue, then `[#]` is superfluous.

st-pasha (commented):
could you move this PR to be based on top of `rel-0.11`? The only change that is needed is to put the changelog notes into `v0.11.1.rst` instead.

docs/releases/v1.0.0.rst
oleksiyskononenko: Do you use issue or PR numbers here? I was going to add a current PR number here.

src/core/lib/mman/mman.cc
oleksiyskononenko: I can add it, but as far as I can see it is 32 bits on all the Windows systems.

docs/releases/v1.0.0.rst
st-pasha: I usually use the issue number, just because PR# doesn't exist at the moment this is written. But I think it could work with the PR# too, if added later. GitHub will handle the redirect just fine.

oleksiyskononenko: Yeah, that's why I left it empty and then wanted to update with the PR number... but I already created the issue, so will update it with the issue numbers.

src/core/lib/mman/mman.cc
st-pasha: A static assert will make this knowledge explicit. Plus, if the new version of windows comes out with 64 bit dwords, then  we'll at least be able to catch it.

oleksiyskononenko: Sure, will add it. But my bet is that when a new version of Windows comes out, the code won't even compile :)

PR #2776, opened 2020-11-26
[DOC] sum() examples

samukweku:
WIP #2604 

- Examples for the `sum`  function

### DIFF
--- a/src/core/expr/head_reduce_unary.cc
+++ b/src/core/expr/head_reduce_unary.cc
@@ -352,6 +352,87 @@ See Also
 
 - :func:`count()` -- function to calculate a number of non-missing values.
 
+Examples
+--------
+
+::
+
+    from datatable import dt, f, by
+
+    df = dt.Frame({'A': [1, 1, 2, 1, 2],
+                   'B': [None, 2, 3,4, 5],
+                   'C': [1, 2, 1, 1, 2]})
+
+    df
+
+.. dtframe::
+    :names: A,B,C
+    :types: int8, int8, int8
+    :shape: 5, 2
+
+    0,1,NA,1
+    1,1,2,2
+    2,2,3,1
+    3,1,4,1
+    4,2,5,2
+
+Get the sum of column A::
+
+    df[:, dt.sum(f.A)]
+
+.. dtframe::
+    :names: A
+    :types: int8
+    :shape: 1, 1
+
+    0,7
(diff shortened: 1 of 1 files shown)


samukweku:
A couple of observations with the `sum` func:

- If you pass a list to the function, it acts like `rowsum`, with the caveat that any row containing null willl return NA:
```python

df = dt.Frame({'A': [1, 1, 2, 1, 2],
                   'B': [None, 2, 3,4, 5],
                   'C': [1, 2, 1, 1, 2]})

df[:, dt.sum([f.A, f.B])]

  | C0
0 | NA
1 | 3
2 | 5
3 | 5
4 | 7
```

Also, if you pass the values individually to `sum`, only the first column is recognized:

```python
df[:, dt.sum(f.A, f.B)]
    A
0 | 7

df[:, dt.sum(f.A, 25)]
    A
0 | 7
```

`sum` also does not offer those nifty features like `mean` and `median`, where you can pass a dictionary and the values will be computed, or a sequence and each column will be treated individually

src/core/expr/head_reduce_unary.cc
```diff
@@ -352,6 +352,87 @@ See Also
...
+
+    df = dt.Frame({'A': [1, 1, 2, 1, 2],
+                   'B': [None, 2, 3,4, 5],
+                   'C': [1, 2, 1, 1, 2]})
+
+    df
+
+.. dtframe::
+    :names: A,B,C
+    :types: int8, int8, int8
```
st-pasha, line 370: `int32`s here and below

src/core/expr/head_reduce_unary.cc
```diff
@@ -352,6 +352,87 @@ See Also
...
+    :names: C,A,B
+    :types: int8,int8,int8
+    :shape: 2, 3
+
+    0,1,4,7
+    1,2,3,7
+
+
+
+:func:`sum()` acts like python's built-in ``sum`` function if the value is not a :ref:`f-expressions`::
```
st-pasha, line 428: We will be changing this by the 1.0 release, so that all reducer functions have the same API.
I guess it's ok to leave this example as-is (given that this is how the function behaves right now), on the other hand we wouldn't want users to start relying on this behavior if it's going to be removed in the next release.

st-pasha (approved):

PR #2778, opened 2020-11-26
[DOC] examples for unique()

samukweku:
WIP #2604 

- Examples for the `unique()` function


src/core/set_funcs.cc
```diff
@@ -182,6 +182,58 @@ See Also
...
+
+    df = dt.Frame({'A': [1, 1, 2, 1, 2],
+                   'B': [None, 2, 3,4, 5],
+                   'C': [1, 2, 1, 1, 2]})
+
+    df
+
+.. dtframe::
+    :names: A,B,C
+    :types: int8, int8, int8
```
st-pasha, line 200: All 3 types will be `int32` here (and in the result frames below)

st-pasha (approved):

PR #2811, opened 2020-12-15
Show types for a frame in  python console

### LINKED ISSUE #2810: Console Frame should display column types
Currently when a  Frame is displayed, it shows  only column names and then data. However, it  would be even better if there were indicators of each  column's type.

st-pasha:
Each column's type is displayed under that column's name. In jupyter output the type of each column is already displayed.

For example:
```
       |  year  month    day  dep_delay  arr_delay  carrier  origin  dest   air_time  distance   hour
       | int32  int32  int32      int32      int32  str32    str32   str32     int32     int32  int32
------ + -----  -----  -----  ---------  ---------  -------  ------  -----  --------  --------  -----
     0 |  2014      1      1         14         13  AA       JFK     LAX         359      2475      9
     1 |  2014      1      1         -3         13  AA       JFK     LAX         363      2475     11
     2 |  2014      1      1          2          9  AA       JFK     LAX         351      2475     19
     3 |  2014      1      1         -8        -26  AA       LGA     PBI         157      1035      7
     4 |  2014      1      1          2          1  AA       JFK     LAX         350      2475     13
```
Closes #2810

### DIFF
--- a/docs/releases/v1.0.0.rst
+++ b/docs/releases/v1.0.0.rst
@@ -21,6 +21,10 @@
       can now be an unsorted list. The list can also contain duplicate
       values.
 
+    -[enh] When a Frame is shown in a python console, it will now display
+      the stype of each column, as a second line under the column names.
+      [#2810]
+
     -[fix] A Frame can now be created properly from a list of numpy bool
       objects. [#2762]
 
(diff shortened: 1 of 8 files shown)


samukweku:
@st-pasha; how do I pull this form into jupyter notebooks. It is clearer, faster, and more intuitive to see the words ('float64', 'int32') , than having to interprete the colours (which is the current form in jupyter notebooks).

st-pasha:
we'd need to update the code that handles HtmlWidget to print the types as values instead of attributes. It's relatively  simple though. We may even update the overall look of the table widget so that it matches what we have in the documentation.

samukweku:
Oh that would be great @st-pasha

PR #2820, opened 2020-12-19
Improve cut() to support binning based on the interval edges

### LINKED ISSUE #2819: cut() should support binning based on the provided binning edges
Right now `cut()` can only bin data based on the total number of bins. It should also be possible to provide interval edges manually, giving the user more control over the binning.

oleksiyskononenko:
Closes #2819 
WIP for #2358


src/core/expr/fexpr_cut.cc
```diff
@@ -54,6 +58,8 @@ class FExpr_Cut : public FExpr_Func {
       if (!py_nbins_.is_none()) {
         out += ", nbins=";
         out += py_nbins_.repr().to_string();
+        out += ", bins=";
+        out += py_bins_.repr().to_string();
```
st-pasha, line 62: we  probably don't want to repr parameters that are at their default values (such as `None`). Especially since having `bins` and `nbins` at the same time is not allowed

src/core/expr/fexpr_cut.cc
```diff
@@ -67,12 +73,31 @@ class FExpr_Cut : public FExpr_Func {
...
 
-      int32_t nbins_default = 10;
       Workframe wf = arg_->evaluate_n(ctx);
+
+      bool defined_bins = !py_bins_.is_none();
+      bool defined_nbins = !py_nbins_.is_none();
+
+      if (defined_bins && defined_nbins) {
+        throw ValueError() << "`bins` and `nbins` cannot be both set at the same time";
+      }
```
st-pasha, line 83: Validation of parameters should generally be done before the evaluation stage. This check, for example, can be performed as early as in `pyfn_cut`

src/core/expr/fexpr_cut.cc
```diff
@@ -110,17 +134,120 @@ class FExpr_Cut : public FExpr_Func {
...
+        // Retrieve bins
+        Column bins = dt->get_column(0);
+        if (!ltype_is_numeric(bins.ltype())) {
+          throw TypeError() << "Bin edges must be provided as the numeric columns only, "
+            << "instead for the frame `" << i << "` the column stype is `"
+            << bins.stype() << "`";
+        }
+        bins.cast_inplace(SType::FLOAT64);
+        validate_bins(bins, i);
+        bins.materialize();
```
st-pasha, line 199: Validation already computes all elements of the `bins` column, why not combine it with materialization?

src/core/expr/fexpr_cut.cc
```diff
@@ -110,17 +134,120 @@ class FExpr_Cut : public FExpr_Func {
...
+
+        // Retrieve bins
+        Column bins = dt->get_column(0);
+        if (!ltype_is_numeric(bins.ltype())) {
+          throw TypeError() << "Bin edges must be provided as the numeric columns only, "
+            << "instead for the frame `" << i << "` the column stype is `"
+            << bins.stype() << "`";
+        }
+        bins.cast_inplace(SType::FLOAT64);
+        validate_bins(bins, i);
```
st-pasha, line 198: Validation  should happen as soon as possible: in this case the validity of bins could have been checked as soon as this FExpr is created.
If the fexpr is reused you're currently re-doing the validation every time.

src/core/expr/fexpr_cut.cc
```diff
@@ -131,7 +258,7 @@ class FExpr_Cut : public FExpr_Func {
 //------------------------------------------------------------------------------
 
 static const char* doc_cut =
-R"(cut(cols, nbins=10, right_closed=True)
+R"(cut(cols, nbins=10, bins=None, right_closed=True)
 --
 .. x-version-added:: 0.11
```
st-pasha, line 264: The help text below should mention that intervals are not necessarily equal width if `bins=` is used

src/core/ltype.cc
```diff
@@ -99,6 +99,11 @@ bool is_ltype_object(PyObject* v) {
 }
 
 
+bool ltype_is_numeric(LType ltype) {
+  return ltype == LType::BOOL || ltype == LType::INT || ltype == LType::REAL;
```
st-pasha, line 103: VOID can be considered numeric too

src/core/expr/fexpr_cut.cc
oleksiyskononenko: It seems that the logic here is wrong even for `right_closed`, I will fix it.

src/core/expr/fexpr_cut.cc
oleksiyskononenko: What do you mean by combine? Do materialization manually along with the validation? It seems to me that materialization may not be necessary if the validation fails.

st-pasha: Validation here involves checking that all values are valid and in increasing order. It involves computing every element, and then comparing it with the previous element. So you could just save those values into a `std::vector<double>` at the same time.

src/core/expr/fexpr_cut.cc
oleksiyskononenko: Yes, not sure why but I have placed all the validation to `evaluate_n()` from the very beginning. It is not just about `validate_bins()`. Will probably move it to `pyfn_cut()` and also `pyfn_qcut()`...

src/core/expr/fexpr_cut.cc
oleksiyskononenko: Ah, yes, it makes sense. Thought you want to combine it with the standard `.materialize()` method.

src/core/expr/fexpr_cut.cc
oleksiyskononenko: I've been thinking about why I first put all the validation into the `evaluate_n()` and not to the corresponding `pyfn_` functions and found that probably I've reused the pattern we already had. `ifelse`, `round`, etc. they all perform validation at the `evaluate_n` stage. Any reason for that?

src/core/column/cut.h
```diff
@@ -197,6 +196,87 @@ class Cut_ColumnImpl : public Virtual_ColumnImpl {
...
+          *out = static_cast<int32_t>(bin_value(value, 0, nbins - 1));
+        }
+
+      }
+      return is_valid;
+    }
+
+
+    size_t bin_value(double value, size_t left, size_t right) const {
+      if (right == left + 1) {
```
st-pasha, line 264: so, if `nbins==2`, then at line 255 we  call `bin_value(value, 0, 1)`, and in this if-statement assign the value to bin 0 unconditionally -- nothing will go into the second bin?

st-pasha, line 264: maybe add an assert that right  is  always greater than left. Because otherwise there'll be an infinite recursion here.

src/core/column/cut.h
```diff
@@ -197,6 +196,87 @@ class Cut_ColumnImpl : public Virtual_ColumnImpl {
...
+      double value;
+      bool is_valid = col_.get_element(i, &value);
+
+      if (is_valid) {
+        size_t nbins = bins_.size();
+        is_valid = false;
+
+        if (gt(value, bins_[0]) && lt(value, bins_[nbins - 1])) {
+          is_valid = true;
+          *out = static_cast<int32_t>(bin_value(value, 0, nbins - 1));
```
st-pasha, line 255: what if nbins==1? or 0?

src/core/column/cut.h
oleksiyskononenko: I guess I just used a bad variable name here. This `nbins` is in reality something that should be called `nedges`, i.e. when we have `nbins==2` we indeed only have one bin, and that's why nothing will go to the second bin.

May be also need to rename `bins` to `bins_edges` or something.

oleksiyskononenko: Will do, but this won't help if we run it in a wrong way in the release mode...

src/core/column/cut.h
oleksiyskononenko: It cannot be, because this is indeed the number of the bin edges that we validate to be `>=2`. I probably need to invent a better name for this variable.

src/core/column/cut.h
```diff
@@ -197,6 +196,89 @@ class Cut_ColumnImpl : public Virtual_ColumnImpl {
...
+    dblvec bin_edges_;
+
+  public:
+
+    CutBins_ColumnImpl(Column&& col, dblvec&& bin_edges)
+      : Virtual_ColumnImpl(col.nrows(), dt::SType::INT32),
+        col_(std::move(col)),
+        bin_edges_(std::move(bin_edges))
+    {
+      xassert(ltype_is_numeric(col_.ltype()));
```
st-pasha, line 212: Add an assert that the size of `bin_edges_` is at least 2? To prevent this class from being used incorrectly

src/core/column/cut.h
st-pasha: Ah, that makes sense now. Yeah, a better name would be helpful in such a case.

st-pasha (approved):

src/core/column/cut.h
oleksiyskononenko: Yeah, let's do it.

PR #2826, opened 2021-01-11
[ENH] f-expression with list/tuple accepted in `j` selector

### LINKED ISSUE #2797: f[column_list] is not working
```
>>> column_list = ['a', 'b']
>>> f[column_list]
TypeError: Column selector should be an integer, string, or slice, not <class 'list'>
```

samukweku:
- resolves #2797 
- Allows lists/tuples of f-expression in the `j` selector

### DIFF
--- a/docs/api/namespace/__getitem__.rst
+++ b/docs/api/namespace/__getitem__.rst
@@ -2,14 +2,10 @@
 .. xmethod:: datatable.Namespace.__getitem__
     :src: src/core/expr/namespace.cc Namespace::m__getitem__
 
-    __getitem__(self, item)
+    __getitem__(self, *items)
     --
 
     Retrieve column(s) by their indices/names/types.
-
-    Also retrieve column(s) via :ref:`f-expressions`
-    containing a list/tuple of
-    column names/column positions/column types.
 
     By "retrieve" we actually mean that an expression is created
     such that when that expression is used within the
@@ -19,12 +15,12 @@
 
     Parameters
     ----------
-    item: int | str | slice | None | type | stype | ltype | Expr
+    items: int | str | slice | None | type | stype | ltype | list | tuple
         The column selector:
 
         ``int``
             Retrieve the column at the specified index. For example,
-            ``f[0]`` denotes the first column, whlie ``f[-1]`` is the
+            ``f[0]`` denotes the first column, while ``f[-1]`` is the
             last.
 
         ``str``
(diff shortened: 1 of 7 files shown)


tests/munging/test-dt-cols.py
```diff
@@ -107,7 +107,19 @@ def test_j_function(dt0):
...
+    frame = dt.Frame(
+    {
+        "id": ["a", "a", "a", "b", "b"],
+        "type": ["in_scope", "in_scope", "exclude", "in_scope", "exclude"],
+        "value": [5, 5, 99, 20, 99],
+    }
+)
+
+    # multiple column selection by label
+    assert frame[:, f['value', 'id']].to_list()[0] == frame[:, ['value', 'id']].to_list()[0]
```
st-pasha, line 120: There is a  function  `assert_equals()`  for comparing frames:
```
assert_equals(frame[:, f['value', 'id']],
              frame[:, ['value', 'id']])
```

st-pasha (commented):
We should probably update documentation at `api/namespace/__getitem__.rst` to mention this selector.

docs/api/namespace/__getitem__.rst
```diff
@@ -43,3 +43,11 @@
...
     --------
     - :ref:`f-expressions` -- user guide on using f-expressions.
+
+    Notes
+    -----
+    .. x-version-changed:: 1.0.0
+
+        :ref:`f-expressions` containing a list/tuple of
+        column names/column positions/column types are
+        accepted within the `j` selector.
```
st-pasha, line 53: Shouldn't this be in the list of types for `item`? Otherwise, a user might not notice this information down the page.

samukweku: fair point. Let me edit that

samukweku: @st-pasha Would have loved to get ``For example, ``f[0, -1]`` will return the first and last columns.``  on another line (semantic break is what I had in mind) but could not pull it off. Any suggestions?

docs/releases/v1.0.0.rst
```diff
@@ -52,6 +52,9 @@
     -[fix] Fix a crash that occurred when using `median()` on virtual
       columns of type ArrayView64. [#2802]
 
+    -[enh] :ref:`f-expressions` now accepts a list/tuple of
+      column names/column positions/column types in the ``j`` section.
```
st-pasha, line 56: Mention the issue number here?

docs/api/namespace/__getitem__.rst
```diff
@@ -7,6 +7,10 @@
 
     Retrieve column(s) by their indices/names/types.
 
+    Also retrieve column(s) via :ref:`f-expressions`
```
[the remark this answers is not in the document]
samukweku: Just as an aside, @st-pasha Thank you for the guidance and training so far. invaluable. Your patience and explanation style has been really helpful.

I will add examples here to make it clearer/explicit, especially for the tuple syntax, as that, in my opinion, would be the best way to clear any confusion, and make the documentation easier to write.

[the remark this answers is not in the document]
st-pasha: Go ahead, make the change. You probably know your way around the documentation better than I do by this point :)

st-pasha (approved):

oleksiyskononenko (approved):

PR #2828, opened 2021-01-15
Optimize i-selector of frame type

chi2liu:
Under these circumstances:
`df[condition, :]` that condition is a frame
If the condition frame contains a more complex virtual column, because the function `ArrayRowIndexImpl::init_from_boolean_column` does not have the parallelized `get_element` function, serialization is slower. **Call the parallelized `materialize` function in advance to optimize performance**.**On the other hand, `materialize` function in advance can also use data locality to improve performance**

Take the data of kaggle competition as an example，can be accelerated **3x** times in my machine.
https://www.kaggle.com/benhamner/sf-bay-area-bike-share/data

image

Machine hardware：
_Intel(R) Core(TM) i9-9820X CPU @ 3.30GHz
20 cores_

### DIFF
--- a/src/core/rowindex_array.cc
+++ b/src/core/rowindex_array.cc
@@ -140,13 +140,15 @@ void ArrayRowIndexImpl::init_from_boolean_column(const Column& col) {
     return;
   }
   int8_t value;
+  Column colcopy = col;
+  colcopy.materialize();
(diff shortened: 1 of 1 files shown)


st-pasha (approved):

chi2liu:
@st-pasha will it be mered ?

Issue #2829, opened 2021-01-19
Segfault when deleting column by reference then trying to rename column to the deleted column.


ChristopherEeles:
- Did you find a bug in datatable, or maybe the bug found you?
@EvgeniyaGorobets found this issue data munging a large datatable in an ETL pipeline leveraging datatable to manipulate tabular data (>million rows).

- How to reproduce the bug?
1. Read in a big datatable, this was ours:
```
>>> gene_drug_dt
         | gene_id          drug_id        estimate  se   n  tstat  …  mDataType  tested_in_human_trials  in_clinical_trials
-------- + ---------------  -----------  ----------  --  --  -----     ---------  ----------------------  ------------------
       0 | ENSG00000160993  Palbociclib   0.0317806  NA  34     NA  …  rna                            NA                  NA
       1 | ENSG00000160999  Palbociclib  -0.182501   NA  34     NA  …  rna                            NA                  NA
       2 | ENSG00000161010  Palbociclib   0.0868562  NA  34     NA  …  rna                            NA                  NA
       3 | ENSG00000161011  Palbociclib   0.0206584  NA  34     NA  …  rna                            NA                  NA
       4 | ENSG00000161013  Palbociclib  -0.147505   NA  34     NA  …  rna                            NA                  NA
       5 | ENSG00000161016  Palbociclib   0.0789258  NA  34     NA  …  rna                            NA                  NA
       6 | ENSG00000161021  Palbociclib  -0.185208   NA  34     NA  …  rna                            NA                  NA
       7 | ENSG00000161057  Palbociclib   0.0916426  NA  34     NA  …  rna                            NA                  NA
       8 | ENSG00000161091  Palbociclib   0.036271   NA  34     NA  …  rna                            NA                  NA
       9 | ENSG00000161217  Palbociclib   0.395108   NA  34     NA  …  rna                            NA                  NA
      10 | ENSG00000161267  Palbociclib   0.0580221  NA  34     NA  …  rna                            NA                  NA
      11 | ENSG00000161270  Palbociclib   0.0946618  NA  34     NA  …  rna                            NA                  NA
      12 | ENSG00000161281  Palbociclib   0.260692   NA  34     NA  …  rna                            NA                  NA
      13 | ENSG00000161381  Palbociclib  -0.0360979  NA  34     NA  …  rna                            NA                  NA
      14 | ENSG00000161395  Palbociclib  -0.183836   NA  34     NA  …  rna                            NA                  NA
[...]

PR #2830, opened 2021-01-19
Fix segfault when renaming untouched columns

oleksiyskononenko:
Closes https://github.com/h2oai/datatable/issues/2829


st-pasha (approved):

st-pasha (commented):
Maybe add another test where we try to rename after a column was deleted (like in the original example)?

oleksiyskononenko:
@st-pasha Yes, I can. But the bug has nothing to do with deleting and renaming to the old column name...

PR #2832, opened 2021-01-21
[DOC] Examples for symdiff()

samukweku:
WIP #2604 
 - Examples for symdiff()

### DIFF
--- a/src/core/set_funcs.cc
+++ b/src/core/set_funcs.cc
@@ -179,7 +179,7 @@ Examples
 --------
 .. code-block:: python
 
-    >>> from datatable import dt, f, by
+    >>> from datatable import dt, f
     >>>
     >>> df = dt.Frame({'A': [1, 1, 2, 1, 2],
     ...                'B': [None, 2, 3,4, 5],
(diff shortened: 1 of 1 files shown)


src/core/set_funcs.cc
```diff
@@ -679,6 +679,53 @@ except: ValueError | NotImplementedError
...
           - raised when one of the columns has stype `obj64`.
 
+Examples
+--------
+.. code-block:: python
+
+    >>> from datatable import dt, f
+    >>>
+    >>> df = dt.Frame({'A': [1, 1, 2, 1, 2],
+    ...                'B': [None, 2, 3,4, 5],
```
oleksiyskononenko, line 689: Minor: I would replace `3,4` with `3, 4`.

src/core/set_funcs.cc
```diff
@@ -679,6 +679,53 @@ except: ValueError | NotImplementedError
...
+       | int32  int32  int32
+    -- + -----  -----  -----
+     0 |     1     NA      1
+     1 |     1      2      2
+     2 |     2      3      1
+     3 |     1      4      1
+     4 |     2      5      2
+    [5 rows x 3 columns]
+
+Symmetric difference of all values in the entire frame::
```
oleksiyskononenko, line 702: May be we better say that here each column is treated as a separate frame, and we do symmetric difference between three frames.

oleksiyskononenko (approved):

st-pasha (approved):

PR #2834, opened 2021-01-21
[DOC] Examples for setdiff()

samukweku:
WIP #2604 

- Example for setdiff()

### DIFF
--- a/src/core/set_funcs.cc
+++ b/src/core/set_funcs.cc
@@ -179,7 +179,7 @@ Examples
 --------
 .. code-block:: python
 
-    >>> from datatable import dt, f, by
+    >>> from datatable import dt
     >>>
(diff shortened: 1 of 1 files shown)


samukweku:
@st-pasha @oleksiyskononenko the function definition for `setdiff` is different from the other set functions. Any reason for this difference? Application wise, I have not noticed any difference.

oleksiyskononenko:
@samukweku You mean that this function takes two parameters and not just one? I guess this is because we want the user to be confident that the `frame0 - frames` subtraction occurs, and not something else. The order of frames is important in this case.

samukweku:
Actually it makes sense now. Thanks.

st-pasha (approved):

PR #2840, opened 2021-01-21
Prevent overflow in FTRL predictions

oleksiyskononenko:
For multinomial classification problems FTRL does `softmax` normalization of the computed predictions. Since it involves computation of exponential functions, it may occasionally overflow on some data. In this PR we fix this issue by trying to avoid calculations of exponential function on large arguments.

WIP for https://github.com/h2oai/h2oai/issues/20286


st-pasha (approved):
Makes sense.

PR #2849, opened 2021-02-11
New Type class

st-pasha:
The class Type is intended as a replacement for stype/ltype enums.
The main difference from `stype` is that the new class is not an enum, and supports arbitrary type descriptors.

This new class is necessary in order to implement datetime types with various precision / timezone parameters. In the future this class will also be able to accommodate columns of "list" types, compound (struct) columns, union columns, etc.

In the future we want to gradually replace all uses of `stype` with the equivalent code that uses `Type`s.

### DIFF
--- a/src/core/python/xobject.h
+++ b/src/core/python/xobject.h
@@ -19,6 +19,7 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
+#include <iostream>
 #ifndef dt_PYTHON_XOBJECT_h
 #define dt_PYTHON_XOBJECT_h
 #include <cstring>    // std::memcpy, std::memset, std::strrchr
@@ -45,6 +46,10 @@ namespace py {
 class XTypeMaker {
   private:
     PyTypeObject* type;
+    size_t object_size;
+    const char* class_name_;
+    bool dynamic_type_;
+    size_t : 56;
     std::vector<PyGetSetDef> get_defs;
     std::vector<PyMethodDef> meth_defs;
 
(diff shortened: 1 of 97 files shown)


st-pasha:
Note: some of the tests are currently failing, but that is addressed by a separate PR #2851 and has nothing to do with this PR.

src/core/python/xobject.h
```diff
@@ -19,6 +19,7 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 // IN THE SOFTWARE.
 //------------------------------------------------------------------------------
+#include <iostream>
```
oleksiyskononenko, line 22: do we need this include?

src/core/types/py_type.h
```diff
@@ -0,0 +1,66 @@
...
+//------------------------------------------------------------------------------
+#ifndef dt_TYPES_PY_TYPE_h
+#define dt_TYPES_PY_TYPE_h
+#include "types/type.h"
+#include "python/xobject.h"
+namespace dt {
+
+
+/**
+  * This class corresponds is python-visible `dt.Type`: it describes
```
oleksiyskononenko, line 30: `This class corresponds is python-visible `dt.Type` — typo somewhere?

tests/test-types.py
```diff
@@ -1,12 +1,189 @@
...
+    with pytest.raises(ValueError, match=msg):
+        Type(0)
+    with pytest.raises(ValueError, match=msg):
+        Type(0.5)
+    with pytest.raises(ValueError, match=msg):
+        Type("nothing")
+    with pytest.raises(ValueError, match=msg):
+        Type(type)
+    with pytest.raises(TypeError):
+        Type()
```
oleksiyskononenko, line 186: Add a test accessing `.types` for a frame?

src/core/python/xobject.h
st-pasha: no, this was temporary

tests/test-types.py
st-pasha: Right, forgot about that...

oleksiyskononenko (approved):

PR #2851, opened 2021-02-11
Change required version of xlrd package

### LINKED ISSUE #2823: xlrd no longer supports xlsx, unhelpful error
Without pinning an old version of xlrd, fread on an xlsx file will report "AttributeError: module 'xlrd' has no attribute 'xlsx'"

- How to reproduce the bug?
```
import datatable as dt


path = "some xlsx path"
return dt.fread(path)

```

  An error specifying that xlrd must be version 1.2.0 to read xlsx files (like the error if xlrd is not installed).

- Environment:
  python 3.8.3, xlrd 2.0.0, datatable 0.11.1

Suggested workaround: install xlrd==1.2.0

st-pasha:
Changed the required version of `xlrd` package to "<=1.2.0", due to the fact that the latest version 2.0.0 removed support for xlsx files. Note that we are only changing the "extra" requirements: these are not forced on the user, rather they are what our package recommends to have.

It might be desirable at this point to either switch to `openpyxl` library for reading xlsx files, or to implement the necessary logic from scratch -- but both of these options require considerable time. 

This change, on the other hand, can be done quickly, and it solves the immediate problem that some of our tests that use xlsx files are now failing.

Closes #2823

### DIFF
--- a/requirements_extra.txt
+++ b/requirements_extra.txt
@@ -1,4 +1,4 @@
 numpy
 pandas
 pyarrow
-xlrd
+xlrd<=1.2.0


oleksiyskononenko (approved):

samukweku:
Building from scratch looks more desirable to get more speed. But yes, it will take up some time

Issue #2858, opened 2021-02-19
Implement date32 column


st-pasha:
- [x] Create `date32` type;
- [x] Display `date32` column in a python console;
- [x] convert `date32` column into a python list;
- [x] convert python list of `datetime.date` objects into a date32 column;
- [x] read/write `date32` column in Jay
- [x] write `date32` column to csv
- [x] read `date32` column from csv (fread)
- [x] convert `date32` column into numpy
- [x] create `date32` column from numpy
- [x] convert `date32` column into pandas
- create `date32` column from pandas (requires time64 type)
- [x] convert `date32` column into pyarrow
- [x] create `date32` column from pyarrow
- [x] implement type casts for `date32` column
- [x] create `date32` column from parts year, month, day
- [x] retrieve `year` from a `date32` column
- [x] retrieve `month` from a `date32` column
- [x] retrieve `day` from a `date32` column
- [x] function `day_of_week()`
- [x] compute stats for `date32` column, make sure `min()`, `max()` and `countna()` work
- [x] sort `date32` column
- [x] relational operators: `<`, `<=`, `==`, `>=`, `>` 

(This issue is part of #1646)

oleksiyskononenko:
fread to `date32`?

Peter-Pasta:
Will there be a function that works the same as time.ymd() but takes a string (e.g. '2021-01-01') as input?

st-pasha:
Yeah, need to add that too

PR #2886, opened 2021-03-10
Convert date32 column to numpy

st-pasha:
- date32 columns can now be converted into numpy arrays.
- void columns can now be converted into numpy arrays.
- Passing an invalid value to the `column=` argument of the `.to_numpy()` method will no longer result in a crash.
- The `stype=` argument of `.to_numpy()` renamed into `type=` and accepts all valid `Type`-like values. The previous name of the argument still works though, so this is not a breaking API change.
- Converting a frame with incompatible types into a numpy array will now raise an error (instead of auto-promoting to object type). However, if the user explicitly requests promotion into the object type then there won't be any error.
- When a frame is converted into a numpy array of floatinng type, then we will produce a regular `np.ndarray` instead of a masked array.
- If an exception occurs when converting a frame into numpy array, that exception will be properly reported to the user (previously, it would result in creating a numpy array of shape 1x1 and containing the Frame object).
- Partially implemented casting date32 type.
- Fixed old warnings in sort.cc.
- Expanded numpy tests.

Internal API changes:
- The py_buffers file was split into 2: one for sending data out via the Buffers protocol, and another for accepting external data via the Buffers protocol;
- Methods `ColumnImpl::cast_const()` and `ColumnImpl::cast_mutate()` combined into one `ColumnImpl::cast_replace()`.
- Created new type "Invalid", for internal use only. Certain methods (such as `Type::promote()`) can return this type to signal an error condition without actually raising an error.

WIP for #2858


oleksiyskononenko (approved):

src/core/sort.cc
```diff
@@ -739,6 +740,7 @@ class SortContext {
                      : static_cast<uint8_t>(128 - xi[j]) >> 6;
         });
     }
+    RESTORE_CLANG_WARNING("-Wpadded")
```
oleksiyskononenko, line 743: So no way to fix this besides disabling?

st-pasha: the warning is in the auto-generated lambda code, so not much. Other than rewriting the lambda itself, but that doesn't seem possible here

PR #2889, opened 2021-03-11
Write date32 to csv

st-pasha:
Also, some minor refactoring:
- added method `XArgs::add_synonym_arg` (same as in `PKArgs`);
- functions `itoa()` and `btoa()` are no longer forced to be inline.

WIP for #2858


src/core/csv/toa.cc
```diff
@@ -0,0 +1,104 @@
+//------------------------------------------------------------------------------
+// Copyright 2018-2021 H2O.ai
```
oleksiyskononenko, line 2: If this is a new file, shouldn't it be just "Copyright 2021 H2O.ai"?

oleksiyskononenko (approved):

src/core/csv/toa.cc
st-pasha: it's new, but it contains code that was previously in `toa.h`, and that was marked as 2018

oleksiyskononenko: Yeah, I saw it, thought we reflect the date the file was created/updated, not the code.

PR #2891, opened 2021-03-12
date32 column save to Jay + read from Jay

st-pasha:
Implement serializing date32 columns into Jay, and reading the file back.

Note: I had to edit `jay_generated.h` manually  inserting 
```
if (!obj) return true;
```
at line 611, because otherwise old Jay files were reported as being invalid.

The `flatbuffers` library used in our project updated as well to the most recent version (it hasn't been updated since 2018).

WIP for #2858


oleksiyskononenko:
Seems like `tests/types/test-date32.py` was updated from some other PR, should we should merge `main` first.

src/core/jay/jay_generated.h
```diff
@@ -23,8 +23,10 @@ struct StatsFloat32;
 struct StatsFloat64;
 
 struct Frame;
+struct FrameBuilder;
```
oleksiyskononenko, line 26: How this `jay_generated.h` file is generated?

tests/types/test-date32.py
```diff
@@ -127,6 +127,23 @@ def test_from_numpy_with_nats(np):
...
+def test_save_to_jay(tempfile_jay):
+    d = datetime.date
+    src = [d(1, 1, 1), d(2001, 12, 13), d(2026, 5, 9), None, d(1956, 11, 11)]
+    DT = dt.Frame(src)
+    DT.to_jay(tempfile_jay)
+    del DT
+    DT2 = dt.fread(tempfile_jay)
+    assert DT2.shape == (5, 1)
+    assert DT2.types == [dt.Type.date32]
+    assert DT2.to_list()[0] == src
```
oleksiyskononenko, line 143: Do we also want to calculate stats on the date column and save/read it to/from jay?

oleksiyskononenko (approved):

src/core/jay/jay_generated.h
st-pasha: There is a comment about this in `jay.fbs`: you run `flatc --cpp jay.fbs`, where `flatc` must be installed through either `brew` or `apt-get` or whatever mechanism is available on your system.

tests/types/test-date32.py
st-pasha: ok

PR #2893, opened 2021-03-13
date32 to/from pyarrow

st-pasha:
Our `date32` column can now be converted into pyarrow's `date32` type;
likewise pyarrow's `date32` and `date64` columns can be converted into our `date32`.

WIP for #2858

### DIFF
--- a/src/core/frame/to_arrow.cc
+++ b/src/core/frame/to_arrow.cc
@@ -166,6 +166,7 @@ std::unique_ptr<dt::OArrowSchema> Column::to_arrow_schema() const {
     case dt::SType::INT64:   (*osch)->format = "l"; break;
     case dt::SType::FLOAT32: (*osch)->format = "f"; break;
     case dt::SType::FLOAT64: (*osch)->format = "g"; break;
+    case dt::SType::DATE32:  (*osch)->format = "tdD"; break;
     case dt::SType::STR32:   (*osch)->format = "u"; break;
     case dt::SType::STR64:   (*osch)->format = "U"; break;
     default:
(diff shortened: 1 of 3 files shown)


src/core/frame/to_arrow.cc
```diff
@@ -345,6 +346,7 @@ Column dt::ColumnImpl::as_arrow() const {
     case SType::INT64: return _as_arrow_fw<int64_t>();
     case SType::FLOAT32: return _as_arrow_fw<float>();
     case SType::FLOAT64: return _as_arrow_fw<double>();
+    case SType::DATE32: return  _as_arrow_fw<int32_t>();
```
oleksiyskononenko, line 349: So we just convert timestamps, right?

src/core/frame/to_arrow.cc
```diff
@@ -166,6 +166,7 @@ std::unique_ptr<dt::OArrowSchema> Column::to_arrow_schema() const {
     case dt::SType::INT64:   (*osch)->format = "l"; break;
     case dt::SType::FLOAT32: (*osch)->format = "f"; break;
     case dt::SType::FLOAT64: (*osch)->format = "g"; break;
+    case dt::SType::DATE32:  (*osch)->format = "tdD"; break;
```
oleksiyskononenko, line 169: May be we add a comment as to what are all these "n", "b", "tdD", etc. identifiers?

oleksiyskononenko (approved):

src/core/frame/to_arrow.cc
st-pasha: Yes, our DATE32 format coincides with arrow's `date32`, so this conversion is straightforward.

src/core/frame/to_arrow.cc
st-pasha: ok

PR #2897, opened 2021-03-16
New function time.ymd()

st-pasha:
Function `dt.time.ymd()` can be  used to create `date32` columns out of year/month/day components.

I've decided that all date/time functions will leave in the `datatable.time` namespace, so that they can be easily discovered  and run less chance of clashing with some other functionality.

WIP for #2858

### DIFF
--- a/docs/api/index-api.rst
+++ b/docs/api/index-api.rst
@@ -28,6 +28,9 @@ Submodules
     * - :mod:`models. <datatable.models>`
       - A small set of data analysis tools.
 
+    * - :mod:`time. <datatable.time>`
+      - Functions for working with date/time columns.
+
 
 
 
@@ -196,6 +199,7 @@ Other
     math.             <math>
     models.           <models>
     options.          <options>
+    time.             <time>
     FExpr             <fexpr>
     Frame             <frame>
     ltype             <ltype>
--- a/docs/api/time.rst
+++ b/docs/api/time.rst
@@ -1,0 +1,18 @@
+
+.. xpy:module:: datatable.time
+
+datatable.time
+==============
+
+.. list-table::
+   :widths: auto
+   :class: api-table
+
+   * - :func:`ymd(y,m,d)`
+     - Create a ``date32`` column from year, month, day components.
+
+
+.. toctree::
+    :hidden:
+
+    ymd()   <time/ymd>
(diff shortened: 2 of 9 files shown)


docs/api/index-api.rst
```diff
@@ -28,6 +28,9 @@ Submodules
     * - :mod:`models. <datatable.models>`
       - A small set of data analysis tools.
 
+    * - :mod:`time. <datatable.time>`
```
oleksiyskononenko, line 31: Why not `datatable.datetime`?

docs/api/time.rst
```diff
@@ -0,0 +1,18 @@
...
+
+datatable.time
+==============
+
+.. list-table::
+   :widths: auto
+   :class: api-table
+
+   * - :func:`ymd(y,m,d)`
+     - Create a ``date32`` column from year, month, day components.
```
oleksiyskononenko, line 12: "year, month, day" -> "year, month and day"?

src/core/expr/time/fexpr_ymd.cc
```diff
@@ -0,0 +1,230 @@
...
+      return Column(new YMD_ColumnImpl(
+          std::move(ycol), std::move(mcol), std::move(dcol)));
+    }
+};
+
+
+
+
+//------------------------------------------------------------------------------
+// Python-facing `qcut()` function
```
oleksiyskononenko, line 169: "qcut()" -> "ymd()"

oleksiyskononenko (approved):

docs/api/index-api.rst
st-pasha: I use the word "time" not in the  sense of "time of day", but as the physical _time_. That thing that keeps everything from happening at once. The fourth dimension.

For me, from a physicist's standpoint, the word "time" is perfectly self-sufficient, and doesn't need to be contorted into a "datetime" or "timestamp" or "time_point" or whatever. Specifically, a "date" is a subset of time -- 24-hour long slices of time to be precise, and the module name `dt.time` is thus suitable for describing everything temporal.

Plus, it's shorter and easier to type.

src/core/expr/time/fexpr_ymd.cc
st-pasha: copy-paste... :)

docs/api/time.rst
st-pasha: ok

PR #2914, opened 2021-03-29
Do data shuffling when fitting a linear model

oleksiyskononenko:
In general, it is recommended to shuffle training data when fitting the model with SGD. The same is suggested in the [paper](http://martin.zinkevich.org/publications/nips2010.pdf) that we base our parallel implementation on. 

In this PR we implement quasi-random data shuffling for the training data, similar to what we currently do in Aggregator. The shuffling could be disabled by setting the seed to zero.

WIP for #2358

### DIFF
--- a/src/core/models/dt_linearmodel.cc
+++ b/src/core/models/dt_linearmodel.cc
@@ -147,9 +147,9 @@ LinearModelFitOutput LinearModel<T>::fit_impl() {
   NThreads nthreads = nthreads_from_niters(iteration_nrows, MIN_ROWS_PER_THREAD);
 
   // Calculate parameters for the modular quasi-random generator.
-  // By default, when seed is zero, we don't do data shuffling.
-  ModularParams mp;
-  if (seed_) mp = modular_random_gen(dt_X_fit_->nrows(), seed_);
+  // By default, when seed is zero, modular_random_gen() will return
+  // multiplier == 1 and increment == 0, so we don't do any data shuffling,
+  auto mp = modular_random_gen(dt_X_fit_->nrows(), seed_);
 
   dt::parallel_region(nthreads,
     [&]() {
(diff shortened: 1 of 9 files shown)


src/core/models/utils.cc
```diff
@@ -23,37 +23,58 @@
...
+
+/**
+ *  Generate a random multiplier and increment for quasi-random generator
+ *  based on the modular arithmetics. When seed is zero, we just return
+ *  the default parameters: multiplier == 1 and increment == 0.
+ */
+ModularParams modular_random_gen(size_t n, unsigned int seed) {
+  ModularParams mp;
+  if (seed) {
+    auto multipliers = calculate_coprimes(n);
```
st-pasha, line 35: Isn't it a bit excessive, to calculate all coprimes here? For example, if `n` is prime, then all numbers `1 .. n-1` are coprimes. 
A simpler strategy could be to just pick a number below n at random, check if it's coprime, and retry if not. After all, [the probability for two numbers to be coprime](https://math.stackexchange.com/questions/64498) is 6/pi^2 ≈ 60%.

src/core/models/dt_linearmodel.cc
```diff
@@ -168,7 +174,8 @@ LinearModelFitOutput LinearModel<T>::fit_impl() {
 
         // Training
         dt::nested_for_static(iteration_size, ChunkSize(MIN_ROWS_PER_THREAD), [&](size_t i) {
-          size_t ii = (iteration_start + i) % dt_X_fit_->nrows();
+          // Do quasi-random data shuffling
+          size_t ii = ((iteration_start + i) * mp.multiplier + mp.increment) % dt_X_fit_->nrows();
```
st-pasha, line 178: One problem with this approach is that it accesses data in semi-random pattern. In `kfold_random`, on the other hand, we randomly partition data into k parts such that each part's indices are sorted.

st-pasha (approved):

Issue #2920, opened 2021-04-01
Implement learning rate schedules


oleksiyskononenko:
Right now the learning rate `eta` can only be constant in the linear model. Since the value of this constant is not known in advance, it is useful to have schedules, where `eta` is automatically adjusted during the model fitting.

WIP for https://github.com/h2oai/datatable/issues/2358

PR #2933, opened 2021-04-05
Implement learning rate schedules for linear model

oleksiyskononenko:
Implement constant, time-based, step-based and exponential learning rate schedules for linear model.

Closes https://github.com/h2oai/datatable/issues/2920

### DIFF
--- a/src/core/models/dt_linearmodel.cc
+++ b/src/core/models/dt_linearmodel.cc
@@ -318,7 +318,7 @@ void LinearModel<T>::adjust_eta(T& eta, size_t iter) {
       eta = eta0_;
       break;
     case LearningRateSchedule::TIME_BASED:
-      eta /= (1 + eta_decay_ * iter);
+      eta = eta0_ / (1 + eta_decay_ * iter);
       break;
     case LearningRateSchedule::STEP_BASED:
       eta = eta0_ * pow(eta_decay_, floor((1 + iter) / eta_drop_rate_));
--- a/src/core/models/dt_linearmodel.cc
+++ b/src/core/models/dt_linearmodel.cc
@@ -59,6 +59,9 @@ LinearModelFitOutput LinearModel<T>::fit(
 {
   // Cast input parameters to `T`
   eta_ = static_cast<T>(params->eta);
+  eta_decay_ = static_cast<T>(params->eta_decay);
+  eta_drop_rate_ = static_cast<T>(params->eta_drop_rate);
+  eta_schedule_ = params->eta_schedule;
   lambda1_ = static_cast<T>(params->lambda1);
   lambda2_ = static_cast<T>(params->lambda2);
   nepochs_ = static_cast<T>(params->nepochs);
@@ -74,13 +77,14 @@ LinearModelFitOutput LinearModel<T>::fit(
 
   auto res = fit_model();
 
-  dt_X_fit_ = nullptr; dt_y_fit_ = nullptr; dt_X_val_ = nullptr;
+  dt_X_fit_ = nullptr;
+  dt_y_fit_ = nullptr;
+  dt_X_val_ = nullptr;
   dt_y_val_ = nullptr;
   nepochs_val_ = T_NAN;
   val_error_ = T_NAN;
   return res;
 }
-
 
 
 /**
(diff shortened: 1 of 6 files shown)


src/core/models/dt_linearmodel.cc
```diff
@@ -301,6 +308,27 @@ LinearModelFitOutput LinearModel<T>::fit_impl() {
...
+ *  Calculate learning rate for a particular schedule.
+ */
+template <typename T>
+void LinearModel<T>::adjust_eta(T& eta, size_t iter) {
+  switch (eta_schedule_) {
+    case LearningRateSchedule::CONSTANT:
+      eta = eta_;
+      break;
+    case LearningRateSchedule::TIME_BASED:
+      eta /= (1 + eta_decay_ * iter);
```
st-pasha, line 321: Did you want to have `eta = eta_ / (1 + eta_decay_ * iter)` here? At least that's what the documentation below implies.

st-pasha (approved):

src/core/models/dt_linearmodel.cc
st-pasha: Hmm, this is the book cited by the wikipedia article: [link](https://d1wqtxts1xzle7.cloudfront.net/57071216/Deep_Learning_A_Practitioners_Approach.pdf?1532536118=&response-content-disposition=inline%3B+filename%3DDeep_Learning_A_Practitioners_Approach.pdf&Expires=1617674321&Signature=VkDIq~IrPzVypJRPJ13LB8YfmiTF2JDcboL7CE5bkCGv0~56oQ2aMSd7GKl8nSoRZI7tT8o5JZtz6fOU6gVrV0kDPgiL3GwC8RLK7xB-Dn8KGDAQ02XR1wGH~HrH6e5v~PNhsm9QG~SUFlPp~AhfHJKcvGnP2lv6L4sQHwybsduRSRwa5AYnh8IqraWGUgSrdMphS1CWrqMFmScwmzS3Oao2f1~sU0EVwZi1bJp4xWAWBEouXOCg8n~dfB2~ph4~XXENHnfGoZvd3OX1xIePXTZAZ5L2O7FAkTYqYkc~eAiHN1abYbqiPmPAHVB2LgVbCuc3~Hb9nJgxkxbBE5bnYg__&Key-Pair-Id=APKAJLOHF5GGSLRBV4ZA). I do not see any "time-based learning rate" formulas there.
Here's another article on arxiv that discusses various LR schedules: [link](https://arxiv.org/pdf/1908.06477.pdf). The closest rule they have is the "inverse time" rule, which is simply `eta0 / (1 + decay * iter)^p` (where `p` is another parameter, which could be 1).

Could it be that the wikipedia article's author made a mistake?

oleksiyskononenko: I can't open the first link, but indeed, somewhere the time-based schedule is defined as `eta = eta0 / (1 + decay * iter)`:
https://keras.io/api/optimizers/learning_rate_schedules/inverse_time_decay/
https://dzlab.github.io/dltips/en/keras/learning-rate-scheduler/
etc.

But there are still some sources that say it is `eta = eta / (1 + decay * iter)`:
https://www.mdpi.com/2073-8994/12/4/660/pdf

I wonder if these are two different rules, or someimes `eta0` is used instead of `eta_n`.

st-pasha: You're right: in the article that you cited the "time-based" method is given in the form that you written.
However, they only mention it in passing, as an example; and they give as a source as _[11] Nesterov Y. Introductory Lectures on Convex Optimization: A Basic Course_. This book can be viewed on Google Books, and even though the preview is limited, the search function does not find any mention of "time-based" in that book. So, this still looks very suspicious...

oleksiyskononenko: The book [11] is from 2004, so no machine learning at that time. They don't say "learning rate schedule" instead they use the term "step-size strategy". That's probably the reason there is no "time-based schedule", it could just be formulated differently.

Anyways, it seems that many different methods could be used for adjusting the learning rate. May be there are even several varieties of the time-based schedule. For the moment, it looks like it is better to revert this schedule back to `eta = eta0 / (1 + decay * iter)`, because when I look at the popular codes like Keras this is what they're doing: https://github.com/keras-team/keras/blob/master/keras/optimizer_v1.py#L187-L193

st-pasha: Yeah, that's why it caught my eye too: the asymptotic behavior of `eta/=(1+decay*iter)` is way worse than even exponential.

PR #2941, opened 2021-04-14
[DOC] Select/filter Data in datatable

samukweku:
WIP #2604 
- Selecting data via rows and columns
- Filtering data via rows
- Examples

### DIFF
--- a/docs/manual/select_and_filter_data.rst
+++ b/docs/manual/select_and_filter_data.rst
@@ -5,7 +5,7 @@ Selecting Data
 Selecting Data - Columns
 ------------------------
 
-Column selection is via the `j` section in the ``DT[i, j, ...]`` syntax. First, let's construct a simple Frame::
+Column selection is via the `j` section in the ``DT[i, j, ...]`` syntax. First, let's construct a simple `Frame`_::
 
    >>> from datatable import dt, f
    >>> from datetime import date
@@ -428,15 +428,6 @@ All the selection options above (except boolean) are also possible via :ref:`f-e
     2 |        3     12.3  NA     
     3 |        4    -13    D      
    [4 rows x 3 columns]
-   ... DT[:, f['integers'::]] 
-      | integers   floats  strings
-      |    int32  float64  str32  
-   -- + --------  -------  -------
-    0 |        1     10    A      
-    1 |        2     11.5  B      
-    2 |        3     12.3  NA     
-    3 |        4    -13    D      
-   [4 rows x 3 columns]
    ... DT[:, f['integers':]]
       | integers   floats  strings
       |    int32  float64  str32  
@@ -611,7 +602,7 @@ Any acceptable sequence of positions is applicable here. Listed below are some (
 
 
 .. note:: The `set`_ sequence is not acceptable in the ``i`` or ``j`` sections.
-.. note:: Except for lists/tuples, all the other sequence types passed into the `i` section can only contain positive integers.
+.. note:: Except for ``lists``/``tuples``, all the other sequence types passed into the ``i`` section can only contain positive integers.
 
 Via Booleans
 ^^^^^^^^^^^^
@@ -831,7 +822,7 @@ Passing single integers into the ``i`` and ``j`` sections returns a scalar value
 Deselect Rows/Columns
 ---------------------
 
-Deselection of rows/columns is possible via `list comprehension`_/`generator expression`_, which either return boolean sequences, or the exact column names/positions/data types
+Deselection of rows/columns is possible via `list comprehension`_/`generator expression`_
 
 - Deselect a Single Column/Row::
 
--- a/docs/manual/select_and_filter_data.rst
+++ b/docs/manual/select_and_filter_data.rst
@@ -94,10 +94,10 @@ With position, you can select with a negative number - the column will be select
 .. note:: When selecting via column name or position, an error is returned if the name or position does not exist
 
    >>> DT[:, 5]
-   >>> ValueError: Column index 5 is invalid for a Frame with 4 columns
+   ValueError: Column index 5 is invalid for a Frame with 4 columns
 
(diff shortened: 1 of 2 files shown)


docs/manual/select_and_filter_data.rst
```diff
@@ -0,0 +1,1173 @@
...
+   >>> ---------------------------------------------------------------------------
+   >>> KeyError                                  Traceback (most recent call last)
+   >>> <ipython-input-64-0b4834f1b510> in <module>
+   >>> ----> 1 DT[:, 'categoricals']
+   >>>
+   >>> KeyError: Column categoricals does not exist in the Frame
+
+By Data Type
+^^^^^^^^^^^^
+Column selection is possible by using python's `built-in types`_:
```
oleksiyskononenko, line 114: I guess we should say here that not all the python's built-in types will work, but only those that have the corresponding datatable ltypes.

docs/manual/select_and_filter_data.rst
```diff
@@ -0,0 +1,1173 @@
...
+    0 | 2000-01-05         1     10    A      
+    1 | 2010-11-23         2     11.5  B      
+    2 | 2020-02-29         3     12.3  NA     
+    3 | NA                 4    -13    D          
+   [4 rows x 4 columns]
+
+
+.. note:: If the columns names are python keywords (``def``, ``del``, ...), the dot notation is not possible with :ref:`f-expressions`; you have to use the brackets notation to access these columns.
+
+.. note:: Selecting columns with ``DT[:, f[None]]`` returns an empty Frame. This is different from ``DT[:, None]``, which returns all the columns
```
oleksiyskononenko, line 471: @st-pasha any reason for `DT[:, None]` not returning an empty frame?

docs/manual/select_and_filter_data.rst
```diff
@@ -0,0 +1,1173 @@
...
+   -- + ----------  --------  -------  -------
+    0 | 2000-01-05         1     10    A      
+    1 | 2010-11-23         2     11.5  B      
+    2 | 2020-02-29         3     12.3  NA     
+    3 | NA                 4    -13    D      
+   [4 rows x 4 columns]
+
+However, when slicing via position, the columns are returned up to, but not including the final position; this is similar to the slicing pattern for Python's sequences
+
+   >>> DT[:, 1 : 3]
```
oleksiyskononenko, line 228: May be format slices as `x:y` (without spaces) then it would be easier to distinguish i and j selectors? Compare `DT[:, 1 : 3]` and `DT[:, 1:3]`.

docs/manual/select_and_filter_data.rst
```diff
@@ -0,0 +1,1173 @@
...
+    9 |
+   10 |
+   11 |
+   [12 rows x 0 columns]
+
+
+Delete Rows/Columns
+-------------------
+
+To actually delete the row/column, use the `del`_ statement; this is an in-place operation, and as such no reassignment is needed
```
oleksiyskononenko, line 1085: "To actually delete the row/column" -> "To actually delete a row or a column"?

docs/manual/select_and_filter_data.rst
```diff
@@ -0,0 +1,1173 @@
...
+
+
+
+.. _`pandas Series`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html
+.. _`numpy`: https://numpy.org/doc/stable/reference/generated/numpy.array.html#:~:text=array,-numpy.&text=An%20array%2C%20any%20object%20exposing,data%2Dtype%20for%20the%20array.
+.. _`range`: https://docs.python.org/3/library/functions.html#func-range
+.. _`generator expression`: https://docs.python.org/3/reference/expressions.html?highlight=generator#generator-expressions
+.. _`Frame`: https://datatable.readthedocs.io/en/latest/api/frame.html
+.. _`set`: https://docs.python.org/3/tutorial/datastructures.html#sets
+.. _`remove`: file:///home/sam/Documents/documentation/datatable/docs/_build/html/manual/f-expressions.html#modifying-a-columnset
```
oleksiyskononenko, line 1168: Some local link here.

docs/manual/select_and_filter_data.rst
```diff
@@ -0,0 +1,1173 @@
...
+
+.. note:: When selecting via column name or position, an error is returned if the name or position does not exist
+
+   >>> DT[:, 5]
+   >>> ---------------------------------------------------------------------------
+   >>> ValueError                                Traceback (most recent call last)
+   >>> <ipython-input-63-c0dafcf7de2e> in <module>
+   >>> ----> 1 DT[:, 5]
+   >>>
+   >>> ValueError: Column index 5 is invalid for a Frame with 4 columns
```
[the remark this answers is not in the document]
st-pasha: Note: the preferred  way of mentioning exceptions in the  documentation is like [here](https://datatable.readthedocs.io/en/latest/manual/comparison_with_pandas.html), see the bottom of the "Row and Column Selection" section.

st-pasha (approved):
Great Manual!
I've took a liberty to push some minor fixes, mostly because it's much faster for me to make the change than to describe it.

Here are some general pointers:
- the lines shouldn't end with whitespaces. My editor removes them automatically, perhaps  you can set up your editor to do the same?
- the `..note::` admonition is quite prominent, so it's probably best to only use it for things that we really want to emphasize;
- the n-dash can be typed as `--`;
- there shouldn't be any explicit internal links -- instead we can use roles like `:class:`  or `:ref:`. These work better with versioned documentation (we have separate documentation for latest, and also for each major release).

samukweku:
@st-pasha thanks for the feedback. I'll keep note of this for the next documentation I author.

If you have some time, could you have a look at this [issue](https://github.com/h2oai/datatable/issues/2946)  and kindly guide me with a code snippet or something, so I can get it rolling.

PR #2953, opened 2021-05-06
Add background for literal text in docs

st-pasha:

PR #2954, opened 2021-05-06
Fix code.literal style spillovers

st-pasha:
Follow-up for #2953

Also: fixed an issue where a comma would sometimes start a new line in a function's parameters list.


oleksiyskononenko (approved):

oleksiyskononenko:
I noticed the spillovers but they actually looked not so bad :)

Issue #2946, opened 2021-04-23
Create Equivalent Methods for Existing Datatable Functions


samukweku:
This is a follow-up from the [discussion here](https://github.com/h2oai/datatable/discussions/2863) on adding methods to `FExpr`.

The method chaining format would look like this ``DT[:, f[:].rowsum().sum()]``, which would be an equivalent for ``DT[:, dt.rowsum(dt.sum(f[:]))]``

samukweku:
@st-pasha , @oleksiyskononenko If you would be kind to show me how to pull this off for one of the existing functions (e.g `sum`). I tried using the same idea as for `len`, since I was not changing any C++ code, but it did not reflect when I restarted the python console

st-pasha:
Hi  @samukweku -- please see PR #2949 which implements the `sum` function. Most of it is self-explanatory: tests, documentation, etc. Most of the changes to C++ are in lines 331-345 of "fexpr.cc".

samukweku:
@st-pasha Thank you very much! I'll go through them tonight. Again, thank you!!

oleksiyskononenko:
@samukweku Do you want to create a list of methods in this issue that we need to implement for the existing dt functions? So that we keep track as to what is implemented and what is left.

samukweku:
@oleksiyskononenko we also need to make method chaining work on `dt.expr` as well

PR #2977, opened 2021-06-06
[ENH] methods for shift, first, last, sd, count

samukweku:
WIP #2946 

- Add methods for shift, first, last, sd, count.

### DIFF
--- a/src/core/expr/fexpr.cc
+++ b/src/core/expr/fexpr.cc
@@ -579,6 +579,91 @@ DECLARE_METHOD(&PyFExpr::rowsum)
     ->name("rowsum")
     ->docs(doc_rowsum);
 
+
+static const char* doc_sd =
+R"(sd()
+--
+
+Equivalent to :func:`dt.sd(self)`.
+)";
+
+oobj PyFExpr::sd(const XArgs&) {
+  auto sdFn = oobj::import("datatable", "sd");
+  return sdFn.call({this});
+}
+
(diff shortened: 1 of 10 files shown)


src/core/expr/fexpr.cc
```diff
@@ -579,6 +579,91 @@ DECLARE_METHOD(&PyFExpr::rowsum)
...
+static const char* doc_shift =
+R"(shift()
+--
+
+Equivalent to :func:`dt.shift(self)`.
+)";
+
+oobj PyFExpr::shift(const XArgs&) {
+  auto shiftFn = oobj::import("datatable", "shift");
+  return shiftFn.call({this});
```
st-pasha, line 609: `dt.shift()` takes optional parameter `n` (the amount of shift), so this method should do the same.

samukweku: Hi @st-pasha haven't had time to look at this due to other constraints ... I dabbled a bit, but couldnt get to include the extra parameter. Could you kindly show me how to do it. Thanks.

st-pasha: done

st-pasha (approved):

samukweku:
@st-pasha thanks! What does the `?` in `args[0]?` signify?

st-pasha:
It's just C syntax for ternary-if condition: `a? b : c` is the same as python's `b if a else c`.

PR #2981, opened 2021-06-12
Implement type casts from time64 into other types

### LINKED ISSUE #2978: Implement  type casts time64 -> (any)

st-pasha:
Closes #2978


tests/types/test-time64.py
```diff
@@ -488,3 +488,78 @@ def test_cast_string_to_time64(ttype):
...
+                   d(1969, 12, 31, 0, 0, 0),
+                   d(1962, 4, 12, 3, 52, 27, 458000),
+                   None])
+    DT[0] = st
+    assert_equals(DT, dt.Frame(["2091-11-28T15:51:27.31",
+                                "1970-01-03T21:05:02.475",
+                                "1969-12-31T23:59:59.999999",
+                                "1969-12-31T00:00:00",
+                                "1962-04-12T03:52:27.458",
+                                None], stype=st))
```
oleksiyskononenko, line 557: What is the reason we're using "T" as the date/time separator and not just a space?

oleksiyskononenko (approved):

tests/types/test-time64.py
st-pasha: It's in the ISO-8601 standard: https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations

oleksiyskononenko: Yes, but the space seems to be more readable and is allowed through the RFC 3339:

```
Separating date and time parts with other characters such as space is not allowed in ISO 8601, but allowed in its profile RFC 3339.[34]

...

RFC 3339". 2002-07-01. "5.6. NOTE: ISO 8601 defines date and time separated by "T". Applications using this syntax may choose, for the sake of readability, to specify a full-date and full-time separated by (say) a space character."
```

st-pasha: ISO = International Standardization Organization
RFC = Request For Comments
So, I take ISO to be more authoritative than RFC. While I agree that space separator is easier on the eyes, the older ISO standard did allow both a "T" and a space here, but in the end, they changed to forbid the space separator. This means, they must have had a good reason for such a change.

oleksiyskononenko: Frankly I can't find that reason, but what I found is this quote from  ISO-8601:

```
NOTE By mutual agreement of the partners in information interchange, the character [T] may be omitted in applications where there is no risk of confusing a date and time of day representation with others defined in this International Standard.
```

So it seems that they wanted to prevent confusion between reading in date and time. But this quote is more about fully eliminating the `T`, if `T` is replaced by a space I don't see a source of confusion. May be there is a reason, just can't get it. Having `T` instead of a space looks kind of messy imho.

st-pasha: I don't know, the EDTF standard also says T: https://www.loc.gov/standards/datetime/

st-pasha: Even god-forsaken `ГОСТ Р 7.0.64-2018` specifies using T: https://docs.cntd.ru/document/1200159341
What other standards are out there?

oleksiyskononenko: Not sure :) Do we also require the `T` separator when parsing strings into time? It seems that pandas, for example, allows both `T` and ` ` separators.

st-pasha: when parsing we allow both space and T

PR #2991, opened 2021-06-16
Add function `ymdt()`

### LINKED ISSUE #2985: Add function ymdt()
The function should construct a time64 column from individual components: year, month, day, hour, minute, second.

st-pasha:
Closes #2985


docs/api/time.rst
```diff
@@ -35,6 +35,10 @@ datatable.time
    * - :func:`ymd(y,m,d)`
      - Create a ``date32`` column from year, month, and day components.
 
+   * - :func:`ymdt(y,m,d,H,M,S)`
+     - Create a ``time64`` column from year, month, day, hour, minute, and
+       second components.
```
oleksiyskononenko, line 40: Do we want to mention nanoseconds here?

st-pasha: Nanoseconds is an optional parameter, that I don't expect to be used very often. Here in the overview we show the minimal signature of the function.

PR #2993, opened 2021-06-16
Visually separate Type variables from properties/methods in docs

st-pasha:
So, the problem with documenting `Type` class is that it has static members such as `Type.str32` or `Type.float64`, and also regular properties such as `Type().name` or `Type().max`. These obviously have different usage (although technically static members are accessible on instances too, it's not how they are supposed to be used), and therefore mixing them up in the documentation creates lots of confusion. 

This PR addresses this problem by adding a small star icon next to each static member, to indicate that they are "special".

### DIFF
--- a/docs/api/type.rst
+++ b/docs/api/type.rst
@@ -42,19 +42,19 @@
 .. toctree::
     :hidden:
 
-    bool8    <type/bool8>
-    date32   <type/date32>
-    float32  <type/float32>
-    float64  <type/float64>
-    int8     <type/int8>
-    int16    <type/int16>
-    int32    <type/int32>
-    int64    <type/int64>
-    min      <type/min>
-    max      <type/max>
-    name     <type/name>
-    obj64    <type/obj64>
-    str32    <type/str32>
-    str64    <type/str64>
-    time64   <type/time64>
-    void     <type/void>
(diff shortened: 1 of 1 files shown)


oleksiyskononenko:
Is this change for developers only?

oleksiyskononenko (approved):

st-pasha:
It's public documentation, the change affects the menu on the left side of the page.

oleksiyskononenko:
Then may be we should add a note as to what these star symbols mean?

st-pasha:
where would you add such a note?

oleksiyskononenko:
Actually, I would just make them italic on the `datatable.Type` page, where would list them as static explicitly. Then I would also make them italic on the left menu to make it obvious why they are visually different.

st-pasha:
italic does not provide sufficient visual contrast:
Screen Shot 2021-06-18 at 10 42 07 AM

oleksiyskononenko:
We can try bold italic and make it bold italic (the highlighted text) on the documentation page? Also adding a note that this property is static. RIght now it is not really obvious what the `*` symbol means.
 
Screen Shot 2021-06-18 at 3 48 19 PM

PR #2997, opened 2021-06-17
Add documentation for .countna(), .max(), .min(), .mean(), .mode(), .nmodal() and .nunique() frame methods

oleksiyskononenko:
Also, add documentation for the corresponding one-column frame methods. Some other minor improvements and rearrangements.

### DIFF
--- a/docs/api/frame.rst
+++ b/docs/api/frame.rst
@@ -140,12 +140,6 @@
         * - :meth:`.countna1()`
           - Count missing values for a one-column frame and return it as a scalar.
 
-        * - :meth:`.min()`
-          - Find the smallest element for each column in the frame.
-
-        * - :meth:`.min1()`
-          - Find the smallest element for a one-column frame and return it as a scalar.
-
         * - :meth:`.max()`
           - Find the largest element for each column in the frame.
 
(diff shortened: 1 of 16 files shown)


docs/api/frame.rst
```diff
@@ -242,16 +293,30 @@
...
+    .min1            <frame/min1>
+    .mode            <frame/mode>
+    .mode1           <frame/mode1>
     .names           <frame/names>
     .ncols           <frame/ncols>
+    .nmodal          <frame/nmodal>
+    .nmodal1         <frame/nmodal1>
     .nrows           <frame/nrows>
+    .nunique         <frame/nunique>
+    .nunique1        <frame/nunique1>
```
st-pasha, line 319: Methods should have parentheses, e.g. `.copy()`, while attributes should not, e.g. `.names`.
Seems you're missing lots of parentheses here.

st-pasha (approved):

docs/api/frame.rst
oleksiyskononenko: Yep, let me change it. I wonder why it doesn't fail though...

PR #3000, opened 2021-06-17
Renamed "types" arguments in Frame constructor

### LINKED ISSUE #2986: type= argument in Frame()
Currently, `Frame()` constructor has argument `stype=`. We should rename this into `type=` (but allow the old name too), and support `Type` values there.

st-pasha:
Two arguments in Frame's constructor were renamed:
   - `stypes=` into `types=`, and
   - `stype=` into `type=`.
 
The old arguments are still available and continue to work as before, although they were excluded from the documentation.

The new parameters now support `Type` values, and also pyarrow's types (in addition to all the types that were supported before).

Some of the tests were converted into using the new `types=` argument, while others continue using `stypes=`. This ensures that both the new and the old arguments work.

Closes #2986

### DIFF
--- a/docs/_ext/xfunction.py
+++ b/docs/_ext/xfunction.py
@@ -1348,7 +1348,9 @@ class XversionActionDirective(SphinxDirective):
                 reftarget=target))
         node = xnodes.div(node, classes=["x-version", action])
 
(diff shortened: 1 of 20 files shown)


st-pasha:
Yay! 3000

PR #3002, opened 2021-06-18
New mechanism for writing documentation

st-pasha:
This PR changes the way we write documentation: instead of embedding the API docs into the code, the documentation can now be written entirely in `doc/` folder, and it will be brought into C++ through an automatic process.

Specifically, the proposed process is as follows:
  - the API documentation is written entirely within the `.rst` files in the `docs/` folder;
  - the ``.. xfunction`` (&co) directive now has an additional parameter `:cvar:` which describes where this documentation should be copied in order to be available from C++;
  - the auto-generator script, which runs automatically on `make build` (or `make debug`), copies the content of the xfunction directive into the specified variable name. The name must be declared in the "documentation.h" file;
  - test `test_documentation_h()` verifies that the correspondence is unbroken: that each variable that is declared in "documentation.h" has a corresponding ":cvar:" declaration within the documentation, and vice versa.

The advantages of the new approach are the following:
  - the documentation is declared within `.rst` files, which allows a text editor to properly highlight the syntax;
  - updating existing documentation is now easier, as it doesn't require editing source code files;
  - updating documentation is now less scary for newcomers;
  - the code base is not over-encumbered with documentation text, and can now have a smaller LoC count overall;
  - it is now possible to extend this approach and write additional filters that would pre-process the RST text, converting it into a format that might be more suitable for pure-text python.

This PR only adds a mechanism for writing new documentation; the actual documentation migration can happen later on. Currently, only 2 docstrings were converted into the new format: `Frame.head()` and `Frame.tail()`, as a proof of concept.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ temp/
 
 # Auto-generated files
 src/datatable/_build_info.py
+src/core/documentation.cc
 
 # vscode settings
 .vscode/settings.json
--- a/ci/ext.py
+++ b/ci/ext.py
@@ -354,9 +354,22 @@ def build_extension(cmd, verbosity=3):
                 "-Wno-unknown-pragmas"
             )
 
+    ext.add_prebuild_trigger(generate_documentation)
+
     # Setup is complete, ready to build
     ext.build()
     return ext.output_file
+
+
+def generate_documentation(ext):
+    hfile = "src/core/documentation.h"
+    ccfile = "src/core/documentation.cc"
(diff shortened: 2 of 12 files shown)


oleksiyskononenko:
So let's say I'm working on adding a new function: implemented it in C++ and added documentation into the RST file. In order to build the code I will first need to build the docs?

oleksiyskononenko:
Sorry, I misread the process description, please disregard my question.

samukweku (approved):

oleksiyskononenko (approved):

PR #3004, opened 2021-06-18
Add Travis build for macOS arm64

st-pasha:
Try to build on Travis on macOS arm64 architecture, see #3003.


st-pasha:
So, turns out Travis does not support arm64 architecture for macs. Closing.

PR #3005, opened 2021-06-19
Refactor documentation for Frame class

st-pasha:
Following PR #3002, here we move documentation for all properties/methods of the Frame class into docs/ folder.


docs/api/frame/ncols.rst
```diff
@@ -1,4 +1,20 @@
...
+    Number of columns in the frame.
+
+    Parameters
+    ----------
+    return: int
+        The number of columns can be either zero or a positive integer.
+
+    Notes
+    -----
+    The expression `len(DT)` also returns the number of columns in the
```
samukweku, line 15: Any reason why `len(DT)` is not preferred? Adding the reason may be helpful, since using `len(DT)` may come to mind more naturally than `.ncols`?

samukweku (approved):

docs/api/frame/ncols.rst
st-pasha: It's a matter of preference, really. To me, `len(DT)` sounds ambiguous, because it's unclear whether "length" refers to the columns or rows direction. Also, it is unclear from this expression how to compute the length in the opposite direction (i.e. rows).

The main reason why we support `len(DT)` at all is so that the *-expansion syntax would work: `*DT` converts a frame into a sequence of columns.

PR #3013, opened 2021-06-23
Fix `sd()`, `sum()` and `mean()` Frame methods when called on date and time columns 

### LINKED ISSUE #2988: Unexpected output of .mean() and .sd() on time64/date32 columns
```
>>> from datatable import dt
>>> DT = dt.Frame(["2021-01-01T20:21:01"])
>>> DT[0] = dt.time64
>>> DT[0] = dt.Type.time64
>>> DT.mean()
   |          C0
   |     float64
-- + -----------
 0 | 1.60953e+18
[1 row x 1 column]
>>> DT.sd()
   |      C0
   | float64
-- + -------
 0 |       0
[1 row x 1 column]
```

oleksiyskononenko:
Also, add and improve the relevant tests.

Closes #2988

### DIFF
--- a/src/core/stats.cc
+++ b/src/core/stats.cc
@@ -1341,14 +1341,13 @@ py::oobj Stats::get_stat_as_pyobject(Stat stat) {
       return pywrap_stat<size_t>(stat);
     }
     case Stat::Mean: {
-      switch (dt::stype_to_ltype(column->stype())) {
-       case dt::LType::DATETIME: {
-          double value;
-          bool isvalid = get_stat(stat, &value);
-          if (!isvalid) return py::None();
-          return py::odatetime(static_cast<int64_t>(value));
-        }
-        default: return pywrap_stat<double>(stat);
+      if (column->type().is_temporal()) {
+        double value;
+        bool isvalid = get_stat(stat, &value);
+        if (!isvalid) return py::None();
+        return py::odatetime(static_cast<int64_t>(value));
+      } else {
+        return pywrap_stat<double>(stat);
       }
     }
     case Stat::Sum:
@@ -1443,15 +1442,14 @@ Column Stats::get_stat_as_column(Stat stat) {
   switch (stat) {
     case Stat::NaCount:
     case Stat::NUnique:
-    case Stat::NModal: {
-      return colwrap_stat<size_t, int64_t>(stat, dt::SType::INT64);
-    }
-    case Stat::Mean:
-      switch (column->stype()) {
-        case dt::SType::DATE32:
-        case dt::SType::TIME64:  return colwrap_stat<double, int64_t>(stat, dt::SType::TIME64);
-        default:                 return colwrap_stat<double, double>(stat, dt::SType::FLOAT64);
+    case Stat::NModal: return colwrap_stat<size_t, int64_t>(stat, dt::SType::INT64);
+    case Stat::Mean: {
+      if (column->type().is_temporal()) {
+        return colwrap_stat<double, int64_t>(stat, dt::SType::TIME64);
+      } else {
+        return colwrap_stat<double, double>(stat, dt::SType::FLOAT64);
       }
+    }
     case Stat::Sum:
     case Stat::StDev:
     case Stat::Skew:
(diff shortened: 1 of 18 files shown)


src/core/stats.cc
```diff
@@ -1336,8 +1340,18 @@ py::oobj Stats::get_stat_as_pyobject(Stat stat) {
     case Stat::NModal: {
       return pywrap_stat<size_t>(stat);
     }
+    case Stat::Mean: {
+      switch (dt::stype_to_ltype(column->stype())) {
```
st-pasha, line 1344: This should really be `if (column->type()->is_temporal()) { ... }`,

where `Type::is_temporal()` is renamed from `Type::is_time()` (the current name is too confusing).

src/core/stats.cc
```diff
@@ -1432,8 +1446,13 @@ Column Stats::get_stat_as_column(Stat stat) {
     case Stat::NModal: {
       return colwrap_stat<size_t, int64_t>(stat, dt::SType::INT64);
     }
-    case Stat::Sum:
     case Stat::Mean:
+      switch (column->stype()) {
+        case dt::SType::DATE32:
+        case dt::SType::TIME64:  return colwrap_stat<double, int64_t>(stat, dt::SType::TIME64);
```
st-pasha, line 1452: same here

src/core/stats.cc
```diff
@@ -1463,6 +1482,25 @@ Column Stats::get_stat_as_column(Stat stat) {
...
+double TimeStats::stdev (bool* isvalid) { return _invalid<double>(isvalid);}
+double DateStats::sum (bool* isvalid) { return _invalid<double>(isvalid); }
+double DateStats::stdev (bool* isvalid) { return _invalid<double>(isvalid); }
+
+
+double DateStats::mean (bool* isvalid) {
+  if (!is_computed(Stat::Mean)) {
+    compute_moments12();
+    // Conversion from days to nanoseconds for further casting to `time64`
+    _mean *= 86400000000000;
```
st-pasha, line 1499: Replace this magic number with a named constant: `constexpr int64_t NANOSECONDS_IN_DAY = 24 * 3600 * 1000000000`.

tests/types/test-date32.py
```diff
@@ -113,6 +114,58 @@ def test_date32_minmax():
...
+
+def test_date32_na_stats():
+    src = [None,
+           d(2000, 10, 18),
+           d(2010, 11, 13),
+           d(2020, 2, 29),
+           None]
+    DT = dt.Frame(src)
+    assert DT.sd1() == None
+    assert DT.sum1() == None
```
st-pasha, line 144: `is None` is better than `== None` (otherwise python linter will complain).

src/core/stats.cc
oleksiyskononenko: Yeah, I've been refactoring this `switch` and forgot to replace it with a simple `if` at the end.

src/core/stats.cc
oleksiyskononenko: Don't you have this number already somewhere in the code? For example, when casting `date32` to `time64` or something similar.

tests/types/test-date32.py
oleksiyskononenko: Yeah, it's not the first time I forget to use `is` for `None`. Everytime I correct it, I find several other places we use `==` in the code. May be there should be an automated tool for this correction in the editor :)

src/core/stats.cc
oleksiyskononenko: Btw, `TypeImpl::is_time()` should be also replaced with `TypeImpl:: is_temporal()`.

src/core/stats.cc
st-pasha: Yes, but it's not a globally-defined constant.

oleksiyskononenko: I can see it now. Also not sure why this [one](https://github.com/h2oai/datatable/blob/main/src/core/types/type_time.cc#L87) is called `SECONDS` and not `NANOSECONDS`.

oleksiyskononenko: We have quite a lot of these actually: https://github.com/h2oai/datatable/blob/main/src/core/python/date.cc#L122 At some point may be we should create something like `time_consts.h` with all these time conversion constants.

st-pasha (approved):

PR #3015, opened 2021-06-25
String slicing

### LINKED ISSUE #1667: FR: Apply a slice to each string in a string column
Hi guys,

I wonder how can I implement the equivalent datatable function of doing the following in pandas dataframe?

`df[:,'col'] = df[:, 'col'][-8:]`

Basically, I am trying to parse all my `df['col']` to get the last 8 letters. But I am facing the error above. Surprisingly, the below code works but it takes too long.

```
for each_row in range(df_len):
	df[each_row, 2] = df[each_row, 2][-8:]
```
Thank you.

st-pasha:
This implements string-slicing function `FExpr[a:b:c]`, or alternatively as a regular function `dt.str.slice(col, start=a, stop=b, step=c)`.

The new function supports both constant slices and expression-slices. 

The reason to have both a function and a slice expression `[a:b]` is that a function is much easier to find in the documentation (esp. once we have more string functions in `dt.str` module), whereas `[]`-based slices are easier to write.

Closes #1667

### DIFF
--- a/tests/expr/str/test-slice.py
+++ b/tests/expr/str/test-slice.py
@@ -55,8 +55,56 @@ def test_slice_wrong_types2(ttype):
     [slice(None, 5), slice(3, None), slice(None, None, 1), slice(-1, None),
      slice(2, 7), slice(3, -1), slice(5, -8), slice(-8, 5), slice(None, -2)])
 def test_slice_normal(slice):
-    src = ["Hydrogen", "Helium", "Lithium", "Berillium", "Boron"]
+    src = ["Hydrogen", "Helium", "Lithium", "Berillium", "Boron", "Carbon",
+           None, "Praseodymium"]
     DT = dt.Frame(A=src)
     RES = DT[:, f.A[slice]]
-    EXP = dt.Frame(A=[s[slice] for s in src])
+    EXP = dt.Frame(A=[None if s is None else s[slice] for s in src])
+    assert_equals(RES, EXP)
+
+
+@pytest.mark.parametrize('seed', [random.getrandbits(32) for _ in range(10)])
+def test_slice_unicode_random(seed):
+    random.seed(seed)
+    src = [
+        "По улиці вітер віє",
+        "Та сніг замітає.",
+        "По улиці попідтинню",
+        "Вдова шкандибає",
+        "Під дзвіницю, сердешная,",
+        "Руки простягати",
+        "До тих самих, до багатих,",
+        "Що сина в солдати",
+        "Позаторік заголили.",
+        "А думала жити...",
+        "Хоч на старість у невістки",
+        "В добрі одпочити.",
+        "Не довелось. Виблагала",
+        "Тую копійчину...",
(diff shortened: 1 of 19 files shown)


tests/expr/str/test-slice.py
```diff
@@ -0,0 +1,174 @@
...
+     dt.Type.date32, dt.Type.time64, dt.obj64])
+def test_slice_wrong_types(ttype):
+    msg = "Slice expression can only be applied to a column of string type"
+    with pytest.raises(TypeError, match=msg):
+        DT = dt.Frame(A=[None], type=ttype)
+        DT[:, f.A[:1]]
+
+
+@pytest.mark.parametrize('ttype',
+    [dt.bool8, dt.float32, dt.float64, dt.str32, dt.str64,
```
oleksiyskononenko, line 69: Is there any reason not to allow `bool8`? This is not a super useful case, of course.

oleksiyskononenko (approved):

tests/expr/str/test-slice.py
```diff
@@ -0,0 +1,174 @@
...
+    assert_equals(RES, EXP)
+
+
+def test_slice_expression():
+    DT = dt.Frame(A=["Monday", "Tuesday", "Wednesday", "Thursday", "Friday"],
+                  s=[0, 3, 2, 1, -4],
+                  e=[5, 4, 0, -1, None])
+    RES = DT[:, (f.A)[f.s:f.e]]
+    EXP = dt.Frame(A=["Monda", "s", "", "hursda", "iday"])
+    assert_equals(RES, EXP)
```
oleksiyskononenko, line 166: May be add a test that the `void` column type is supported in expressions?

tests/expr/str/test-slice.py
st-pasha: I don't see how this would ever be useful. In general, we have bools, integers, and floats, which are all considered numeric, but otherwise in separate categories.

tests/expr/str/test-slice.py
st-pasha: done

tests/expr/str/test-slice.py
oleksiyskononenko: For instance, we have some data that could be both "something" and "@something". And there is a boolean `start` that tells if `s[0] == "@"`. Then we strip the actual data by slicing from the boolean `start`.

st-pasha: That would have been a rather confusing expression: `f.s[f.s[0] == "@" :]`. But you can always force a boolean expression into integer with `+`: `f.s[+(f.s[0] == "@") :]`, though, admittedly, this doesn't aid readability.

oleksiyskononenko: The first one doesn't look very unreadable to me. Anyways, if we don't support booleans here - that's not a big deal. If we support, this may be useful in some cases when one wanna make a shortcut.

Issue #2999, opened 2021-06-17
Add missing statistical reducers


oleksiyskononenko:
RIght now the following reducers are only available as the corresponding `Frame` methods, but can't be used in the square brackets syntax as `dt.*()`:

- [ ] `mode()`
- [ ] `nmodal()`
- [ ] `nunique()`
- [ ] `countna()`

pradkrish:
I accidentally converted the last task into an issue and then closed immediately realising the mistake. It is still linked here somehow, anyway to remove it? Sorry for the inconvenience.

st-pasha:
That's ok. Looks like it's a new feature they added only a few days ago. Good to know!

pradkrish:
Should I link the PR for `countna` to this issue or shall I reopen the closed issue and link it to that one?

oleksiyskononenko:
@pradkrish Yeah, seems to be a new feature. I've restored this issue to its original state, in this case I guess it is better to keep track of the missing reducers from one issue rather than distributing it into four smaller ones. For other issues, we may want to use the feature you just discovered :) 

On your PR just replace "Related" with "WIP for".

Issue #3023, opened 2021-06-26
statistical reducers output mismatch  


pradkrish:
The output for statistical reducers when used in the square brackets syntax does not match that of Frame methods. 

```python 
srcs =  ["a", "bc", "def", None, -2.5, 3.7] / dt.obj64
df = dt.Frame(srcs)
RES1 = df.min()
RES2 = df[:, dt.min(f[:])] # What is the expected output here?
```

 RES1 and RES2 should match but RES2 gives `TypeError: Unable to apply reduce function min() to a column of type obj64` instead. This mismatch is also seen for other functions like `max` and `mean`. I haven't checked thoroughly but it's likely to be seen for other reducers too. As seen in the comments, what is the expected output at line 4?

st-pasha:
This is by design. The stats functions such as `mean()` or `max()`, when called as methods of a Frame, will apply to each column of that frame and produce a frame of shape `ncols x 1` in the result. If the stat function is not applicable to a particular column, there will be an NA in that place in the result.

The regular stat functions such as `dt.mean()` or `dt.max()` behave differently: they can apply to one or more columns given as the argument, but if one of those columns has an incorrect type, an exception will be raised. This is better from the user's perspective: if a column has an unexpected type, it's better to catch that early.

pradkrish:
In that case, is it okay if the outputs for `min`, and other reducers, calculated in two different ways do not match if the frames are of type `obj64`?

oleksiyskononenko:
@st-pasha Frankly, I don't understand why behavior should be different. Frame can also consists from one or more columns, and a new frame can easily be constructed that would only contain columns with the valid column types...

Btw, output in the first case looks strange:

```
>>> RES1 = df.min()
>>> RES1
   | C0       
   | obj64    
-- + ---------
 0 | <unknown>
[1 row x 1 column]
```

From what you're saying, I would expect that to be `NA`.

st-pasha:
Yeah, this simply because our handling of Obj columns is really bad:
```
>>> dt.Frame(range(5), type=object)
   | C0       
   | obj64    
-- + ---------
 0 | <unknown>
 1 | <unknown>
 2 | <unknown>
 3 | <unknown>
 4 | <unknown>
[5 rows x 1 column]
```
We simply don't have any better way of rendering these obj values.

Ideally, they should be repr-d here, but we don't have that logic implemented yet.

pradkrish:
[...]

PR #3024, opened 2021-06-26
make .countna callable from inside the square brackets

pradkrish:
WIP for #2999


pradkrish:
Currently, a couple of tests fail for reasons mentioned in #3023.

pradkrish:
The behaviour of `dt.countna` and  `DT[:, dt.counta(..)]` is different for `obj64` columns but how to deal with `obj64` columns is not clear at the present moment. Refer to #3023

st-pasha:
We have not defined the semantics of reducer functions for obj64 columns; so it's ok for now to raise a TypeError for columns of that type.

st-pasha (commented):
**Documentation**
We have recently (#3002) moved to a new convention where the documentation for any new function should primarily reside inside `.rst` files, and then it will be copied into python during the build stage.
As such, can you please move the contents of `doc_countna` variable into the `docs/api/dt/countna.rst` file, and instead declare a `doc_dt_countna` variable in `documentation.h`.

pradkrish:
@st-pasha any idea why the test-docs are failing? :thinking:

docs/api/dt/countna.rst
```diff
@@ -0,0 +1,71 @@
...
+    -- + -----
+     0 |     2
+    [1 row x 1 column]
+
+
+
+See Also
+--------
+
+- :func:`count()` -- function to count the number of non-missing values.
```
st-pasha: The text of the documentation needs to be indented, so that it's "inside" the `..xfunction` block. (In rst, like in python, blocks are denoted via indentation). Without this, the generated variable `doc_dt_countna` will end up empty, as you can see in the generated `documentation.cc` file.

st-pasha:
Also, please go into the docs folder and run `make html` to generate the docs. Currently, it will issue you an error about "document not in a toctree". This means you need to tell sphinx where this `countna.rst` file should be in the table of contents. Since this is a top-level `dt.` function, put it into `index-api.rst`: both in the "list of functions" section, and in the `..toctree` directive at the bottom.

docs/api/dt/countna.rst
pradkrish: this was already fixed in a later commit.

st-pasha:
Let's wait with this PR until after the release

pradkrish:
> Let's wait with this PR until after the release

no worries, good luck with the release :)

docs/api/dt/countna.rst
```diff
@@ -0,0 +1,71 @@
...
+    cols: Expr
+        Input columns.
+
+    return: Expr
+        f-expression having one row, and the same names and number of columns
+        as in `cols`. All the returned column stypes are `int64`.
+
+    except: TypeError
+        The exception is raised when one of the columns from `cols`
+        has a non-numeric and non-string type.
```
st-pasha, line 22: We should be able to count NAs in all columns except `obj64`.

docs/api/dt/countna.rst
```diff
@@ -0,0 +1,71 @@
...
+        -- + -----  -----  -----
+         0 |     1     NA      1
+         1 |     1      2      2
+         2 |     2      3      1
+         3 |    NA      4      1
+         4 |     1     NA      2
+         5 |     2      5      4
+        [6 rows x 3 columns]
+
+    Get the countna of all rows:
```
st-pasha, line 48: *columns

Also, it's probably better to say "count of NAs" instead of "countna" here.

docs/api/index-api.rst
```diff
@@ -156,21 +156,23 @@ Functions
...
-      - Find the smallest element per a column
+      - Find the smallest element per column
     * - :func:`sd()`
-      - Calculate the standard deviation per a column
+      - Calculate the standard deviation per column
     * - :func:`sum()`
-      - Calculate the sum of all values per a column
+      - Calculate the sum of all values per column
+    * - :func:`countna()`
+      - Count the number of NA values per column
```
st-pasha, line 175: Please, keep this list in alphabetical order.

docs/releases/v1.0.0.rst
```diff
@@ -179,6 +179,7 @@
     -[api] Method :meth:`dt.FExpr.re_match` has been deprecated and replaced
       with a function :func:`dt.re.match()`. [#3017]
 
+    -[enh] Added reducer function :func:`countna` [#2999]
```
st-pasha, line 182: This would have to be moved to file `v1.1.0.rst`.

src/core/documentation.h
```diff
@@ -23,6 +23,7 @@
 #define dt_DOCUMENTATION_h
 namespace dt {
 
+extern const char* doc_dt_countna;
```
st-pasha, line 26: Please move this down into the block of all `doc_dt_*` constants, alphabetized.

src/core/expr/head_reduce_unary.cc
```diff
@@ -1573,7 +1618,8 @@ Workframe Head_Reduce_Unary::evaluate_n(
       case Op::MIN:
       case Op::MAX:
       case Op::FIRST:
-      case Op::LAST:   fn = compute_gfirstlast; break;
+      case Op::LAST:
+      case Op::COUNTNA:fn = compute_gfirstlast; break;
```
st-pasha, line 1622: I don't think this will work. We need a test for how `countna()` behaves when computed within group-by.

tests/test-reduce.py
```diff
@@ -551,3 +580,18 @@ def test_corr_multiple():
...
+#-------------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("src", srcs_all)
+def test_dt_count_na(src):
+    dt0 = dt.Frame(src)
+    ans = dt0.countna()
+    dtr = dt0[:, dt.countna(f[:])]
+    frame_integrity_check(dtr)
+    assert_equals(ans, dtr)
```
st-pasha, line 597: Can you please move this test into a file dedicated to `countna()` function? It should be "tests/dt/test-countna.py".
We're trying to [move to a system](https://github.com/h2oai/datatable/issues/3030) where each function has a dedicated test file, linked from the corresponding `.rst` file. This will let us see clearer which functions are well-tested, and which aren't.

st-pasha (changes requested):
There are few minor things to fix, as I mention in the comments. Though I suspect that groupby countna may not be working -- and it won't be exactly minor. Still, `countna()` is conceptually very close to `count()`.

Another thing that I didn't mention, is that we want to have this and other reducer functions converted into FExprs (#2562). Though maybe this could be left to a future time.

docs/releases/v1.0.0.rst
st-pasha: (You'll need to merge `main` into this branch)

src/core/expr/head_reduce_unary.cc
pradkrish: I agree. I didn't look properly into this. Can you give an example for the expected behaviour of `countna()` in the presence of grouby and contrast that output with how `count()` would behave in the presence of groupby, thanks.

st-pasha: `count(f.X)` returns the number of valid (non-NA) elements in column X in each group. The `countna(f.X)` is supposed to return the number of invalid (NA) elements in column X in each group. The `count()` function without an argument returns the total number of elements in each group. Thus, the following invariant must hold:
```
countna(f.X) == count() - count(f.X)
```

pradkrish: Let me make sure we are on the same page here. The following frame
```python
df = dt.Frame({"group1": ["A", "A", None, "B", "B", "A"],
               "group2": [1, 0, 1, 1, 1, 1],
               "var": [343, 345, 567, 345, 2, 212]})

print(df[:, dt.count()])
print(df[:, dt.count(f[:])])

print(df[:, dt.countna()])
print(df[:, dt.countna(f[:])])

```
gives

```python
   | count
   | int32
-- + -----          # df[:, dt.count()]
 0 |     6
[1 row x 1 column]

   | group1
   |  int64
-- + ------          # df[:, dt.count(f.group1)]       
 0 |      5
[1 row x 1 column]

   |   C0
   | void
-- + ----          # df[:, dt.countna()]
 0 |   NA
[1 row x 1 column]

   | group1
   |  int64
-- + ------          # df[:, dt.countna(f.group1)]
 0 |      1
[1 row x 1 column]

```

The relationship `countna(f.X) == count() - count(f.X)` does hold. The only weird result here is for `dt.countna()`, which takes the `else` path in head_reduce_unary.cc:1614 and it needs fixing. is that correct?

st-pasha: Yeah, I don't think `dt.countna()` without an argument makes sense. Should throw a type error immediately.

But my original concern was about `countna()` in the presence of a groupby clause, i.e. the following two expressions should produce same results:
```
DT[:, [dt.countna(f.var), dt.count() - dt.count(f.var)], by(f.group1)]
``` 
(throw some NAs into the column `var` so that the example is not too trivial).

pradkrish: This is what I see now
```python
df = dt.Frame({"group1": ["A", "A", None, "B", "B", "A"],
                "var": [343, None, 567, None, 2, 212]})
                                                           
print(df[:, [dt.countna(f.var), dt.count() - dt.count(f.var)], dt.by(f.group1)])
```
gives
```python
  | group1    var     C0
   | str32   int64  int64
-- + ------  -----  -----
 0 | NA          0      0
 1 | A           1      1
 2 | B           1      1
[3 rows x 3 columns]

```
isn't this the expected output? 

Perhaps you can help me understand this. One thing I am not sure is under what circumstances the program takes the `else` path in `head_reduce_unary.cc:1614`? Because that was your initial concern in the review and I am not sure when that happens.

st-pasha: Hmm, could it be group function applied to another group function?

pradkrish: not sure, perhaps an example to illustrate what you mean?

st-pasha: Suppose you have `DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])`, then simple aggregate produces
```
>>> DT[:, [f.V, dt.mean(f.V)], by(f.G)]
   |     G      V      V.0
   | int32  int32  float64
-- + -----  -----  -------
 0 |     1     NA       NA
 1 |     1     NA       NA
 2 |     1     NA       NA
 3 |     2     NA        4
 4 |     2      3        4
 5 |     2      5        4
[6 rows x 3 columns]
```
Now imagine we applied `countna()` to both of these columns:
```
>>> DT[:, [dt.countna(f.V), dt.countna(dt.mean(f.V))], by(f.G)]
   |     G      V    V.0
   | int32  int64  int64
-- + -----  -----  -----
 0 |     1      3      3
 1 |     2      1      0
[2 rows x 3 columns]
```
is this the result you're getting?

pradkrish: I see

```python
  |     G      V      V.0
   | int32  int64  float64
-- + -----  -----  -------
 0 |     1      3       NA
 1 |     2      1        4
[2 rows x 3 columns]

```
and `countna` on the third column also take the `else` path in `head_reduce_unary.cc:1614` That is a clear example where it fails.

docs/api/dt/countna.rst
pradkrish: Ack

docs/api/dt/countna.rst
pradkrish: Ack

docs/api/index-api.rst
pradkrish: Ack

docs/releases/v1.0.0.rst
pradkrish: Ack

src/core/documentation.h
pradkrish: Ack

tests/test-reduce.py
pradkrish: Ack

st-pasha (approved):

PR #3028, opened 2021-06-28
Fix creating a frame from a list of numpy strings

### LINKED ISSUE #3026: Wrong shape when creating Frame from list of `np.str_`s
```
>>> dt.Frame([np.str_('a'), np.str_('b')])
   | C0     C1   
   | str32  str32
-- + -----  -----
 0 | a      b    
[1 row x 2 columns]
```

Expected output:
```
>>> dt.Frame([np.str_('a'), np.str_('b')])
   | C0   
   | str32
-- + -----
 0 | a    
 1 | b    
[2 rows x 1 column]
```

st-pasha:
Closes #3026

### DIFF
--- a/tests/frame/test-create.py
+++ b/tests/frame/test-create.py
@@ -1305,6 +1305,12 @@ def test_create_from_numpy_floats_mixed(numpy):
     assert DT.to_list() == [[4.0, 3.5, None, 9.33]]
 
 
+def test_create_from_numpy_strings(np):
+    src = ["alpha", "beta", "gamma"]
+    DT = dt.Frame([np.str_(s) for s in src])
+    assert_equals(DT, dt.Frame(src))
+
+
 def test_create_from_numpy_reversed(numpy):
     DT = dt.Frame(numpy.array(range(10), dtype=numpy.int64)[::-1])
     assert_equals(DT, dt.Frame(range(9, -1, -1), stype=dt.int64))
--- a/docs/releases/v1.0.0.rst
+++ b/docs/releases/v1.0.0.rst
@@ -91,6 +91,9 @@
     -[fix] Sorting in reverse order now works correctly in the presence
       of a groupby. [#2838]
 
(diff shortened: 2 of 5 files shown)


tests/frame/test-create.py
```diff
@@ -1305,6 +1305,12 @@ def test_create_from_numpy_floats_mixed(numpy):
     assert DT.to_list() == [[4.0, 3.5, None, 9.33]]
 
 
+def test_create_from_numpy_strings(np):
+    src = ["alpha", "beta", "gamma"]
+    DT = dt.Frame([np.str_(s) for s in src])
+    assert_equals(DT, dt.Frame(src))
+
+
```
oleksiyskononenko, line 1313: I would also add a test for `np.str()` even though it seems to work fine atm.

oleksiyskononenko (approved):

oleksiyskononenko:
Do we control which version of numpy we're testing with? It seems that this issue was not there in the older numpy's.

tests/frame/test-create.py
st-pasha: That's way too unlikely. Currently `np.str` is the same as builtin `str`. What could they change it into? Not into a wrapper class for strings, because they already have `np.str_` for that. They could remove it I guess, or convert it into a submodule, and then your proposed test would fail. But there is no need for us to even know that such change will have happened.

st-pasha:
We're testing against the latest versions of numpy and pandas.

tests/frame/test-create.py
oleksiyskononenko: Well, when we write a test we normally don't know in advance what would cause it to fail.

st-pasha: A test is only useful when its failure would signal that we need to fix something in datatable.
Whereas for this test its failure will only cause us to delete the test.

oleksiyskononenko: It depends on why it would fail.

Issue #3029, opened 2021-06-28
Move API documentation into docs/


st-pasha:
After #3002, the existing API documentation should be converted into pure RST format and moved into docs/ folder.

PR #3033, opened 2021-06-29
Add a deprecation warning and adjust docs for `dt.split_into_nhot()`

oleksiyskononenko:
This is in preparation to fully replace `dt.split_into_nhot()` with `dt.str.split_into_nhot()` in `1.1.0`. All the associated documentation is moved from C++ to RST.

WIP for https://github.com/h2oai/datatable/issues/3029.


src/datatable/__init__.py
```diff
@@ -52,13 +52,13 @@
     setdiff,
     shift,
     sort,
+    split_into_nhot_depr as split_into_nhot,
```
oleksiyskononenko, line 55: @st-pasha After your changes, `split_into_nhot_depr` should be changed to `split_into_nhot_deprecated` here.

src/core/str/py_str.cc
```diff
@@ -140,9 +63,32 @@ static oobj split_into_nhot(const PKArgs& args) {
...
+  auto w = DeprecationWarning();
+  w << "Function `dt.split_into_nhot()` is deprecated since version 1.0.0, "
+       "and will be removed in version 1.1.0.\n"
+       "Please use `dt.str.split_into_nhot()` instead.";
+  w.emit_warning();
+  return split_into_nhot(args);
 }
 
+DECLARE_PYFN(&split_into_nhot_deprecated)
+    ->name("split_into_nhot_depr")
```
oleksiyskononenko, line 85: Didn't notice you left the old name here. May be let's use `split_into_nhot_depr` or `split_into_nhot_deprecated` everywhere?

src/datatable/__init__.py
st-pasha: ok

st-pasha (approved):

PR #3042, opened 2021-07-01
Bring Linux and macOS builds to AppVeyor

oleksiyskononenko:
Recent Travis [changes](https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/) made it not usable for datatable anymore. Open-source products are only given a limited amount of credits now, and we ran out of those pretty fast. In this PR we eliminate Travis, instead, bringing Linux builds to AppVeyor. For completeness, we also add macOS builds there. Some other changes:

- Travis config file was removed, the corresponding webhook has been removed too;
- `appveyor.yml` was moved to `ci/`, this required setting up a custom config file name on the AppVeyor website;
- to make it faster, configured AppVeyor to only build and test with the most recent Python 3.9, for Windows we also include Python 3.8. However, in the release mode on Windows it will build/test with all the supported Python versions.


st-pasha:
Why forward slashes? Doesn't windows prefer backslashes in a path?

oleksiyskononenko:
I want to test an idea to use the same PowerShell code for both Linux and Windows. In theory it should work, but in practice I couldn't activate a correct Python environment on Linux from PowerShell. Will see how it goes.

ci/appveyor.yml
```diff
@@ -0,0 +1,251 @@
...
+
+      $DT_WHEEL = ls dist/*-cp36-*.whl
+
+      echo "----- _build_info.py for Python 3.6 ------------------------------"
+
+      cat src/datatable/_build_info.py
+
+      echo "------------------------------------------------------------------"
+
+      pip --disable-pip-version-check install $DT_WHEEL pytest docutils pandas
```
st-pasha, line 181: Let's install latest version of pip instead

oleksiyskononenko: I'm trying that now to see how long it takes

ci/appveyor.yml
```diff
@@ -95,7 +95,9 @@ build_script:
 
     echo "------------------------------------------------------------------"
 
-    pip --disable-pip-version-check install $DT_WHEEL pytest docutils pandas
+    python -m pip install –upgrade pip
```
st-pasha, line 98: ```python -m pip install pip --upgrade```

oleksiyskononenko: What is the difference with https://pip.pypa.io/en/stable/installing/#upgrading-pip ?

oleksiyskononenko: hm, somehow `--` got converted to `-`, thanks will fix

st-pasha (approved):

ci/appveyor.yml
```diff
@@ -0,0 +1,261 @@
...
+
+    echo "----- _build_info.py for Python 3.9 ------------------------------"
+
+    cat src/datatable/_build_info.py
+
+    echo "------------------------------------------------------------------"
+
+    python -m pip install --upgrade pip
+
+    pip install $DT_WHEEL pytest docutils pandas
```
st-pasha, line 259: In theory, `pip` and `python -m pip` could invoke different versions of pip package.
While it's unlikely to be happening here, I'm not sure what's the reason of using 2 different call conventions.

oleksiyskononenko: Just happened with no particular reason.

PR #3049, opened 2021-07-06
fix for expected behavior of min max operations

### LINKED ISSUE #2758: [BUG]Max() and Min() operate differently to Mean()
- Did you find a bug in datatable, or maybe the bug found you?
  The `mean()` function allows an iterable(list, tuple, dict)  of single and multiple columns and returns the average of each column; min() and max() does not replicate this. I had a look at the issues, and did not find any reference to this; it may very well be a design choice; it just seems odd and a glaring difference.

- How to reproduce the bug?
```python
df = dt.Frame({"A": [1, 1, 2, 1, 2],
               "B": [None, 2, 3, 4, 5],
               "C": [1, 2, 1, 1, 2]})

  | A | B | C
0 | 1 | NA| 1
1 | 1 | 2 | 2
2 | 2 | 3 | 1
3 | 1 | 4 | 1
4 | 2 | 5 | 2

# mean can accept a list of columns:
df[:, dt.mean([f.A, f.B])]

      A | B
0 | 1.4 | 3.5

# it can take a tuple
df[:, dt.mean((f.A, f.B))]
      A | B
0 | 1.4 | 3.5

# it works well with a dictionary
df[:, dt.mean({"A_mean": f.A, "B_mean": f.B})]
  A_mean | B_mean
0 |  1.4 | 3.5

# max() and min() however, cannot replicate this. 
# you have to individually pass the aggregation function:

df[:, [dt.max(f.A), dt.max(f.B)]]
	A	B
0	2	5

# if you attempt to pass the max to all the columns in a list, it fails:
df[:, dt.max([f.A, f.B])]

TypeError                                 Traceback (most recent call last)
<ipython-input-253-ac650448862a> in <module>
----> 1 df[:, dt.max([f.A, f.B])]

~/miniforge3/envs/datatable/lib/python3.9/site-packages/datatable/expr/reduce.py in max(*args, **kwds)
    123         return args[0].max()
    124     else:
--> 125         return _builtin_max(*args, **kwds)
    126 
    127 

TypeError: Expression f.B > f.A cannot be cast to bool.

You may be seeing this error because either:
  * you tried to use chained inequality such as
        0 < f.A < 100
    If so please rewrite it as
        (0 < f.A) & (f.A < 100)

  * you used keywords and/or, for example
        f.A < 0 or f.B >= 1
    If so then replace keywords with operators & or |:
        (f.A < 0) | (f.B >= 1)
    Be mindful that & / | have higher precedence than and
    or or, so make sure to use parentheses appropriately.

  * you used expression in the if statement, for example:
        f.A if f.A > 0 else -f.A
    You may write this as a ternary operator instead:
        (f.A > 0) & f.A | -f.A

  * you explicitly cast the expression into bool:
        bool(f.B)
    this can be replaced with an explicit comparison operator:
        f.B != 0

# with a dictionary, it just returns the column, with no aggregation:
df[:, dt.min({"A": f.A})]
	A
0	1
1	1
2	2
3	1
4	2

# same for a list with a single column: 
df[:, dt.max([f.A])]
	A
0	1
1	1
2	2
3	1
4	2

# whereas mean() would reduce : 
df[:, dt.mean({"A": f.A})]
          A
0	1.4

# As a side note, it seems max() and min() can accept a key, 
# similar to python's built -ins; however, I am not sure of what possible uses there are for this:
df[:, dt.min(f.A, key=lambda d: 2)]
        A
0	1
 ```

  Personally, I feel it wont be a bad idea if `max()` and `min()` act the same way mean() does. I believe it is a convenient option to pass an iterable of columns to the aggregation functions, and could possibly allow for more complex operations, where you write your code out in stages and at the end pass it to the aggregation. 

- Your environment?
datatable version : '1.0.0a0+build.1604537228.sam'
python version :  '3.9.0 | packaged by conda-forge | (default, Oct 14 2020, 22:59:50) \n[GCC 7.5.0]'
operating system: linux

### PR DESCRIPTION
**Pull request by Mathanraj-Sharma**
Now` min() `and `max()` functions works similar to `mean()`

1. fixed bug in accepting list of columns 
2. breaked `minmax` function into `min` and `max ` to accept dict with new col names, similar to mean function. 

```
>>> from datatable import f, by
>>> import datatable as dt

>>> df = dt.Frame({'A': [1, 1, 1, 2, 2, 2, 3, 3, 3],
... 'B': [3, 2, 20, 1, 6, 2, 3, 22, 1]})


>>> df[:, dt.max(f.B)]
   |     B
   | int32
-- + -----
 0 |    22
[1 row x 1 column]


>>> df[:, [dt.max(f.A), dt.max(f.B)]]
   |     A      B
   | int32  int32
-- + -----  -----
 0 |     3     22
[1 row x 2 columns]


>>> df[:, dt.max(f[:])]
   |     A      B
   | int32  int32
-- + -----  -----
 0 |     3     22
[1 row x 2 columns]


>>> df[:, dt.max(f.B), by("A")]
   |     A      B
   | int32  int32
-- + -----  -----
 0 |     1     20
 1 |     2      6
 2 |     3     22
[3 rows x 2 columns]


>>> df[:, dt.max([f.A, f.B])]
   |     A      B
   | int32  int32
-- + -----  -----
 0 |     3     22
[1 row x 2 columns]


>>> df[:, dt.max({"A_max": f.A, "B_max": f.B})]
   | A_max  B_max
   | int32  int32
-- + -----  -----
 0 |     3     22
[1 row x 2 columns]
>>> 

```

Closes #2758

### CODE REVIEW

**Comment by samukweku** [comment]
@Mathanraj-Sharma thanks for taking this on. Kindly add tests to your code. And maybe add to the docs examples showing the new functionality. Any plans to add the dictionary options?

**Comment by Mathanraj-Sharma** [comment]
> @Mathanraj-Sharma thanks for taking this on. Kindly add tests to your code. And maybe add to the docs examples showing the new functionality. Any plans to add the dictionary options?

@samukweku I made some code changes to accept dict with new column names, similar to mean. 


@st-pasha Do I need to add separate tests to these changes?

File: src/core/expr/head_reduce_unary.cc [hunk 1]
```diff
@@ -1188,14 +1197,58 @@ per group::
      1 |     2      1
      2 |     3      1
     [3 rows x 2 columns]
+)";
+#endif
 
+template <typename T, typename U>
```
st-pasha, line 1203: why do you need `U` here?
Seems like you always have `U == T`.

src/datatable/expr/reduce.py
```diff
@@ -127,4 +131,4 @@ def max(*args, **kwds):
 
 sum.__doc__ = _builtin_sum.__doc__
 min.__doc__ = _builtin_min.__doc__
-max.__doc__ = _builtin_max.__doc__
+max.__doc__ = _builtin_max.__doc__
```
st-pasha, line 134: Newline is missing here

st-pasha (commented):
Please add unit tests to verify that the functions `min()` and `max()` work as expected.

src/core/expr/head_reduce_unary.cc
Mathanraj-Sharma: removed unwanted U.

Issue #1692, opened 2019-02-28
Implement Array column type


st-pasha:
Array column is a column where each element is an array of items.
Obviously, there could be different array stypes depending on (a) whether the offsets are int32 or int64, and (b) the type of the elements in each array.

Currently it is unclear whether the content elements can be of any supported type, or only a predefined subset. If the content can be anything, then there probably will be only 2 stypes for the array: ARR32 / ARR64, and the type of the content will be retrieved by an additional property. On the other hand, we currently only really need arrays of type `Arr32[Str32]` and `Arr64[Str64]`.

Arrays can be implemented in a way similar to strings: as list of offsets into a column of some other type. For str32/str64 stypes this "other column" is really just a `char*` string data section; for arrays it will be a more generic `T*` content. Otherwise, arrays are not too different from strings.

The reason we need arrays is so that we can support string split function and regular expression matches. The operators we can support for arrays are: 
  - [ ] extract an item from an array
  - [ ] concatenate arrays
  - [ ] slice an array
  - [ ] append something to the array
  - [ ] sort each array
  - [ ] string-join values in the array
  - [ ] use array in a format string (e.g. "$0-$1-$2 $3:$4" to format a 5-element array as a time string)
  - [ ] make array out of several columns
  - [ ] expand array into columns
  - [ ] "melt" array into rows (essentially, retrieve the "content" column)

PR #3050, opened 2021-07-07
New types list32 and list64

st-pasha:
This PR adds 2 new type objects: `list32(T)` and `list64(T)`. They are not yet used anywhere, but we have to have a `Type` before a column can be implemented.

WIP for #1692

### DIFF
--- a/docs/api/type/list32.rst
+++ b/docs/api/type/list32.rst
@@ -1,0 +1,13 @@
+
+.. xmethod:: datatable.Type.list32
+    :src: src/core/types/py_type.cc PyType::list32
+    :cvar: doc_Type_list32
+    :signature: list32(T)
+
+    List type with 32-bit offsets. In a column of this type, every element
+    is a list of types `T`.
+
+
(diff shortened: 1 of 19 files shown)


docs/api/type/list32.rst
```diff
@@ -0,0 +1,13 @@
+
+.. xmethod:: datatable.Type.list32
+    :src: src/core/types/py_type.cc PyType::list32
+    :cvar: doc_Type_list32
+    :signature: list32(T)
+
+    List type with 32-bit offsets. In a column of this type, every element
+    is a list of types `T`.
```
oleksiyskononenko, line 8: I would clarify what offsets mean in this context.

oleksiyskononenko (approved):

docs/api/type/list32.rst
st-pasha: Oh sure, we'd need to expand the documentation more when there's a better understanding of how this type works.

PR #3066, opened 2021-07-15
rbind(force=True) will no longer produce a TypeError

### LINKED ISSUE #3062: rbind(force=True) should perform more aggressive type casts
In a situation like this there shouldn't be an error:
```
>>> dt.rbind(dt.Frame([1]), dt.Frame(["1"]), force=True)
TypeError: Cannot rbind column of type str32 to a column of type int8
```
Instead, we want to promote the integer column to string.

st-pasha:
Rbind-ing columns of disparate types is now possible with parameter `force=True`.

Closes #3062

### DIFF
--- a/docs/releases/v1.1.0.rst
+++ b/docs/releases/v1.1.0.rst
@@ -16,5 +16,5 @@
     -------
 
     -[imp] Parameter `force=True` in function :func:`rbind()` (or method
-        :meth:`dt.Frame.rbind()`) will now allow also allow combining of
-        columns of mismatched types. [#3062]
+        :meth:`dt.Frame.rbind()`) will now allow combining columns
+        of incompatible types. [#3062]
--- a/docs/releases/v1.1.0.rst
+++ b/docs/releases/v1.1.0.rst
@@ -14,3 +14,7 @@
 
     General
     -------
+
+    -[imp] Parameter `force=True` in function :func:`rbind()` (or method
+        :meth:`dt.Frame.rbind()`) will now allow also allow combining of
+        columns of mismatched types. [#3062]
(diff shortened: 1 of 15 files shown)


docs/releases/v1.1.0.rst
```diff
@@ -14,3 +14,7 @@
 
     General
     -------
+
+    -[imp] Parameter `force=True` in function :func:`rbind()` (or method
+        :meth:`dt.Frame.rbind()`) will now allow also allow combining of
```
oleksiyskononenko, line 19: A typo here: `will now allow also allow`.

tests/test-options.py
```diff
@@ -316,13 +316,13 @@ def test_debug_logger_object():
...
         DT = dt.rbind([])
-        assert "dt.rbind([]) {" in logger.msg
+        assert "datatable.rbind([]) {" in logger.msg
         assert re.search(r"} # \d+(?:\.\d+)?(?:[eE][+-]?\d+)? s", logger.msg)
         logger.msg = ""
 
         with pytest.raises(TypeError):
             dt.rbind(4)
-        assert "dt.rbind(4) {" in logger.msg
+        assert "datatable.rbind(4) {" in logger.msg
```
oleksiyskononenko, line 325: `dt` won't work for some reason?

oleksiyskononenko (approved):

tests/test-options.py
st-pasha: This is just a message that gets printed in the console. Previously the message said "dt.rbind" and now it's "datatable.rbind" because of the switch to XArgs.

docs/releases/v1.1.0.rst
st-pasha: fixed

PR #3084, opened 2021-07-19
Move `LinearModel` documentation to `/docs` folder

oleksiyskononenko:
WIP for #3029


st-pasha:
The webhook for readthedocs worked perfectly.

st-pasha (approved):

oleksiyskononenko:
Yes, but if we don't fail on errors then won't it pass all the time?

PR #3086, opened 2021-07-19
Added config file for readthedocs

st-pasha:
Created a config file for readthedocs, because this is the only way to turn on the option "fail the build in case of warnings/errors"


st-pasha:
Test results for failure in the following scenarios:
- [ ] wrong attribute in `:src:` option of `..xfunction` directive;
- [x] invalid rst markup;
- [ ] wrong target of a reference;

oleksiyskononenko (approved):

PR #3089, opened 2021-07-20
Sort void column

### LINKED ISSUE #3088: Cannot sort void column
```
>>> dt.Frame([np.nan, np.nan])[:, dt.count(), dt.by(0)]
NotImplementedError: Unable to sort Column of stype void
```

st-pasha:
- Allow void columns to be used in `sort()`
- Sorting will be skipped for any constant column (i.e. having `Const_ColumnImpl`).

Closes #3088

### DIFF
--- a/src/core/sort.cc
+++ b/src/core/sort.cc
@@ -1433,9 +1433,11 @@ RiGb group(const std::vector<Column>& columns,
   const Column& col0 = columns[0];
 
   size_t nrows = col0.nrows();
-  #if DT_DEBUG
-    for (const Column& col : columns) xassert(col.nrows() == nrows);
-  #endif
+  size_t n_const_cols = 0;
+  for (const Column& col : columns) {
+    xassert(col.nrows() == nrows);
+    n_const_cols += col.is_constant();
+  }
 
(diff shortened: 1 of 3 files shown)


oleksiyskononenko (approved):

src/core/sort.cc
```diff
@@ -1433,19 +1433,21 @@ RiGb group(const std::vector<Column>& columns,
...
   // For a 0-row Frame we return a rowindex of size 0, and the
   // Groupby containing zero groups.
   if (nrows == 0) {
     result.second = Groupby::zero_groups();
     return result;
   }
-  if (nrows == 1) {
-    Buffer buf = Buffer::mem(sizeof(int32_t));
-    buf.set_element<int32_t>(0, 0);
+  if (nrows == 1 || n_const_cols == n) {
```
oleksiyskononenko, line 1448: So this will only skip sorting only when all the columns are constant, right? If we have two columns to be sorted, and only one of them is constant — we will still sort both columns?

st-pasha: yes. It can be improved, as you have noted, but will take more work to account for sortflags etc.

PR #3106, opened 2021-07-27
`.is_nnn` properties in Type class

### LINKED ISSUE #3101: Add query methods `.is_integer()`, `.is_numeric()`, etc to Type class
These methods already exist in the internal `Type` class, now we only need them to be visible outside.

st-pasha:
Added 10 properties into the `dt.Type` class
- `.is_array`
- `.is_boolean`
- `.is_compound`
- `.is_float`
- `.is_integer`
- `.is_numeric`
- `.is_object`
- `.is_string`
- `.is_temporal`
- `.is_void`

Closes #3101

### DIFF
--- a/docs/api/type.rst
+++ b/docs/api/type.rst
@@ -44,6 +44,36 @@
         :widths: auto
         :class: api-table
 
+        * - :attr:`.is_array`
+          - Is this an array type?
+
+        * - :attr:`.is_boolean`
+          - Is this a boolean type?
+
+        * - :attr:`.is_compound`
+          - Is this a compound type?
+
+        * - :attr:`.is_float`
+          - Is this a float type?
+
+        * - :attr:`.is_integer`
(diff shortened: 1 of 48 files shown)


oleksiyskononenko (commented):
In the issue https://github.com/h2oai/datatable/issues/3101 you wanted to make them methods, why you decided to go for properties? Also, we never test that these properties are read-only.

oleksiyskononenko (approved):

st-pasha:
[Google style doc](https://google.github.io/styleguide/pyguide.html#2134-decision) recommends using getters when the underlying implementation is sufficiently lightweight. Also, in [this name guide](https://dart.dev/guides/language/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable), they also recommend using simple properties for variables that describe what an object *is*.

PR #3131, opened 2021-08-03
Create arr32 column from a python list

### LINKED ISSUE #3054: Convert python list of `list` objects into an `arr32` column

st-pasha:
`arr32` column can now be created from a python list. The elements in the python list must all have compatible types.

In addition, `arr32` column will now work properly with many other virtual columns, such as `Repeated`, `Truncated`, `RBound`, `SliceView`, etc.

Closes #3054


tests/types/test-array.py
```diff
@@ -114,3 +114,79 @@ def test_create_from_arrow(pa):
...
+    assert DT.type == a(a(a(a(a(dt.Type.void)))))
+    assert DT.to_list() == [src]
+
+
+def test_create_from_python_array_incompatible_child_types():
+    src = [[1, 2, 3], ["one"]]
+    msg = "Cannot create column: element at index 3 is of type " \
+          "<class 'str'>, whereas previous elements were int32"
+    with pytest.raises(TypeError, match=msg):
+        DT = dt.Frame(E=src)
```
oleksiyskononenko, line 192: What about adding a test that will modify the original list and then will make sure the frame's content didn't change?

oleksiyskononenko (approved):

tests/types/test-array.py
st-pasha: Since there is not even a theoretical way for the python list of objects to affect the content of a datatable's Column, I don't think there is a need for such a test.

PR #3137, opened 2021-08-10
Arr32 column can now be displayed in a terminal

### LINKED ISSUE #3052: Display `arr32` column in a python console;

st-pasha:
Closes #3052

### DIFF
--- a/tests/types/test-array.py
+++ b/tests/types/test-array.py
@@ -235,3 +235,16 @@ def test_arr32_arr32_repr():
         " 3 | [[-1], [0, 13]]    \n"
         "[4 rows x 1 column]\n"
     )
+
+
+def test_arr32_of_strings_repr():
+    DT = dt.Frame(W=[['ad', 'dfkvjn'], ['b b, f', None], ['r', 'w', 'dfvdf']])
+    assert str(DT) == (
+        "   | W            \n"
+        "   | arr32(str32) \n"
+        "-- + -------------\n"
+        " 0 | [ad, dfkvjn] \n"
+        " 1 | [b b, f, NA] \n"
+        " 2 | [r, w, dfvdf]\n"
+        "[3 rows x 1 column]\n"
+    )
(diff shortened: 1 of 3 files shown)


tests/types/test-array.py
```diff
@@ -190,3 +190,61 @@ def test_create_from_python_array_incompatible_child_types():
...
+    )
+
+
+def test_arr32_of_strings_repr():
+    DT = dt.Frame(W=[['ad', 'dfkvjn'], ['b b, f', None], ['r', 'w', 'dfvdf']])
+    assert str(DT) == (
+        "   | W            \n"
+        "   | arr32(str32) \n"
+        "-- + -------------\n"
+        " 0 | [ad, dfkvjn] \n"
```
oleksiyskononenko, line 246: I understand that when we print content of a string column we don't add quotes, because we are kind of printing the raw data. If one needs to copy and paste a string somewhere into the python code, it is not a big deal to add quotes.

However, in the case of printing `arr*` it could be a little bit confusing, because we stick to the python's list syntax and a user may think that copy-pasting a value into python may work. But, quotes should be added around each element and that could be boring.

oleksiyskononenko (approved):

tests/types/test-array.py
st-pasha: The discrepancy is more than just quotation marks. NA elements in the list will get rendered as `NA`, not `None` (which I think is good), the list itself will get shortened with an ellipsis `...` if there are too many elements, the individual elements in the list will get truncated too if they are too long -- see the test below for such a use case.

Obviously, the fact that we don't have quotes for strings here is somewhat problematic. For example, a string value may contain embedded `,` or square brackets, or even spaces -- which will make the output somewhat confusing. However, strings are not the only use-case. What if it's an array of dates? Right now it'll look as `[2000-01-01, 2001-02-22]`, but that is certainly not a proper python syntax either.

I'm not sure what the best approach should be here. Special handling for strings only? However, it won't be as simple as surrounding each element with quotation marks -- you also need to escape the quotes inside the string.

In any case, this seems like a very minor issue that can be postponed to some future time (or filed as "Beginner task" for the potential newcomers).

oleksiyskononenko: May be we could add a Frame method called `.to_python()` or something that will return a valid python list of lists.

st-pasha: We already have one: `.to_list()`. Alternatively, `.to_dict()` or `.to_tuples()`.

PR #3143, opened 2021-08-13
Fix a crash when `.ltypes` or `.stypes` were accessed for frames with array columns

### LINKED ISSUE #3142: Error when accessing `.ltypes` or `.stypes` of a Frame that contains array columns
Accessing `.ltypes` triggers segfaut:
```python
from datatable import dt

DT = dt.Frame(A = [[1]])
DT.ltypes #segfault
```

Accessing `.stypes` triggers `AssertionError: Assertion '!value.is_undefined()' failed in src/core/python/tuple.cc, line 105`:
```python
from datatable import dt

DT = dt.Frame(A = [[1]])
DT.stypes #AssertionError
```

oleksiyskononenko:
Handling of array columns was fixed by adding `ltype.invalid` and `stype.arr32`/`stype.arr64`. Also, several relevant tests were adjusted.

Closes #3142

### DIFF
--- a/docs/releases/v1.1.0.rst
+++ b/docs/releases/v1.1.0.rst
@@ -25,6 +25,9 @@
     -[fix] Fixed a crash which could have occurred when sorting very long
         identical or nearly identical strings. [#3134]
 
+    -[fix] Fixed a crash when :attr:`.ltypes` or :attr:`.stypes` properties
+        were accessed for a frame containing array columns. [#3142]
+
     -[api] Converting a column of :attr:`void <dt.Type.void>` type into pandas
         now produces a pandas ``object`` column filled with ``None``s. Converting
         such column back into datatable produces a ``void`` column again. [#3063]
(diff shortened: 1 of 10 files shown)


st-pasha (approved):

docs/releases/v1.1.0.rst
```diff
@@ -25,6 +25,9 @@
     -[fix] Fixed a crash which could have occurred when sorting very long
         identical or nearly identical strings. [#3134]
 
+    -[fix] Fixed a crash when :attr:`.ltypes` or :attr:`.stypes` properties
+        were accessed for a frame containing array columns. [#3142]
```
st-pasha, line 29: This is not fixing something from the previous  version, but rather something that we just added recently. As such, it's not needed to be mentioned in the release notes.

oleksiyskononenko: Yeah, let's remove it.

PR #3155, opened 2021-08-31
Allow other columns to be cast into array type

st-pasha:
- In `Arrow_ColumnImpl` methods `num_buffers()` and `get_buffer()` renamed into `get_num_data_buffers()` and `get_data_buffer()` respectively, which are defined in the base `ColumnImpl` class;
- Added type casts of `void -> arr32<T>`, `obj -> arr32<T>` and `arr32<S> -> arr32<T>`;
- In `DT[...] = type` assignment, the RHS can now be a compound type (before, only types convertible to an stype were supported);
- Common functionality for `arr32` and `arr64` types moved into the `Type_Array` class;
- Added function `TypeImpl::make_type()` for creating a `Type` object out of an existing `TypeImpl`;

### DIFF
--- a/src/core/column.cc
+++ b/src/core/column.cc
@@ -40,6 +40,11 @@ Column Column::new_data_column(size_t nrows, dt::SType stype) {
 }
 
 
+Column Column::new_na_column(size_t nrows, dt::Type type) {
+  return Column(new dt::ConstNa_ColumnImpl(nrows, type));
+}
+
+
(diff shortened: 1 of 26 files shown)


oleksiyskononenko (approved):

st-pasha:
I'm going to merge then -- the only test failure is because one jenkins machine ran out of disk space.

Issue #1691, opened 2019-02-27
Implement categorical columns


st-pasha:
A categorical column is semantically equivalent to a string column, except that it uses integer codes to store the values. The layout of such column is therefore:
```
T values[n];  // array of indices into a dictionary
StringColumn<int32> dict;  // "dictionary" column
```
(where `T` could be `int8`, `int16` or `int32`).

XiaomoWu:
Just a quick question: the  `Rdatatable` avoids categorical column (in R it's called `factor`) partially because it slows downs the performance. Just wonder will the performance of `pydatatable` be affected if introducing categorical column.

st-pasha:
Hmm, interesting, I have not heard about that. Perhaps, there are specific scenarios where the factor variables become slower? Like, when the number of factors approaches the number of rows?

There are clearly situations where factors would be preferable. For example, if there are only few of them: this would speed up sorting for example, and also have the potential to greatly reduce the required storage space.

Anyways, the categorical type should be in addition to, not as a replacement for the regular string type. So the user will be able to use whatever format better suits his/her need.

jangorecki:
@XiaomoWu the reason to avoid factor was not speed but problems with its levels when combining, filtering, or performing string operations like `paste`. Factors are faster than character, and can be processed in parallel, while character's R global cache is not thread safe.

jangorecki:
When implemented it might allow to read 1e9 data sets into db-benchmark for pandas, currently `to_pandas()` fails with OOM (afair). Having categoricals instead of objects could significantly reduce memory footprint.
Recent attempt to optimise pandas `read_csv` has failed, see https://github.com/h2oai/db-benchmark/issues/99

PR #3158, opened 2021-09-04
Implement `CategoricalColumn_Impl` and support for basic operations

oleksiyskononenko:
- implemented `CategoricalColumn_Impl`;
- added support for categorical columns in `dt.Frame()`;
- added support for categorical columns in a terminal, also allowing the element access through `[i, j]` selector;
- added and modified some relevant tests.

WIP for #1691

### DIFF
--- a/src/core/column/categorical.h
+++ b/src/core/column/categorical.h
@@ -55,6 +55,7 @@ class Categorical_ColumnImpl : public Virtual_ColumnImpl {
     bool get_element(size_t, py::oobj*) const override;
     bool get_element(size_t, Column*)   const override;
 
+    void materialize(Column&, bool) override;
 };
 
 
--- a/src/core/column/categorical.h
+++ b/src/core/column/categorical.h
@@ -22,7 +22,6 @@
 #ifndef dt_COLUMN_CATEGORICAL_h
 #define dt_COLUMN_CATEGORICAL_h
 
-#include "column/categorical.h"
 #include "column/virtual.h"
 
 
--- a/src/core/column/categorical.h
+++ b/src/core/column/categorical.h
@@ -1,0 +1,68 @@
+//------------------------------------------------------------------------------
+// Copyright 2021 H2O.ai
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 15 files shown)


src/core/column/categorical.h
```diff
@@ -0,0 +1,70 @@
...
+#include "column/virtual.h"
+
+
+namespace dt {
+
+template <typename T>
+class Categorical_ColumnImpl : public Virtual_ColumnImpl {
+  private:
+    Buffer codes_;
+    Column categories_;
```
st-pasha, line 34: There is a question of how to encode the NA values.
With your current approach, you only support NAs when they appear within the categories column. This could be too limited when one wants to set up categories to be a specific array. Also, this would make it harder to interoperate with the Arrow format.

The arrow's approach here is to have a separate `validity` buffer that uses a bitmask to encode whether each value is valid or not.

oleksiyskononenko: >This could be too limited when one wants to set up categories to be a specific array. 

What do you mean? Let's say there is data `["cat", "dog", "mouse"]` and an array of categories `["cat", "dog"]`. Do you propose to encode `"mouse"` with `N/A`? I guess there could be a flag to enforce it, but by default the code should throw an exception saying that there is an unknown category.

st-pasha: What i mean is suppose the categories are `["cat", "dog", "mouse"]`. Now, new data comes in, and it's, e.g., an `"aardvark"`. We don't have that among the categories, so this value has to be encoded as `NA`. But NA is also not among the categories, so what do you do now?

A more robust approach would be to allow the column of `codes` to be nullable, by adding an extra validity buffer.

oleksiyskononenko: >so what do you do now?

Well, there are several options: either throw an error by saying there is an unknown category, or adding `"aardvark"` as a new category, or adding `N/A` as a new category and encode `"aardvark"` as `N/A`. I agree that having a separate buffer will simplify interconnection with arrow. We can always add another buffer with the N/A bitmask, but if instead of a buffer we can just add one element in the category column — why we shouldn't?

oleksiyskononenko: In principle, if we establish convention that `N/A` (unknown categories) are always encoded as `0` — then, to create an N/A bitmask it would be enough just to cast the buffer of codes into `bool`. To go back, from the bitmask to the correct buffer or codes, it would be enough just to apply the mask to each element in the codes array.

oleksiyskononenko:
I have added a task in https://github.com/h2oai/datatable/issues/1691 to think through and implement effective `N/A` handling for categoricals. Merging this PR as it seems that there are no other issues here.

src/core/column/categorical.h
oleksiyskononenko: btw, arrow actually allows an `N/A` category

```
>>> pa.DictionaryArray.from_arrays([0, 1, 2, 0, 1], [None, "a", "b"])
<pyarrow.lib.DictionaryArray object at 0x7fc37ee31828>

-- dictionary:
  [
    null,
    "a",
    "b"
  ]
-- indices:
  [
    0,
    1,
    2,
    0,
    1
  ]
```

Issue #3178, opened 2021-10-06
[bug] sys not imported in fread


lucasjamar:
When I do fread on an excel, I am getting the followign error message:
```
  Traceback (most recent call last):
  File "C:\Users\JAMAR_L\Anaconda3\envs\myenv\Lib\site-packages\datatable\utils\fread.py", line 156, in _resolve_source_file
    if sys.platform == "win32":
NameError: name 'sys' is not defined
```

- Your environment?
Windows 10, Python 3.8.6, datatable 1.0.0

PR #3179, opened 2021-10-06
added import sys to fread.py

lucasjamar:
in response to https://github.com/h2oai/datatable/issues/3178

### DIFF
--- a/src/datatable/utils/fread.py
+++ b/src/datatable/utils/fread.py
@@ -26,6 +26,7 @@ import os
 import pathlib
 import re
 import shutil
+import sys
 import tempfile
 import warnings
 


oleksiyskononenko:
Thanks, @lucasjamar. 

It seems that the reason such an obvious bug was never caught by the tests is the Windows file descriptor check being placed in the try-catch block. So the `NameError: name 'sys' is not defined` exception was caught and silenced.

As a part of your PR, could you please move that Windows check out? Something like this (instead of https://github.com/h2oai/datatable/blob/main/src/datatable/utils/fread.py#L155-L157) will work
```python
        if sys.platform == "win32":
            raise Exception("Do not use file descriptors on Windows")
        try:
```

Then, if the `import sys` gets somehow removed we will be notified by the tests.

Issue #2632, opened 2020-09-16
[BUG]Fread options not extended to excel files


samukweku:
- Did you find a bug in datatable, or maybe the bug found you?
I cant select columns or limit number of rows read when working on an excel file

- How to reproduce the bug?
```python
import urllib

url = 'https://github.com/samukweku/Extracting-Data-from-Excel-with-Python/blob/master/names.xlsx?raw=true'

filename = 'test.xlsx'
        
urllib.request.urlretrieve(url, filename)

fread("test.xlsx/naija")
      Name	      Age	Height
0	Tolu	        24       2
1	Chukwuka 	50	1.8
2	Ogor	        15	1.5

# column selection fails
fread('test.xlsx/naija', columns={"Name"})

# nrows fails as well
fread("test.xlsx/naija", max_nrows=2)
```

- What was the expected behavior?

```python
  # column selection : 
fread('test.xlsx/naija', columns={"Name"})
      Name
0	Tolu
1	Chukwuka
2	Ogor

# row selection : 
fread("test.xlsx/naija", max_nrows=2)
      Name	     Age	Height
0	Tolu	     24       	2
1	Chukwuka     50      	1.8

```
- Your environment?
```python
dt.__version__
'0.11.0a0+build.1600188275.sam'

sys.version
'3.8.5 | packaged by conda-forge | (default, Aug 29 2020, 01:22:49) \n[GCC 7.5.0]'

operating system
'Linux-5.4.0-7642-generic-x86_64-with-glibc2.10'
```

pradkrish:
@samukweku I can confirm that I am able to reproduce your observation for xlsx files.

pradkrish:
Just briefly looked into the code. I don't think it's even implemented for xls files, perhaps @st-pasha could comment on this??

st-pasha:
That's correct. The logic for reading Excel files is completely separate from the code that reads CSV files, and consequently every option has to be implemented for Excel parsing separately.

pradkrish:
That's what I gathered looking at the code. is there an appetite to do so for Excel files? My guess is csv files are probably much more common than Excel files in data manipulation tasks making this a low priority.

Jiezuel:
I had the same problem when I tried to read `.xlsx` file.
I think the reason is probably that `fread` uses xlrd to read `.xlsx` or `.xls` file. See `fread.py in line 325` and 'datatable/xls.py'.
[...]

PR #3191, opened 2021-10-28
Upgrade to xlrd 2.0.0 + openpyxl

carrascomj:
Related to #2632 

### Description
[xlrd](https://pypi.org/project/xlrd/) dropped support for anything but `xls` at [2.0.0](https://github.com/python-excel/xlrd/blob/master/CHANGELOG.rst#200-11-december-2020) https://github.com/python-excel/xlrd/pull/371. Using the old version may cause security vulnerabilities  and potential incorrect parsing. It is also a problem for people that have installed pandas in their environment (with, more likely, openpyxl and xlrd>2.0).

### Implementation
* Use [openpyxl](https://bitbucket.org/openpyxl/openpyxl) for `xlsx`, the recommended alternative.
* openpyxl was added as extra requirements and xlrd was upgraded to >=2.0.0.

I would like to know if this is desirable before writing any unitests.

Thanks!

### DIFF
--- a/src/datatable/xlsx.py
+++ b/src/datatable/xlsx.py
@@ -71,37 +71,22 @@ def read_xlsx_worksheet(ws, subrange=None):
         subview = ws[range2d]
         # the subview is a tuple of rows, which are tuples of columns
         ncols = len(subview[0])
-        nrows = len(subview) - 1
         colnames = [str(n.value) for n in subview[0]]
-        coltypes = [None] * ncols
-        outdata = [[None] * nrows for _ in range(ncols)]
-        for irow, row in enumerate(subview[1:]):
-            for icol, cell in enumerate(row):
-                cell_value = cell.value
-                cell_type = cell.data_type
-                if not cell_type or cell_type == "e" or cell_value is None:
-                    cell_value = None
-                    cell_type = None
-                elif cell_type == "b":
-                    cell_value = bool(cell_value)
-                    cell_type = bool
-                else:
-                    # float, int, date, etc
-                    cell_type = type(cell.value)
 
-                outdata[icol][irow] = cell_value
-                prev_type = coltypes[icol]
-                if prev_type is None and cell_type is not None:
-                    coltypes[icol] = cell_type
-                elif cell_type is None or cell_type == prev_type:
-                    pass
-                elif cell_type is str:
-                    # str always win
-                    coltypes[icol] = str
+        outdata = [
+            [
+                row[i]._value if row[i].data_type != "e" else None
+                for row in subview[1:]
+            ]
+            for i in range(ncols)
+        ]
+        coltypes = [
+            str if any(isinstance(cell, str) for cell in col) else None
+            for col in outdata
+        ]
 
         # let the frame decide on the types of non-str cols
-        types = [str if coltype == str else None
-                 for coltype in coltypes]
+        types = [str if coltype == str else None for coltype in coltypes]
         frame = dt.Frame(outdata, names=colnames, types=types)
         results[range2d] = frame
     return results


carrascomj:
I implemented some tests in the same fashion as the ones for jay. For them to run on CI, openpyxl has to be installed during the appveyor build ([here](https://github.com/h2oai/datatable/blob/main/ci/appveyor.yml#L102) and in the subsequent pip-installs, I guess).

oleksiyskononenko:
@carrascomj thanks for your contribution! Nice to know we can't use `xlrd` for `.xlsx` anymore. Let me review it and also see why some tests are failing on jenkins.

oleksiyskononenko:
@carrascomj do you have any ideas if the issue mentioned [here](https://github.com/h2oai/datatable/issues/2823#issuecomment-755781337) has ever been fixed? Have you had a chance to test `openpyxl` on some larger files?

src/datatable/xlsx.py
```diff
@@ -0,0 +1,107 @@
...
+        import openpyxl
+    except ImportError:
+        raise dt.exceptions.ImportError(
+            "Module `openpyxl` is required in order to read Excel file '%s'"
+            % filename)
+
+    if subpath:
+        wb = openpyxl.load_workbook(filename, data_only=True)
+        range2d = None
+        if subpath in wb.get_sheet_names():
```
oleksiyskononenko, line 32: So the tests are failing because `.get_sheet_names()` is deprecated, and `.sheetnames` should be used instead: https://openpyxl.readthedocs.io/en/stable/api/openpyxl.workbook.workbook.html#openpyxl.workbook.workbook.Workbook.get_sheet_names

carrascomj:
> @carrascomj do you have any ideas if the issue mentioned [here](https://github.com/h2oai/datatable/issues/2823#issuecomment-755781337) has ever been fixed? Have you had a chance to test `openpyxl` on some larger files?

Ups, I had not seen that issue nor the PR associated with it, sorry. In terms of the mentioned 10x performance decrease, it was a problem with an edge-case: the xlsx contained a link to another file (see https://github.com/pandas-dev/pandas/pull/35029#issuecomment-734888490).

With that said, the code in this PR is far from optimal. I will run some benchmarks and see how far openpyxl can go. However, I agree that a custom parser goes more in line with the perfomance (like it was suggested [before](https://github.com/h2oai/datatable/pull/2851)) expected from datatable.

src/datatable/xlsx.py
oleksiyskononenko: Now it is `DeprecationWarning: Call to deprecated function get_sheet_by_name (Use wb[sheetname])`. Feel free to modify `appveyor.yml` by adding `openpyxl` to the `pip install ...` command, then you should see the same errors on AppVeyor.

src/datatable/xlsx.py
```diff
@@ -0,0 +1,92 @@
...
+            "Module `openpyxl` is required in order to read Excel file '%s'"
+            % filename)
+
+    if subpath:
+        wb = openpyxl.load_workbook(filename, data_only=True)
+        range2d = None
+        if subpath in wb.sheetnames:
+            sheetname = subpath
+        else:
+            if "/" in subpath:
```
carrascomj, line 35: Appveyor is failing in Windows because of the conversion to a path with backslashes.

1. `C:\Path\to\file.xlsx\Sheet\A2:B9` fails (also for xls files with the current xlrd-based parser).
2. `C:\Path\to\file.xlsx\Sheet/A2:B9` also fails, since there is not such a `C:\Path\to\file.xlsx\Sheet` sheet, aftersplitting.
3. `C:\Path\to\file.xlsx/Sheet/A2:B9` works.

The intended behavior in Windows is to work with (1), right? I guess both xlsx.py and xls.py should check for backslashes if no `(sheetname, range2d)` pair was found when a subpath is specified.

oleksiyskononenko: Windows support was added after the excel reading feature, so, you're right, we probably missed the backslash issue. My feeling is that on Windows we should support both types of slashes. What do you think?

oleksiyskononenko:
Thanks for testing @carrascomj. "2x performance decrease" seems quite significant...

src/datatable/xlsx.py
carrascomj: Yes, I guess it would be more consistent for the user if they do not have to change the code for different platforms. I'll give it a try in a day.

oleksiyskononenko: This issue has been fixed as of https://github.com/h2oai/datatable/pull/3220

PR #3198, opened 2021-11-07
Fix `.stype.dtype` for temporal types

argenisleon:
Fix errors when retrieving `df[col_name].stype.dtype` from a `date32` and `time64` columns.

### DIFF
--- a/src/datatable/types.py
+++ b/src/datatable/types.py
@@ -278,6 +278,8 @@ def _init_numpy_transforms():
             stype.str32: np.dtype("object"),
             stype.str64: np.dtype("object"),
             stype.obj64: np.dtype("object"),
+            stype.date32: np.dtype("datetime64"),
+            stype.time64: np.dtype("datetime64")
         }
         _init_value2members_from([
             (np.dtype("void"), stype.void),
@@ -290,6 +292,8 @@ def _init_numpy_transforms():
             (np.dtype("float64"), stype.float64),
             (np.dtype("str"), stype.str64),
             (np.dtype("object"), stype.obj64),
+            (np.dtype("datetime64"), stype.date32),
+            (np.dtype("datetime64"), stype.time64),
         ])
     except ImportError:
         pass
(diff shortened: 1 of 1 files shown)


pradkrish:
Thank you for the contribution. Can you also add some tests?

samukweku:
I might be wrong, but I thought stype/stypes/ltypes are to be deprecated?

oleksiyskononenko:
Thanks, @argenisleon. We're actually going to deprecate and remove `.stype` property (to be replaced by `.Type`): https://datatable.readthedocs.io/en/latest/api/stype.html

But it is obviously a good idea to fix errors like you spotted. Probably, some other new types, like arrays or categoricals, have the same issue.

src/datatable/types.py
```diff
@@ -278,6 +278,8 @@ def _init_numpy_transforms():
             stype.str32: np.dtype("object"),
             stype.str64: np.dtype("object"),
             stype.obj64: np.dtype("object"),
+            stype.date32: np.dtype("datetime64"),
+            stype.time64: np.dtype("datetime64")
```
oleksiyskononenko, line 282:
```suggestion
            stype.date32: np.dtype("datetime64[D]"),
            stype.time64: np.dtype("datetime64[ns]")
```

src/datatable/types.py
```diff
@@ -290,6 +292,8 @@ def _init_numpy_transforms():
             (np.dtype("float64"), stype.float64),
             (np.dtype("str"), stype.str64),
             (np.dtype("object"), stype.obj64),
+            (np.dtype("datetime64"), stype.date32),
+            (np.dtype("datetime64"), stype.time64),
```
oleksiyskononenko, line 296:
```suggestion
            (np.dtype("datetime64[D]"), stype.date32),
            (np.dtype("datetime64[ns]"), stype.time64),
```

oleksiyskononenko (approved):

Issue #2892, opened 2021-03-12
Combination of range() and count() inside update() leads to TypeError


Peter-Pasta:
Hi there I want to enumerate the rows in each group. 
```
from datatable import dt, f, by, update, count

df = dt.Frame(x=["b"] * 3 + ["a"] * 3 + ["c"] * 3)

df['a'] = range(df.nrows) # works fine
df[:, update(b=count()), by('x')] # works fine
df[:, update(c=range(count())), by('x')] # TypeError
```
The last line of code leads to the following error:
```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
TypeError: 'Expr' object cannot be interpreted as an integer
```
Is there a way fix or circumvent this error to get the desired output (only column 'c' is of importance)?:

```
   | x    a   b   c
-- + --  --  --  --
 0 | b    0   3   0
 1 | b    1   3   1
 2 | b    2   3   2
 3 | a    3   3   0
 4 | a    4   3   1
 5 | a    5   3   2
 6 | c    6   3   0
 7 | c    7   3   1
 8 | c    8   3   2
[9 rows x 4 columns]
```

(I asked the same on SO here: https://stackoverflow.com/questions/66585631/python-datatable-assign-new-column-by-group-to-enumerate-elements)

Thank you for your help :)

st-pasha:
Thanks Peter for the question.
The reason `range(count())` doesn't work is because `range` is a pure-python function that expects its argument to be an integer. In this case, `count()` is not an integer - it is an expression that eventually evaluates to an integer within each group of a `DT[i,j,by]` call. However, as `range()` expects a plain integer, it throws an error here.

Unfortunately, I can't think of a way to implement what you're asking -- it may require creating a new function for this task.

st-pasha:
I've checked some other packages, and here's how they do it:
- pandas: [cumcount()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.GroupBy.cumcount.html);
- data.table: `1:.N` or `seq_len(.N)`, both of which are R equivalents of python's `range()`. However, it works in R because they have lazy evaluation semantics built in;
- tidyr: [cumsum()](https://rdrr.io/r/base/cumsum.html)
[...]

PR #3201, opened 2021-11-11
[WIP] Cumsum

samukweku:
- code for dt.cumsum
- WIP
- aim towards cumcount for #2892


samukweku:
@oleksiyskononenko @pradkrish Do you mind having a look at my code, and kindly point out where I'm getting it wrong. My aim is to keep the column height while summing. Thanks

oleksiyskononenko:
@samukweku so what is exactly the issue you're facing here?

samukweku:
@oleksiyskononenko I am not getting a relevant output (note the section I added a comment - line 45). It is a dummy test to return the number of rows (or anything apart from the column values), it just keeps returning the column itself; I'm wondering where I am getting it wrong

src/core/expr/cumsum.cc
```diff
@@ -0,0 +1,120 @@
...
+    const Column& child(size_t) const override {
+      return acol_;
+    }
+
+
+    bool get_element(size_t i, T* out) const override {
+      T a;
+      bool isvalid = acol_.get_element(i, &a);
+      if (isvalid) {
+        *out = static_cast<T>(acol_.nrows()); // testing a dummy output here to return no of rows
```
oleksiyskononenko, line 45: I'm not sure how you say this line could return the column itself. `.nrows()` obviously cannot return a column, because it returns `size_t`. Why do you think this method returns a column?

src/core/expr/cumsum.cc
```diff
@@ -0,0 +1,120 @@
...
+    Column evaluate1(Column&& a) const {
+
+  switch (a.stype()) {
+    case SType::BOOL:
+    case SType::INT8:
+    case SType::INT16:
+    case SType::INT32: return a;
+    case SType::INT64: return a;
+    case SType::FLOAT32: return a;
+    case SType::FLOAT64: return a;
```
oleksiyskononenko, line 92: The issue here is that you always return `a`, that is the original column. What you should do is to call `make(a)` and then return the result.

samukweku: @oleksiyskononenko  is this what you mean :``case SType::INT32: make(a);``

oleksiyskononenko: You should also return it, so something like this will work `case SType::INT32: return make(a);`

samukweku: thanks @oleksiyskononenko ; my next step is how to pass a vector that is the same size ; so instead of return a.nrows(), I wish to return [2, 4, 5, 6] or some other vector that is the same size. I'll gladly take any pointers/guide

oleksiyskononenko: Take a look at documentation "[Creating a new FExpr](https://datatable.readthedocs.io/en/latest/develop/create-fexpr.html)". Though for `cumsum()` it could be more efficient to fill out all the column values with one pass. Otherwise, you will have to re-calculate partial sums over and over again.

samukweku: @oleksiyskononenko still not getting the flow, so some pointers/guidance is definitely needed for me. I thought I could use `partial_sum` func here, but not sure how to do that. The other thought would be to get the first value and store in a variable, and then cumulatively add the rest of the values in the column; I do not know how to select that first value

oleksiyskononenko: `partial_sum` would require a vector or some similar data structure. However, we don't store data in vectors and to access any given element one should use `get_element()` method. Take a look at the link I posted above for an example of how `get_element()` should be used.

samukweku: @oleksiyskononenko made some edits; my outputs are all 0 though. Kindly have a look, when you can and guide me on how to fix it. thanks

src/core/expr/cumsum.cc
```diff
@@ -0,0 +1,127 @@
...
+        if (isvalid){
+          tmp = a + tmp;
+          a = tmp;
+          *out = a;
+        }
+        return true;
+      }
+
+      return false;
+     }
```
oleksiyskononenko, line 55:
```suggestion
    bool get_element(size_t i, T* out) const override {
      bool out_valid = false;
      for (size_t i = 0; i < acol_.nrows(); ++i){
        T a;
        bool isvalid = acol_.get_element(i, &a);
        if (isvalid){
          out_valid = true;
          *out += a;
        }
      }
      return out_valid;
     }
```

oleksiyskononenko: In your code the loop stops after the first iteration due to `return true;`. Also, it will return `true` no matter if the cumsum is valid or not. Take a look at my suggestion, I didn't test it but it should give you a rough idea as to what should happen in the `get_element()`. 

Please note, this code is just a sample. In reality it won't be efficient due to the fact that we re-calculate cumsums over and over again. To avoid it, we should do something similar to [qcut implementation](https://github.com/h2oai/datatable/blob/main/src/core/expr/fexpr_qcut.cc), i.e. wrap the column with the `Latent_ColumnImpl` and calculate all the cumsums in one loop producing material column.

samukweku:
Hi @oleksiyskononenko and happy new year! thanks for the pointers; it is a bit clearer to me; however, I am still stuck on the fact that the values are a single value : 

```py
DT = dt.Frame({'A': range(5)})

DT[:, cumsum(f.A)]
Out[3]: 
   |    C0
   | int32
-- + -----
 0 |    10
 1 |    10
 2 |    10
 3 |    10
 4 |    10
[5 rows x 1 column]
```

I havent focused on the latent column implementation yet, since I want to get the right output before jumping into that. Kindly have a look when you can and please guide me in where my application is wrong. 

Cheers

src/core/expr/cumsum.cc
```diff
@@ -0,0 +1,125 @@
...
+      T tmp = 0;
+      for (size_t i = 0; i < acol_.nrows(); ++i){
+        T a;
+        bool isvalid = acol_.get_element(i, &a);
+        if (isvalid){
+          out_valid = true;
+          tmp = a + tmp;
+          *out = tmp;
+        }
+      }
```
oleksiyskononenko, line 52:
```suggestion
      for (size_t j = 0; j <= i; ++j){
        T a;
        bool isvalid = acol_.get_element(j, &a);
        if (isvalid){
          out_valid = true;
          tmp = a + tmp;
          *out = tmp;
        }
      }
```

oleksiyskononenko:
Thanks, @samukweku. Happy New Year to you too! 

I made some typos when suggesting you a change before. Take a look at the new suggestion, of course, we should make the loop running as long `j <= i`, and not `i <= acol_.nrows()`.

samukweku:
@oleksiyskononenko , do you mind explaining the code pls. what is the value of `I` in the code? shouldn't it be iterating through the values in the column, via a counter ( so through the length of that column). thanks in advance for the explanation.

src/core/expr/cumsum.cc
```diff
@@ -0,0 +1,125 @@
...
+      size_t n_children() const noexcept override {
+      return 1;
+    }
+
+    const Column& child(size_t) const override {
+      return acol_;
+    }
+
+
+    bool get_element(size_t i, T* out) const override {
```
oleksiyskononenko, line 41: @samukweku `i` is the row id that is passed to the `get_element()` method. `j` goes from `0` to `i` and `tmp` accumulates that sum of elements from `0`th to `i`th.

samukweku:
apologies @oleksiyskononenko ; attempted your update, but still failed. my lack of C++ knowledge is showing me up here; I'll close it and hopefully someone else can pick it up. thanks for the guidance.

PR #3220, opened 2021-12-24
Support forward slash, backslash and their mix when reading Excel files

### LINKED ISSUE #3221: `fread` should support both forward and backslash when reading Excel files
Right now we only support forward slash as a separator for Excel subpath, however, since we support datatable on Windows we also need to support the backslash.

oleksiyskononenko:
Closes #3221


oleksiyskononenko:
We still need to find a good way to test this on Windows. The larger `fread` tests were all passed on linux and mac.

PR #3236, opened 2022-02-08
Fix compiler warnings on Py_SET_REFCNT()

vstinner:
Cast size_t reference count to Py_ssize_t.


oleksiyskononenko (approved):

oleksiyskononenko:
AppVeyor failure is probably related to some temporary performance glitch in their system and has nothing to do with this PR. Merging.

PR #3237, opened 2022-02-09
Use `reinterpret_cast` and `nullptr` in `pythoncapi_compat.h`

vstinner:
Adjust `pythoncapi_compat.h` to produce no warnings in C++ compilers.

### DIFF
--- a/src/core/lib/pythoncapi_compat.h
+++ b/src/core/lib/pythoncapi_compat.h
@@ -27,24 +27,34 @@ extern "C" {
 // the inline keyword in C (only in C++): use __inline instead.
 #if (defined(_MSC_VER) && _MSC_VER < 1900 \
      && !defined(__cplusplus) && !defined(inline))
-#  define inline __inline
-#  define PYTHONCAPI_COMPAT_MSC_INLINE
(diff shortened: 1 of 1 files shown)


oleksiyskononenko:
Thanks, now I don't see any warnings produced by `pythoncapi_compat.h`. Is this PR still WIP or you're ok it to be merged?

vstinner:
I was awaiting until the CIs completed for checking for warnings. Thanks for checking for me! The PR is now ready for review ;-)

I updated datatable copy of pythoncapi_compat.h. I already merged the C++ compatibility in pythoncapi_compat ;-)

oleksiyskononenko (approved):

vstinner:
> Thanks, now I don't see any warnings produced by pythoncapi_compat.h.

That's great :-) I tried many options to avoid emitting warnings when including this file in C++.

I still don't get why some pythoncapi_compat.h functions emit warnings, whereas Python.h doesn't.

It seems like "old-style cast" is triggered when **a static function implements a cast inside the static inline function** used in C++. I don't understand if it only emits a warning if the function is used or not.

PR #3248, opened 2022-03-19
Improve documentation for `dt.cut()` and `dt.qcut()`

### LINKED ISSUE #3247: Improve documentation for `dt.cut()`
Hello,

python version is Python 3.9.7 (default, Sep 16 2021, 16:59:28) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32, the version of datatable is version='1.0.0'.

When using the dt.cut() function when passing a range type I get the following:

```
TypeError: bins parameter must be a list or a tuple, instead got <class 'range'>
```

Then I pass it a tuple of interval ranges which looks like this

```
((1, 1001),
 (10001, 11001),
 (20001, 21001),
 (30001, 31001))
```
and get the following error:

```
Expected a Frame, instead got <class 'tuple'>
```

Then tried passing it a Frame (datatable frame), which is the same tuple, but just put into a Frame:

```
bins parameter must be a list or a tuple, instead got <class 'datatable.Frame'>
```

I looked into the "cut()" source code and "Expected a Frame, instead got <class 'tuple'>" error was not found, so I do not know what else to try to pass to the command call, and the documentation of the function does not provide an example class/type which should be passed into the function call so I am a bit stuck, and don't know whether it's a bug or my error?

In R's data table I am used to using the "cut_width()" function which handles this sort of problems easily and pandas cut function also works flawlessly but is very slow on my large dataset (6 tables of 700mil rows).

Also one additional question I have is wether you plan on implementing advanced regex features (not only match) into the datatable FEexpr system, similar to the pandas.str_extract str_remove syntax or the same system already implemented in R?

oleksiyskononenko:
Closes #3247

### DIFF
--- a/docs/api/dt/cut.rst
+++ b/docs/api/dt/cut.rst
@@ -25,10 +25,10 @@
         in `cols`.
 
     bins: List[Frame]
-        List/tuple of single-column frames containing interval edges
-        in strictly increasing order, that will be used for binning
-        of the corresponding columns from `cols`. The list/tuple
-        length must be equal to the number of columns in `cols`.
+        A list or a tuple of frames for binning the corresponding columns from `cols`.
+        Each frame should have only one column that is populated with the edges of
+        the binning intervals in strictly increasing order. Number of elements in
+        `bins` must be equal to the number of columns in `cols`.
 
     right_closed: bool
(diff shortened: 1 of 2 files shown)


samukweku (approved):
Yayy. It should make binning easier to understand

PR #3257, opened 2022-04-07
[ENH] Implement `cumsum()` function

samukweku:
Implement `cumsum()` function.

WIP for #3279


samukweku:
@oleksiyskononenko I would love your input as I am taking another stab at this. I am not sure on how to go about making line 11 ``static auto sum {0};`` better, so it can deal with floats as well. I feel like it is not the right way to do it; your advice here will be really helpful. 

Another issue is the latent column implementation which you referred to in our previous discussion; where does it go to? what does it do? how does it work under the hood?

Also, how do I go about implementing this within a groupby? currently groupby has reducer functions; this however maintains the same height of the the dataframe, like a naive rolling or `dt.update`. fair to say, I do not know how to go about it.

As always, any tips or pointers will help me. thanks.

oleksiyskononenko:
Thanks @samukweku. 

First of all, you can create your branches directly in the datatable repo. In such a case I will be able to push commits to your branch to make adjustments. So it will make our communication more effective.

To do this simply clone datatable repo locally, create a new branch, push your commits to that branch and then push your changes to the origin:

```
git clone https://github.com/h2oai/datatable.git
git checkout -b samukweku/cumsum
# make commits to `samukweku/cumsum` branch
git push
```

I propose you create `samukweku/cumsum` branch under the datatable repo and open a new PR. Then we will go step-by-step through your questions.

oleksiyskononenko:
Sorry, I didn't notice that's the new PR and you closed the old one. This is is good to go and I can push some adjustments here. No actions are needed from your side, let me look at the code and answer your questions.

src/core/expr/cumsum.cc
```diff
@@ -0,0 +1,110 @@
...
+#include "documentation.h"
+#include "expr/fexpr_func.h"
+#include "expr/eval_context.h"
+#include "expr/workframe.h"
+#include "python/xargs.h"
+#include "stype.h"
+namespace dt {
+namespace expr {
+
+static auto sum {0};
```
oleksiyskononenko, line 11: Implementing `dt.cumsum()` based on the static variable will not work. Try to call this function twice on the same data or simply look at the random row and this is what you will see:

```python
>>> from datatable import dt
>>> DT = dt.Frame(range(5))
>>> DT[:, dt.cumsum(dt.f[0])]
   |    C0
   | int32
-- + -----
 0 |     0
 1 |     1
 2 |     3
 3 |     6
 4 |    10
[5 rows x 1 column]
>>> DT[:, dt.cumsum(dt.f[0])]
   |    C0
   | int32
-- + -----
 0 |    10
 1 |    11
 2 |    13
 3 |    16
 4 |    20
[5 rows x 1 column]
```

or

```python
>>> from datatable import dt
>>> DT = dt.Frame(range(5))
>>> DT[1, dt.cumsum(dt.f[0])] # Note, the result should be `1`
   |    C0
   | int32
-- + -----
 0 |     2
[1 row x 1 column]
>>> 
```

oleksiyskononenko: This is because the result is accumulated in the static variable between the calls. Also, there is no loop to calculate `cumsum`, so accessing the random element will not actually do the supposed work. 

If you prefer, I can adjust your code to make it work in the way it is designed now (through the virtual columns) and we will start improving the code from there.

samukweku: @oleksiyskononenko yes, let's do that. it will help me understand more

samukweku:
@oleksiyskononenko just checking if anything is required of me at the moment. Thanks

samukweku:
@oleksiyskononenko thanks for the update! I understand it better now. trapping it within the get_element fixes it up. sweet!

oleksiyskononenko:
@samukweku So this is a working solution, however, it's performance is questionable. That's because for each element we calculate cumulative sum from scratch. For example, if we have a frame

```python
DT = dt.Frame([0, 1, 2, 3, 4])
```

and then do something like
```python
print(DT[:, cumsum(f[:])])
```

then the calculation of `cumsum` will be as follows
```python
0 = 0
1 = 0 + 1
3 = 0 + 1 + 2
6 = 0 + 1 + 2 + 3
10 = 0 + 1 + 2 + 3 + 4
```

However, to calculate the element `i` we can re-use the value of `cumsum` for the element `i - 1`  as 
```
cumsum(i) = cumsum(i - 1) + element(i)
```

Then, cumsum calculation for the example above would look like this
```python
0 = 0
1 = 0 + 1
3 = 1 + 2
6 = 3 + 3
10 = 6 + 4
```

Since it is much more effective to calculate all the cumsums in one pass, it makes no sense to keep the cumsum column virtual. Instead, on the first element access it should do one pass through the data and calculate cumsum for all the rows. This could be implemented with the help of `Latent_ColumnImpl`: https://github.com/h2oai/datatable/blob/main/src/core/column/latent.h

As an example, please take a look at implementation of the `qcut` function: https://github.com/h2oai/datatable/blob/main/src/core/column/qcut.h Especially at the `materialize()` method and how  this column is wrapped with the `Latent_ColumnImpl`: https://github.com/h2oai/datatable/blob/main/src/core/expr/fexpr_qcut.cc#L123

So this is the way to proceed. Let me know if you have any questions.

samukweku:
@oleksiyskononenko made a change regarding the latent column implementation; is that the appropriate application?

src/core/expr/fexpr_cumsum.cc
```diff
@@ -0,0 +1,116 @@
...
+        case SType::FLOAT64: return make<double>(std::move(cumsum), SType::FLOAT64);
+        default:
+            throw TypeError() << "Invalid column of type " << stype0 << " in " << repr();
+      }
+    }
+
+    template <typename T>
+    Column make(Column&& cumsum, SType stype0) const {
+      cumsum.cast_inplace(stype0);
+      return Column(new Latent_ColumnImpl(new Column_cumsum<T>(std::move(cumsum))));
```
oleksiyskononenko, line 99: We also need to change the way we calculate elements for `Column_cumsum`. As I said, instead of calculating elements one by one with `.get_element()`, we need to implement `.materialize()` that will do only one pass through the data.

samukweku: Ahhh ... I see ... Thanks

samukweku: What line is the materialise method in the qcut code?

oleksiyskononenko: Would also be nice if you could move the `Column_cumsum` implementation to the appropriate `.h` file under the `src/core/column/` folder. Just to be consistent with the other virtual columns.

samukweku:
happy holidays @oleksiyskononenko.


I thought I could study `qcut` and reproduce the materialize option, but I am not making it work... I had a look at range's implementation of the materialize and got stuck there. I have updated the PR; I'd appreciate explanation or a break down of where I am getting it wrong. cheers.

src/core/expr/fexpr_cumsum.cc
```diff
@@ -96,7 +120,7 @@ class FExpr_cumsum : public FExpr_Func {
     template <typename T>
     Column make(Column&& cumsum, SType stype0) const {
       cumsum.cast_inplace(stype0);
-      return Column(new Latent_ColumnImpl(new Column_cumsum<T>(std::move(cumsum))));
+      return Column(new Column_cumsum<T>(std::move(cumsum)));
```
oleksiyskononenko, line 123: It is the `Latent_ColumnImpl` that will call `.materialize()` for the `Column_cumsum` column on the first data access. So you should wrap `Column_cumsum` with the `Latent_ColumnImpl` as it was before.

oleksiyskononenko: Btw, the class should be named `Cumsum_ColumnImpl`, because it is inherited from `Virtual_ColumnImpl`.

src/core/expr/fexpr_cumsum.cc
```diff
@@ -23,6 +23,30 @@ class Column_cumsum : public Virtual_ColumnImpl {
...
 
+    void _materialize(Column& out, bool) const {
+      Column newcol = Column::new_data_column(nrows_, stype());
+      T* data = static_cast<T*>(newcol.get_data_editable(0));
+      parallel_for_static(nrows_,
+          [&](size_t i) {
+            data[i] = (i == 0) ? data[i] : data[i] + data[i - 1];
+          });
+        out = newcol;
+      }
```
oleksiyskononenko, line 34: This will not work, because you never use the data from `col_` and this is the column you should calculate running total for.

samukweku:
@oleksiyskononenko made another attempt, with compilation errors, for your guidance again. I have not made the class name changes yet; will do that once I can wrap my head around materialize (and C++ generally). looking forward to your feedback

oleksiyskononenko:
I'm looking at it now, but it seems that there are quite a lot of changes needed to make it work. I will mention some of them in the review and push corrections. Then we will start from there.

src/core/expr/fexpr_cumsum.cc
```diff
@@ -23,30 +24,45 @@ class Column_cumsum : public Virtual_ColumnImpl {
...
-          default:
-            throw RuntimeError() << "Invalid stype for cumsum";
-        }
-      }
+    void materialize(Column& out, bool) override {
+        Column tmp = Column::new_data_column(col_.nrows(), stype());
+        auto data = static_cast<T*>(tmp.get_data_editable());
+         parallel_for_static(col_.nrows(),
+         [&](size_t i) {
+        for (size_t i, i < col_.nrows(), i++) {
```
oleksiyskononenko, line 32: That's not a proper syntax for a for-loop, also, `i` shadows `(size_t i)`. Logic here is also kind of wrong, because `parallel_for_static` already is a loop, so no need to have another loop inside.

oleksiyskononenko:
@samukweku Please take a look at the implementation I did. Still to be done: 
- handle group by;
- add tests just like we have for other f-expressions.
You're welcome to pick it up. Let me know if you have any questions.

oleksiyskononenko:
- `.get_data_editable()` gets you a pointer to the column data buffer that we need to fill out with the running total;
- to study the datatable codebase you can start with [core/datatable.h](https://github.com/h2oai/datatable/blob/main/src/core/datatable.h) gradually moving to other files and reading comments. We have quite a lot of comments in the header files, for instance, if you're interested in the `.get_data_editable()` you can take a look here: 
https://github.com/h2oai/datatable/blob/main/src/core/column.h#L193-L201
- `parallel_for_static` and `parallel_for_dynamic` are different in the way they map iteration blocks to the threads. We have not much on that in the comments: https://github.com/h2oai/datatable/blob/main/src/core/parallel/api.h#L102-L107 
because it is a standard conception similar to the one used in OpenMP: 
https://stackoverflow.com/questions/10850155/whats-the-difference-between-static-and-dynamic-schedule-in-openmp.

Unfortunately, not all the datatable code is full of comments, however, most of the class/method/property names are self-explanatory. If you have questions when browsing the code, don't hesitate to ask here or on gitter.

samukweku:
Once again thanks @oleksiyskononenko ... You made it simple for me to grok... I will continue with it ... Just some things have happened to me that I need some time with. I'll pick it up and update

samukweku:
thanks for the patience @oleksiyskononenko ... I built off using a lot of your ideas from qcut; at face value it makes sense, get the individual groups, run cumsum on each of them, then recombine with rbind; however I get an error when assigning back to `colv[j]` : ``src/core/expr/fexpr_cumsum.cc:69:58: error: unknown type name 'Cumsum_ColumnImpl'; did you mean 'Const_ColumnImpl'?
              colv[j] = Column(new Latent_ColumnImpl(new Cumsum_ColumnImpl(std::move(coli_group))));
                                                         ^~~~~~~~~~~~~~~~~
                                                         Const_ColumnImpl``

oleksiyskononenko:
@samukweku The reason why I had to convert groups to (sub)columns was that `qcut()` is binning values by using its own group-by operation and that group-by needs a column as an input. In the case of `cumsum()` the logic is straightforward once you have access to the group's data: it is just a running sum based within that group. No need to convert the group to a column. It is better to use the `shift()` function as an example: just loop through the groups data and immediately calculate the resulting column values.

See if what I said helps, if not, I will implement it and then you will see what I mean.

samukweku:
thanks @oleksiyskononenko I'll give it a shot

samukweku:
@oleksiyskononenko if you do not mind, kindly implement so I can see what you mean

samukweku:
Thanks @oleksiyskononenko ... Any reason for the parallel use here? Under what conditions do you use parallel static vs dynamic?

oleksiyskononenko:
@samukweku we want to go parallel as much as we can to improve datatable performance. As for the static vs dynamic, I guess we already had this discussion here https://github.com/h2oai/datatable/pull/3257#issuecomment-1108272710

Since groups could have different sizes, sometimes orders of magnitude different — it makes more sense to use dynamic scheduling.

samukweku:
Ahh yes U did explain it previously. I ask about the parallel implementation because in the head_reduce_unary file, it isn't used; now that I understand the benefits, I feel there might be benefits there (naive tests show that it is slower than equivalent pandas/numpy aggregations)

oleksiyskononenko:
Unary reducer returns a virtual column that doesn't process all the groups at once, but only computes values when requested: https://github.com/h2oai/datatable/blob/main/src/core/expr/head_reduce_unary.cc#L97-L101

For cumsum we process all the groups when any element is accessed, because it is more efficient to sweep through all the data just once. So here we are parallelizing the `.materialize()` method over the groups, and there is no sense to parallelize `.get_element()` above, since it only process one value.

oleksiyskononenko:
Yeah, some reducers could be slower because they're implemented as virtual columns — at the same time they save quite a lot of memory as the virtual columns are computed on the fly.

oleksiyskononenko:
@samukweku so look at the code and see if it's ok with you. If it is, you can expand documentation and tests to see if we missed any corner cases.

samukweku:
Done @oleksiyskononenko thanks for d guidance

docs/api/dt/cumsum.rst
```diff
@@ -0,0 +1,83 @@
+
+.. xfunction:: datatable.cumsum
+    :src: src/core/expr/fexpr_cumsum.cc pyfn_cumsum
+    :tests: tests/dt/test-cumsum.py
+    :cvar: doc_dt_cumsum
+    :signature: cumsum(cols)
+
+    .. x-version-added:: 1.1.0
+
+    For each column from `cols` calculate cumulative sum.
```
oleksiyskononenko, line 10: May be add that this function is group-aware?

docs/api/index-api.rst
```diff
@@ -108,6 +108,8 @@ Functions
       - Ternary if operator
     * - :func:`shift()`
       - Shift column by a given number of rows
+    * - :func:`cumsum()`
+      - Calculate cumulative sum
```
oleksiyskononenko, line 112: I guess you already have this function listed in the block below, so here you can remove it.

src/datatable/__init__.py
```diff
@@ -24,6 +24,7 @@
 from .expr import (mean, min, max, sd, isna, sum, count, first, abs, exp,
                    last, log, log10, f, g, median, cov, corr, countna, nunique, prod)
 from .lib._datatable import (
+    cumsum,
```
oleksiyskononenko, line 27: Let's keep alphabetic order here.

src/datatable/__init__.py
```diff
@@ -77,6 +78,7 @@
 
 
 __all__ = (
+    "cumsum",
```
oleksiyskononenko, line 81: and here...

tests/dt/test-cumsum.py
```diff
@@ -0,0 +1,75 @@
...
+import math
+import pytest
+import random
+from datatable import dt, stype, f, cumsum, FExpr, by
+from tests import assert_equals
+
+
+#-------------------------------------------------------------------------------
+# Normal
+#-------------------------------------------------------------------------------
```
oleksiyskononenko, line 33: I would add some tests that throw exceptions: no arguments, unsupported column types, etc. You can take a look at the similar tests for `qcut()` or other functions.

Also, we need a test for string representation, similar to `test_qcut_str`.

oleksiyskononenko:
@samukweku thanks, looks good, let's just improve couple of things here and there, then could be merged.

samukweku:
@oleksiyskononenko  added some tests based on the feedback; is there anything I'm missing there. 

Also, if you don't mind, how do I assign values to a subset of rows (I'm assuming it involves apply row index, not sure), say I have a list of 10 values and I want to change the values at positions 2, 4, 6. The list still contains 10 values, but values at positions 2,  4, 6 are changed. that is achievable with sth similar to this `DT[[2, 4, 6], f.A] = [2, 4, 8]`. just thinking how this is achievable within the `src` code, if I am scribbling a new function.

tests/dt/test-cumsum.py
```diff
@@ -73,3 +79,18 @@ def test_cumsum_grouped_column():
...
+
+def test_cumsum_non_numeric_by():
+    DT = dt.Frame(list('abcde'))
+    with pytest.raises(TypeError, match = 'Invalid column of type.+'):
+        DT[:, cumsum(f[0]), by(f[0])]
+
+def test_cumsum_no_argument():
+    DT = dt.Frame([2, 1, None, 1, 2])
+    with pytest.raises(TypeError, match = r'Function datatable\.cumsum.+'):
+        DT[:, cumsum()]
```
oleksiyskononenko, line 96: Newline is missing here.

tests/dt/test-cumsum.py
```diff
@@ -73,3 +79,18 @@ def test_cumsum_grouped_column():
...
 
+
+def test_cumsum_non_numeric():
+    DT = dt.Frame(list('abcde'))
+    with pytest.raises(TypeError, match = 'Invalid column of type.+'):
+        DT[:, cumsum(f[0])]
+
+def test_cumsum_non_numeric_by():
+    DT = dt.Frame(list('abcde'))
+    with pytest.raises(TypeError, match = 'Invalid column of type.+'):
```
oleksiyskononenko, line 90: May be instead of `.+` add explicitly what is the invalid type?

oleksiyskononenko:
@samukweku nope, rowindex is not related. You just need to get the data writable and update values based on the needed indices. I guess it is better to discuss it on a separate PR or in the gitter chat when you have an example source code.

oleksiyskononenko (approved):

PR #3283, opened 2022-05-19
Make `sum()` and `prod()` reducer be applicable to `void` columns

### LINKED ISSUE #3281: `dt.prod()` cannot be applied to void columns
```python
from datatable import dt, f
DT = dt.Frame([None])
DT[:, dt.prod(f.C0)]
```
results in
```python
TypeError: Unable to apply reduce function prod() to a column of type void
```

### LINKED ISSUE #3282: `dt.sum()` cannot be applied to void columns
```python
from datatable import dt, f
DT = dt.Frame([None])
DT[:, dt.sum(f.C0)]
```
results in
```python
TypeError: Unable to apply reduce function sum() to a column of type void
```

oleksiyskononenko:
WIP for #3284 
Closes #3281 
Closes #3282

### DIFF
--- a/docs/api/dt/cumsum.rst
+++ b/docs/api/dt/cumsum.rst
@@ -7,7 +7,8 @@
 
     .. x-version-added:: 1.1.0
 
-    For each column from `cols` calculate cumulative sum. In the presence of :func:`by()`,
+    For each column from `cols` calculate cumulative sum. The sum of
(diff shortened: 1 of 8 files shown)


oleksiyskononenko:
My feeling is that we also need to check other reducers like `min()`, `max()`, etc. when applied to void columns.

samukweku (approved):

Issue #3279, opened 2022-05-14
Implement cumulative functions


samukweku:
The list of functions to be implemented and the corresponding PR's
- [x] `cumsum()` https://github.com/h2oai/datatable/pull/3257
- [x] `cumprod()` https://github.com/h2oai/datatable/pull/3304
- [x] `cummax()` https://github.com/h2oai/datatable/pull/3288
- [x] `cummin()` https://github.com/h2oai/datatable/pull/3288
- [x] `cumcount()` https://github.com/h2oai/datatable/pull/3310
- [x] `ngroup()` - not strictly cumulative https://github.com/h2oai/datatable/pull/3310
- [x] `fillna()` for forward/backward fill https://github.com/h2oai/datatable/pull/3311
- [x] `fillna()` for filling with a value https://github.com/h2oai/datatable/pull/3344 
~~- [ ] rank~~ continued on #3148 
~~- [ ] rolling aggregations~~    continued on #1500

vopani:
These would be great and long due.

Is there an ETA of when this would be available (next release)?

samukweku:
@vopani currently working on them ... cant specify an ETA though ... contributions are welcome also. You can test the `cumsum` function by downloading the latest dev version

vopani:
I tried out the dev version to implement the [AMEX Default Metric](https://www.kaggle.com/competitions/amex-default-prediction/overview/evaluation) as described here: https://www.kaggle.com/code/rohanrao/amex-competition-metric-implementations#Datatable-Implementation

When I use the native `cumsum` method, it runs much slower compared to converting to `pandas` and using its `cumsum` and converting back to `datatable`.

Here's the entire code for reference: 

```python
# install latest datatable
!pip install git+https://github.com/h2oai/datatable

## import packages
import datatable as dt
import random

## check version
print(dt.__version__)

>>> 
1.1.0a0+build.1653978586.root
>>>

# load data
train_labels = dt.fread('../input/amex-default-prediction/train_labels.csv')['target']
predictions = dt.Frame((train_labels.to_numpy().ravel() / 2.0 + [random.random() for i in range(train_labels.nrows)]) / 2.0)

## amex metric
def amex_metric_old(y_true: dt.Frame, y_pred: dt.Frame) -> float:

    # create datatable frame
    df = dt.Frame(target=y_true, prediction=y_pred)

    # sort by descending prediction values
    df = df[:, :, dt.sort(-dt.f.prediction)]

[...]

PR #3288, opened 2022-05-30
[ENH] Add `cummin()` and `cummax()` functions for cumulative min/max calculation

samukweku:
WIP for #3279


oleksiyskononenko:
Thanks @samukweku, I guess this could be easily templated to support both `cummax()` and `cummin()` at the same time. Look at how this is implemented for `rowmax()`/`rowmin()`, for example.

docs/api/dt/cummax.rst
```diff
@@ -0,0 +1,88 @@
...
+.. xfunction:: datatable.cummax
+    :src: src/core/expr/fexpr_cummax.cc pyfn_cummax
+    :tests: tests/dt/test-cummax.py
+    :cvar: doc_dt_cummax
+    :signature: cummax(cols)
+
+    .. x-version-added:: 1.1.0
+
+    For each column from `cols` calculate cumulative maximum. The maximum of
+    the missing values is calculated as zero. In the presence of :func:`by()`,
```
oleksiyskononenko, line 11: Normally, maximum of the missing value is also missing:
```python
>>> import datatable as dt
>>> DT = dt.Frame([None])
>>> DT.max()
   |   C0
   | void
-- + ----
 0 |   NA
[1 row x 1 column]
```

docs/api/dt/cummax.rst
```diff
@@ -0,0 +1,88 @@
...
+    the cumulative maximum is computed per group.
+
+    Parameters
+    ----------
+    cols: FExpr
+        Input data for cumulative maximum calculation.
+
+    return: FExpr
+        f-expression that converts input columns into the columns filled
+        with the respective cumulative maximum.
```
oleksiyskononenko, line 21: "maximum" -> "maximums"?

docs/api/dt/cummax.rst
```diff
@@ -0,0 +1,88 @@
...
+        >>> DT[:, dt.cummax(f[:-1])]
+           |     A      B        C
+           | int64  int64  float64
+        -- + -----  -----  -------
+         0 |     2      0      5.4
+         1 |     2      0      5.4
+         2 |     5      0      5.4
+         3 |     5      0      5.4
+         4 |     5      0      5.4
+        [5 rows x 3 columns]
```
oleksiyskononenko, line 74: My feeling is that the second column should be calculated as all missings. See my comment above.

docs/api/dt/cummax.rst
```diff
@@ -0,0 +1,88 @@
...
+        >>> DT[:, dt.cummax(f[:]), by('D')]
+           | D          A      B        C
+           | str32  int64  int64  float64
+        -- + -----  -----  -----  -------
+         0 | a          2      0    5.4  
+         1 | a          2      0    5.4  
+         2 | b          5      0    2.2  
+         3 | b          5      0    4.323
+         4 | b          5      0    4.323
+        [5 rows x 4 columns]
```
oleksiyskononenko, line 88: Same here.

src/core/column/cummax.h
```diff
@@ -0,0 +1,93 @@
...
+      auto offsets = gby_.offsets_r();
+      dt::parallel_for_dynamic(
+        gby_.size(),
+        [&](size_t gi) {
+          size_t i1 = size_t(offsets[gi]);
+          size_t i2 = size_t(offsets[gi + 1]);
+
+          T val;
+          bool is_valid = col_.get_element(i1, &val);
+          data[i1] = is_valid? val : 0;
```
oleksiyskononenko, line 60: So logic should be adjusted here, if the element is not valid then instead of `0` we should return `GETNA<T>()`.

src/core/column/cummax.h
```diff
@@ -0,0 +1,93 @@
...
+
+          for (size_t i = i1 + 1; i < i2; ++i) {
+            is_valid = col_.get_element(i, &val);
+            val = is_valid? val : 0;
+            data[i] = data[i - 1] > val ? data[i - 1] : val;
```
oleksiyskononenko, line 65: Also here, if the element is not valid we should check the previous maximum validity:
- if the previous maximum is not valid, then set the current maximum to be `GETNA<T>()`;
- if the previous maximum is valid, then set the current maximum to be equal to the previous maximum.

src/core/expr/fexpr_cummax.cc
```diff
@@ -0,0 +1,113 @@
...
+        default: throw TypeError()
+          << "Invalid column of type " << stype << " in " << repr();
+      }
+    }
+
+
+    template <typename T>
+    Column make(Column&& col, SType stype, const Groupby& gby) const {
+      if (col.stype() == SType::VOID) {
+        return Column(new ConstInt_ColumnImpl(col.nrows(), 0, stype));
```
oleksiyskononenko, line 89:
```suggestion
        return Column(new ConstNa_ColumnImpl(col.nrows()));
```

tests/dt/test-cummax.py
```diff
@@ -0,0 +1,104 @@
...
+    DT = dt.Frame()
+    expr_cummax = cummax(DT)
+    assert isinstance(expr_cummax, FExpr)
+    assert_equals(DT[:, f[:]], DT)
+
+
+def test_cummax_void():
+    DT = dt.Frame([None, None, None])
+    DT_cummax = DT[:, cummax(f[:])]
+    assert_equals(DT_cummax, dt.Frame([0, 0, 0]/dt.int64))
```
oleksiyskononenko, line 73: Need to adjust this test after we fix the issue with the void columns returning zero instead of a void column.

oleksiyskononenko:
Look at this example. In the current implementation `cummax()` becomes zero once a missing value is encountered:
```python
>>> from datatable import dt
>>> DT = dt.Frame([-1, None, 1])
>>> DT[:, dt.f.C0.cummax()]
   |    C0
   | int64
-- + -----
 0 |    -1
 1 |     0
 2 |     1
[3 rows x 1 column]
```

Once we fix the issues I've mentioned in the comments above the output will be
```
   |    C0
   | int64
-- + -----
 0 |    -1
 1 |     -1
 2 |     1
[3 rows x 1 column]
```
and that will be more consistent.

oleksiyskononenko:
Btw, I've just pushed a fix for the sphinx build configuration. You should merge `main` onto your branch to make your build green.

samukweku:
> Thanks @samukweku, I guess this could be easily templated to support both `cummax()` and `cummin()` at the same time. Look at how this is implemented for `rowmax()`/`rowmin()`, for example.

thanks for the hint ... I'll give it a shot

oleksiyskononenko:
@samukweku the building error is not related to the changes on this PR, just some glitch with the jenkins filesystem:
```
[2022-06-04T00:58:55.741Z]     def test_fread_from_url2():
[2022-06-04T00:58:55.741Z]         import pathlib
[2022-06-04T00:58:55.741Z]         root = os.path.join(os.path.dirname(__file__), "..", "..")
[2022-06-04T00:58:55.741Z]         path = os.path.abspath(os.path.join(root, "LICENSE"))
[2022-06-04T00:58:55.741Z]         url = pathlib.Path(path).as_uri()
[2022-06-04T00:58:55.741Z]         d0 = dt.fread(url, sep="\n")
[2022-06-04T00:58:55.741Z]         frame_integrity_check(d0)
[2022-06-04T00:58:55.741Z] >       assert d0.shape == (372, 1)
[2022-06-04T00:58:55.741Z] E       assert (0, 0) == (372, 1)
[2022-06-04T00:58:55.741Z] E         At index 0 diff: 0 != 372
[2022-06-04T00:58:55.741Z] E         Full diff:
[2022-06-04T00:58:55.741Z] E         - (372, 1)
[2022-06-04T00:58:55.741Z] E         + (0, 0)
```

So just continue what you're doing, if the error persists, I will look into it.

src/core/expr/fexpr_cummax.cc
```diff
@@ -0,0 +1,113 @@
...
+
+    Column evaluate1(Column&& col, const Groupby& gby) const {
+      SType stype = col.stype();
+      switch (stype) {
+        case SType::VOID:
+        case SType::BOOL:
+        case SType::INT8:
+        case SType::INT16:
+        case SType::INT32:
+        case SType::INT64: return make<int64_t>(std::move(col), SType::INT64, gby);
```
oleksiyskononenko, line 77: `cummax()` should return the same type as the source column has. Here we return `int64` for all the integer/boolean/void types.

samukweku: Ahhh my bad... I'll fix it. I'll need ur help on templating and multiple functions. Couldn't wrap my head around the solution in rowminmax. I won't mind getting a solution for this with comments. I'll make the updates so you can input yours

src/core/expr/fexpr_cummax.cc
```diff
@@ -0,0 +1,113 @@
...
+          << "Invalid column of type " << stype << " in " << repr();
+      }
+    }
+
+
+    template <typename T>
+    Column make(Column&& col, SType stype, const Groupby& gby) const {
+      if (col.stype() == SType::VOID) {
+        return Column(new ConstNa_ColumnImpl(col.nrows()));
+      } else {
```
oleksiyskononenko, line 90: If we add the above shortcut, this block can be safely removed — since `make()` won't be called on the void columns. Actually, we can even remove the `stype` parameter from `make()`, because the stype returned should always be the same as `col.stype()`.

src/core/expr/fexpr_cummax.cc
oleksiyskononenko: Sure, I can help you with templating this function. First, let's fix couple of issues in the current implementation of `cummax()`, then we will template it to also support `cummin()`.

samukweku: unsure what to use as the type for bool ... used bool8, but it seemed to have an issue with get_na<T>

oleksiyskononenko: `bool` should be handled in the same way as `int8`.

samukweku: I assumed we needed to pass a make for book, similar to the other types

oleksiyskononenko: Yes, but internally `bool` is equivalent to `int8`, so you just pass `int8` as a template parameter and the `stype` for the resulting column will be the same as the source column stype.

docs/api/dt/cummin.rst
```diff
@@ -0,0 +1,87 @@
...
+    return: FExpr
+        f-expression that converts input columns into the columns filled
+        with the respective cumulative minimums.
+
+    except: TypeError
+        The exception is raised when one of the columns from `cols`
+        has a non-numeric type.
+
+
+    Examples
```
oleksiyskononenko, line 27: @samukweku May I ask you to adjust your examples for `cummin()`? Currently I just copy-pasted these examples from `cummax()`.

oleksiyskononenko:
@samukweku Let me know if you have any questions on the templated implementation of `cummin()`/`cummax()` functions.

samukweku:
thanks @oleksiyskononenko , I understand the template implementation. I'm sure I'll come back to it several times to study it more.

samukweku:
@oleksiyskononenko kindly have a look, when you can; I have updated the docs for cumin and added tests for `f.cummin`

docs/api/index-api.rst
```diff
@@ -163,6 +163,8 @@ Functions
       - Count non-missing values per column
     * - :func:`countna()`
       - Count the number of NA values per column
+    * - :func:`cummax()`
+      - Calculate the cumulative maximum of values per column
```
oleksiyskononenko, line 167: Isn't `cummin()` is missing here?

docs/releases/v1.1.0.rst
```diff
@@ -97,6 +101,9 @@
     -[fix] Reducers and row-wise functions now support :attr:`void <dt.Type.void>`
        columns. [#3284]
 
+    -[fix] All the row-wise functions now support :attr:`void <dt.Type.void>`
+       columns. [#3284]
+
```
oleksiyskononenko, line 106: This should be removed as we already mention `row-wise functions` above. Probably this piece of code is left from one of the old merges from main.

docs/api/index-api.rst
```diff
@@ -234,6 +236,8 @@ Other
     count()           <dt/count>
     countna()         <dt/countna>
     cov()             <dt/cov>
+    cummin()          <dt/cummin>
+    cummax()          <dt/cummax>
```
oleksiyskononenko, line 240: I guess I put it in the wrong order here, `cummax()` should go first alphabetically. May be you could change the order here and below.

src/core/documentation.h
```diff
@@ -30,6 +30,8 @@ extern const char* doc_dt_corr;
 extern const char* doc_dt_count;
 extern const char* doc_dt_countna;
 extern const char* doc_dt_cov;
+extern const char* doc_dt_cummin;
+extern const char* doc_dt_cummax;
```
oleksiyskononenko, line 34: Order.

src/core/documentation.h
```diff
@@ -279,6 +281,8 @@ extern const char* doc_FExpr;
 extern const char* doc_FExpr_as_type;
 extern const char* doc_FExpr_count;
 extern const char* doc_FExpr_countna;
+extern const char* doc_FExpr_cummin;
+extern const char* doc_FExpr_cummax;
```
oleksiyskononenko, line 285: Same here.

src/core/expr/fexpr.h
```diff
@@ -182,6 +182,8 @@ class PyFExpr : public py::XObject<PyFExpr> {
     py::oobj as_type(const py::XArgs&);
     py::oobj count(const py::XArgs&);
     py::oobj countna(const py::XArgs&);
+    py::oobj cummin(const py::XArgs&);
+    py::oobj cummax(const py::XArgs&);
```
oleksiyskononenko, line 186: Same here.

src/datatable/__init__.py
```diff
@@ -27,6 +27,8 @@
     as_type,
     by,
     cbind,
+    cummin,
+    cummax,
```
oleksiyskononenko, line 31: Also here.

src/datatable/__init__.py
```diff
@@ -85,6 +87,8 @@
     "corr",
     "count",
     "cov",
+    "cummin",
+    "cummax",
```
oleksiyskononenko, line 91: And here.

tests/test-f.py
```diff
@@ -444,3 +444,15 @@ def test_cumsum():
...
+    assert str(dt.cummax(f.A)) == str(f.A.cummax())
+    assert str(dt.cummax(f[:])) == str(f[:].cummax())
+    DT = dt.Frame(A = [9, 8, 2, 3, None, None, 3, 0, 5, 5, 8, None, 1])
+    assert_equals(DT[:, f.A.cummax()], DT[:, dt.cummax(f.A)])
+
+def test_cummin():
+    assert str(dt.cummin(f.A)) == str(f.A.cummin())
+    assert str(dt.cummin(f[:])) == str(f[:].cummin())
+    DT = dt.Frame(A = [9, 8, 2, 3, None, None, 3, 0, 5, 5, 8, None, 1])
+    assert_equals(DT[:, f.A.cummin()], DT[:, dt.cummin(f.A)])
```
oleksiyskononenko, line 458: Newline is missing.

src/datatable/__init__.py
samukweku: by the way is there any particular reason for the alphabetical arrangement?

oleksiyskononenko: When we have tons of functions it just makes it easier to navigate and look for some specific things. Both for the source code (developers) and documentation (users).

src/core/expr/fexpr.cc
```diff
@@ -510,103 +588,16 @@ DECLARE_METHOD(&PyFExpr::shift)
...
+  auto sumFn = oobj::import("datatable", "sum");
+  return sumFn.call({this});
 }
 
-DECLARE_METHOD(&PyFExpr::cummin)
-    ->name("cummin")
-    ->docs(dt::doc_FExpr_cummin);
+DECLARE_METHOD(&PyFExpr::sum)
+    ->name("sum")
+    ->docs(dt::doc_FExpr_sum);
```
oleksiyskononenko, line 598: Oh, that was just a request to arrange the lists. No need to worry about all the function implementations, otherwise we will need to move so much code :)

oleksiyskononenko (approved):

PR #3308, opened 2022-07-06
forward and backward fill

samukweku:
- implementation for forward and backward fill
- WIP #3279


samukweku:
@oleksiyskononenko when you have some time, pls could you guide me on string/timestamp operations; I tried using the `make` function to replicate the strings but it failed and I couldn't really decipher the error message.

I also need your help with the documentation; I'm not sure where I bungled it, but the docs do not render

src/core/expr/fexpr_ffill_bfill.cc
```diff
@@ -0,0 +1,125 @@
...
+        default: throw TypeError()
+          << "Invalid column of type `" << stype << "` in " << repr();
+      }
+    }
+
+
+    template <typename T>
+    Column make(Column&& col, const Groupby& gby) const {
+      return Column(new Latent_ColumnImpl(
+        new Ffill_ColumnImpl<T, MIN>(std::move(col), gby)
```
oleksiyskononenko, line 94: I don't think that replacing missings requires a latent column. Latent columns are used to materialize the virtual column on the first data access. When we need to replace missings, we don't need to create a material column. `get_element()` gets a row id, then we just find the first non-missing value (in the appropriate direction, depending if that's the forward or backward fill) and return it.

oleksiyskononenko:
Before moving forward, didn't you think about implementing this as a single `fillna()` function?

src/core/expr/fexpr_ffill_bfill.cc
samukweku: thanks for the feedback; I'll spend some time to rework the code with this logic; if I get stuck I'll bounce it off you for guidance

samukweku:
> Before moving forward, didn't you think about implementing this as a single `fillna()` function?

That was my initial plan, I was not sure if it was accepted. I'll give it a go

samukweku:
I'll start afresh on this; hopefully fix the doc rendering issue

PR #3310, opened 2022-07-07
[ENH] Add `cumcount()` and `ngroup()` functions

### LINKED ISSUE #2892: Combination of range() and count() inside update() leads to TypeError

samukweku:
WIP for #3279 
Closes #2892


samukweku:
@oleksiyskononenko; I decided to continue with `cumcount` - approaching the functions in a logical order as arranged in #3279 

I need your input here; my idea for cumcount is to add a new column that shows the row numbers basically; I am modelling it based on Pandas approach , using the `reverse` parameter (similar to that in sort) where the numbers are in ascending or descending order, depending on the boolean value passed (as an aside, is it more intuitive to use `reverse` or `ascending`?)

How do I add a new column, since the arguments does not need a column. Kindly have a look at the code to see where I am getting it wrong. The aim is once I can get the column, I can then create a `.h` file and simply create a materialised column to get the row numbers (if there is a more efficient path, I am open to that as well).

Thanks @oleksiyskononenko

samukweku:
@oleksiyskononenko I am stuck at this point and would appreciate your feedback. this is the error message I encounter: 

```
src/core/column/cumcount.h:42:5: error: expected '{' or ','
    void materialize(Column& col_out, bool) override {
    ^
````

**Comment by samukweku** [comment]
@oleksiyskononenko I have implemented the `cumcount` and `ngroup` functions. When you have the time, kindly have a look and review. 

I also have an issue where `update` does not work on the functions; it works if `by` is involved; without a by though it fails

**Comment by oleksiyskononenko** [comment]
I've tried to build your PR locally and getting quite a lot of warnings

```cpp
In file included from src/core/expr/fexpr_cumcountngroup.cc:23:
src/core/column/cumcountngroup.h:33:12: warning: non-static data member 'nrows_' of 'CUMCOUNTNGROUP_ColumnImpl' shadows member inherited from type 'ColumnImpl' [-Wshadow-field]
    size_t nrows_;
           ^
src/core/column/column_impl.h:56:12: note: declared here
    size_t nrows_;
           ^
src/core/expr/fexpr_cumcountngroup.cc:35:11: warning: padding size of 'dt::expr::FExpr_CumCount<true>' with 7 bytes to alignment boundary [-Wpadded]
    class FExpr_CumCount : public FExpr_Func {
          ^
src/core/expr/fexpr_cumcountngroup.cc:35:11: warning: padding size of 'dt::expr::FExpr_CumCount<false>' with 7 bytes to alignment boundary [-Wpadded]
In file included from src/core/expr/fexpr_cumcountngroup.cc:23:
src/core/column/cumcountngroup.h:35:13: warning: padding class 'dt::CUMCOUNTNGROUP_ColumnImpl<true>' with 7 bytes to align 'gby_' [-Wpadded]
    Groupby gby_;
...
```

See if you can fix them. If not, let me know and I will help.

src/core/column/cumcountngroup.h
```diff
@@ -0,0 +1,81 @@
...
+          *out = reverse_?i1-1-i:i-i0;
+          } else {
+            *out = reverse_?grp_size-1-j:j;
+          }
+          break;
+         }
+      }
+      return true;
+
+    }
```
oleksiyskononenko, line 73: I guess we can implement a binary search here, since the offsets array is sorted.

oleksiyskononenko:
@samukweku so what is your expectation for `dt.ngroup()` to do with no group by context? Do you want to throw an error or silently produce an all zeros group?

src/core/column/cumcountngroup.h
samukweku: What advantage does a binary search offer here? If you don't mind could you explain more

samukweku:
> @samukweku so what is your expectation for `dt.ngroup()` to do with no group by context? Do you want to throw an error or silently produce an all zeros group?

I'm open to any idea. Ngroup is really useful only in a grouped context

src/core/column/cumcountngroup.h
oleksiyskononenko: It will just be faster in identifying the correct group. Right not you just go from group 0 all the way until the needed group. With the binary search you start with the middle group and then move to left or right depending on where your element should be.

Look here: https://en.wikipedia.org/wiki/Binary_search_algorithm#Algorithm

samukweku:
> I've tried to build your PR locally and getting quite a lot of warnings
> 
> ```c++
> In file included from src/core/expr/fexpr_cumcountngroup.cc:23:
> src/core/column/cumcountngroup.h:33:12: warning: non-static data member 'nrows_' of 'CUMCOUNTNGROUP_ColumnImpl' shadows member inherited from type 'ColumnImpl' [-Wshadow-field]
>     size_t nrows_;
>            ^
> src/core/column/column_impl.h:56:12: note: declared here
>     size_t nrows_;
>            ^
> src/core/expr/fexpr_cumcountngroup.cc:35:11: warning: padding size of 'dt::expr::FExpr_CumCount<true>' with 7 bytes to alignment boundary [-Wpadded]
>     class FExpr_CumCount : public FExpr_Func {
>           ^
> src/core/expr/fexpr_cumcountngroup.cc:35:11: warning: padding size of 'dt::expr::FExpr_CumCount<false>' with 7 bytes to alignment boundary [-Wpadded]
> In file included from src/core/expr/fexpr_cumcountngroup.cc:23:
> src/core/column/cumcountngroup.h:35:13: warning: padding class 'dt::CUMCOUNTNGROUP_ColumnImpl<true>' with 7 bytes to align 'gby_' [-Wpadded]
>     Groupby gby_;
> ...
> ```
> 
> See if you can fix them. If not, let me know and I will help.

thanks @oleksiyskononenko , how do I go about the padding warning?

samukweku:
@oleksiyskononenko updated the code with your feedback. I still have issues using `update` for non grouping operations for  `cumcount`

oleksiyskononenko:
I've been thinking through it and guess that doing search in `cumcount()` and `ngroup()` may not be computationally effective. If these functions return material columns, we could trivially fill them with the proper value/group ids. Yes, in this case we may need somewhat more memory to be used, but then we save a lot in the computation time.

Let me correct the way it works now and also look into the update issue.

samukweku:
@oleksiyskononenko any pointers on what determines when to materialise vs sticking to virtual columns

oleksiyskononenko:
In theory, we should try to save as much memory as possible by sticking to virtual columns. When the function's logic is simple and there is no way to profit from calculating all the elements at once, the virtual column is a good choice. 

In the cases when the calculation of one element is quite expensive (like we need to run a search for each element) and calculation of all elements at once is much faster than nrows * (calculation of one element), it makes sense to opt for the material column. That's the case for `qcut()`, `cumsum()`, `cummax()`, etc. 

Here, if we only need to enumerate rows with no group by context, we can return a `Range_ColumnImpl()` — that's a virtual column. For `ngroup()` with no group by context, we can also return a virtual const column.

However, within the group by context, we need to run a search for each element to know which group it is in. Even with the binary search it seems to be an overkill, because if the user needs all the `cumcount()`s, we need to run nrows of searches. However, If we just loop through all the groups, we can trivially fill `cumcount`s and `ngroup`s. That's why I believe we should switch to material columns for group by context.

oleksiyskononenko (approved):

Issue #2551, opened 2020-07-23
group by (datatable vs pandas)


acpguedes:
Hi, 

I tested a `group_by` operation in a 9.6GB file, my results don't agree with [this benchmark](https://h2oai.github.io/db-benchmark/).

env:
2 x Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Ram 256G  
```
Python 3.7.1 (default, Dec 14 2018, 19:28:38) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.
```
Test: **Read**
```
import datatable as dt
%timeit dt.fread("/projects/salmonella/data/cgr.liv.ac.uk/parsed/cgr.parsed.features.tsv")
#42.3 s ± 3.05 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
import pandas as pd
%timeit pd.read_csv("/projects/salmonella/data/cgr.liv.ac.uk/parsed/cgr.parsed.features.tsv", sep = "\t")
#1min 28s ± 347 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I also try figure out the conversion from `dataframe` to `pandas`:
```
%timeit dfp = df.to_pandas()                                                                                                                 
57.8 s ± 1.81 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Too expensive conversions, there is others package that converts R data.frame to Pandas really fast. 

But the point is, I tried the `group_by` with `count`:
test: **group by**
```
#datatable 
%timeit df[ : , : , dt.by("nucleotide")]                                                                                                     
6.7 s ± 939 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
#panda 
%timeit dfp.groupby("nucleotide").size()                                                                                                     
2.47 s ± 41.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Or there is a better way to perform count by group with `datatable`?

samukweku:
``dt.count`` is used to count groupings, but this is missing in your code. is it a typo?

goldentom42:
Hi @acpguedes as mentioned by @samukweku you can use `dt.count` in this way:

```
df[ : , dt.count() , dt.by("nucleotide")]
```
 
Hope this helps.

acpguedes:
Thanks, @samukweku and @goldentom42 

Actually it was a typo, but...
```
[...]

Issue #2541, opened 2020-07-21
performance regression when by acts on multiple columns


chi2liu:
For example, we have a frame which the shape of it is(34284584, 18), 
when by only acts on one column or a few columns, like:

`df[:, {"A":min(f.A), by("B")}] or df[:, {"A":min(f.A), by(["B", "C"])}]`

It takes about seconds.

But when by acts on multiple columns, the time-consuming increases to the minute level, like:

`df[:, {"A":min(f.A), by(["B", "C", "D", "E", "F", "H", "G", "J", "K", "M")}]`

Performance is even weaker than single-threaded pandas.

Is there a performance problem with groupby based on multiple columns?
Thanks a lot!

chi2liu:
Here is a simple running examples:

![image](https://user-images.githubusercontent.com/19861628/88053290-208b0e00-cb8e-11ea-8707-8106e203ec51.png)

PR #3311, opened 2022-07-14
[ENH] Add `dt.fillna()` function to impute missing values

samukweku:
Add `dt.fillna()` function to replace missing values with the previous/subsequent non-missing.

WIP for #3279


samukweku:
I can piggyback on #3310 on resolving the padding warnings, once you are done with the update. For fillna, I feel all columns should be able to take advantage of it, excluding maybe the array columns. If you do not mind @oleksiyskononenko , how do I go about fixing the error when running the function on string and time/date columns? thanks

samukweku:
I'm also thinking it might be ok, as a convenience option to fill nulls here with a scalar, similar to what `replace`/`if else` would do @oleksiyskononenko

oleksiyskononenko:
Yes, I think `fillna()` could be applicable to all the columns. Internally, `DATE32` should be processed at `INT32` and `TIME64` as `INT64`.

src/core/column/fillna.h
```diff
@@ -0,0 +1,101 @@
...
+      :Virtual_ColumnImpl(col.nrows(), col.stype()),
+       col_(std::move(col)),
+       direction_(direction),
+       gby_(gby)
+    {
+      xassert(col_.can_be_read_as<T>());
+    }
+
+    void materialize(Column &col_out, bool) override {
+      Column col = Column::new_data_column(col_.nrows(), col_.stype());
```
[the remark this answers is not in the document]
samukweku: I'll look into it and see how far I go. thanks for the prompts @oleksiyskononenko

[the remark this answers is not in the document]
samukweku: @oleksiyskononenko how do I get the row indices?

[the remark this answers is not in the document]
oleksiyskononenko: You need to create it. Look at how sorting works, for instance. We do not create a new data column with the sorted data, instead, we create a `RowIndex` and fill it with ids of the sorted elements. 

So what you need to do here is to create a new `RowIndex` and 
- for non-missing elements fill it with the same row ids as in the input column;
- for missing elements fill it with the row ids of the corresponding non-missings.

Once you have such a `RowIndex` you will simply apply it to your original column as `col.apply_rowindex(ri)`.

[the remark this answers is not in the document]
samukweku: I'll need your help more on this @oleksiyskononenko - if you can scribble sth on how to use the row index, that would be helpful and definitely be useful in other PRs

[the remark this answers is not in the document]
oleksiyskononenko: I guess most of the `RowIndex` methods are documented within the code here: https://github.com/h2oai/datatable/blob/main/src/core/rowindex.h Please take a look. If you have any other particular questions, please let me know.

docs/api/dt/fillna.rst
```diff
@@ -0,0 +1,157 @@
...
+    For each column from `cols` fill the null values down or up with the previous/next non-null value. 
+
+    In the presence of :func:`by()`, the last non-null value is propagated down or up within each group.
+
+    Parameters
+    ----------
+    cols: FExpr
+        Input data for filling the nulls values down or up.
+
+    direction: string
```
oleksiyskononenko, line 19: May be we call it `reverse` and make it `bool`? To be consistent with all the other places where we have the `reverse` parameter.

samukweku: Ahhhh nice pick up. Will do

samukweku:
@oleksiyskononenko when you can kindly have a look at my mock code; haven't gotten far with it, as I am still unsure about the RowIndex. My attempts are expensive when converting the boolean column to a row index, from microseconds to milliseconds for 5 rows.

src/core/expr/fexpr_fillna.cc
```diff
@@ -58,56 +60,17 @@ namespace dt {
...
-          //   //wf.replace_column(i, std::move(coli));
-          // }
-          return wf;
+          Workframe out(ctx);
+          for (size_t i = 0; i < wf.ncols(); ++i) {
+            Column coli = wf.retrieve_column(i);
+            coli = Column(new Isna_ColumnImpl<int8_t>(std::move(coli)));
+            RowIndex ri = RowIndex(std::move(coli));
+            coli.apply_rowindex(ri.negate(coli.nrows())); // i get bools here
+            out.add_column(std::move(coli), std::string(), out.get_grouping_mode());
```
oleksiyskononenko, line 69: What you need to do here is to go through all the elements in the column `coli` and depending on their validity fill the RowIndex `ri`:
- if `i`-th element is valid, then `ri[i]` should be set to `i`;
- if `i`-th element is not valid, then `ri[i]` should be set to the index of the last valid element.

At the end you need to apply `ri` to the original column `coli`.

samukweku: @oleksiyskononenko I am getting this error message : 
```
src/core/expr/fexpr_fillna.cc:73:17: error: type 'RowIndex' does not provide a subscript operator
              ri[j] = prev;
```

oleksiyskononenko: Yeah, the subscript operator is defined for the underlying `Buffer` only. See, for example, how it is done for the `shift()` function: https://github.com/h2oai/datatable/blob/main/src/core/expr/head_func_shift.cc#L43-L44

Once the `Buffer` is filled out, you simply create a `RowIndex` as https://github.com/h2oai/datatable/blob/main/src/core/expr/head_func_shift.cc#L64

samukweku: @oleksiyskononenko updated. is this what you have in mind?

oleksiyskononenko: Good job, @samukweku. Yes, we're pretty close now. Let's just fix the tests, because they seems to use the old `direction` variable instead of the `reverse` one. Once the tests are fixed, we will see the real issues if any.

samukweku: all done. I need your expertise on this @oleksiyskononenko ... I am getting segmentation fault - it has to do with memory access but I can't figure it out, specifically lines 117 - 128. I am trying to implement fillna if a value is supplied; I am using the if else implementation for this.

oleksiyskononenko: I suggest we start with `reverse` parameter only and once this PR is green, we merge it. Then, we gradually add other options to `fillna()`. For the moment, it seems that after you fixed the tests the PR is still red, meaning that we have some issues with even the initial implementation.

samukweku: @oleksiyskononenko done

oleksiyskononenko: Thanks, one test is still failing though

```
FAILED tests/test-docs.py::test_documentation_h - AssertionError: Variable doc_FExpr_fillna declared in documentation.h not found as a :cvar: among *.rst files
```

oleksiyskononenko:
I've merged `main` onto this PR, because we have some changes in the building pipeline.

src/core/expr/fexpr_fillna.cc
samukweku: All done; speed looks ok; however, there are some scenarios where `by` is quite slow compared to pandas and affects the outcome - in fact based on a naive observation, most of the time is spent in the grouping. I'm hoping we can have a look at the by internals sometime in the future and see if there are any low hanging fruits than can help improve the speed. #2551.       #2541

oleksiyskononenko: Thanks, let me review the PR then.

tests/dt/test-fillna.py
```diff
@@ -0,0 +1,141 @@
...
+
+def test_fillna_grouped_column():
+     DT = dt.Frame([2, 1, None, 1, 2])
+     DT_bfill = DT[:, [fillna(f[0], reverse = False), fillna(f[0], reverse = True)], by(f[0])]
+     DT_ref = dt.Frame([
+                 [None, 1, 1, 2, 2],
+                 [None, 1, 1, 2, 2],
+                 [None, 1, 1, 2, 2]
+              ])
+     assert_equals(DT_bfill, DT_ref)
```
oleksiyskononenko, line 140: When the column is grouped, we could actually have a shortcut: `fillna(col) == col`. An example of how to check if a particular column is grouped: https://github.com/h2oai/datatable/blob/main/src/core/expr/fexpr_qcut.cc#L126-L129

samukweku: Do you mind explaining this more? Is this wrt the tests or the actual function?

oleksiyskononenko: The actual function. We can detect if the column in question is grouped and then  simply return it as a result, instead of running the complicated filling procedure. That’s because the values are constant within the group of the grouped column.

samukweku: Maybe I am missing the definition of grouped. If you don't mind kindly explain. My notion of grouped is `5, 3,NA,4` belong to group 1. I'm probably misunderstanding the grouped column meaning

oleksiyskononenko: What you say is a column from the grouped frame, but the frame is grouped by a different column. 

Look at the above example, the content of the column by which the frame is grouped is always a set of groups with constant values.

samukweku: done. makes sense to possibly add it to other pre existing cumulative functions

oleksiyskononenko (approved):

oleksiyskononenko:
@samukweku I did some changes to this PR, mostly cosmetic. See if they all make sense to you, otherwise I guess we are ready to merge it. Thanks!

src/core/expr/fexpr_fillna.cc
```diff
@@ -0,0 +1,132 @@
...
+        gby = ctx.get_groupby();
+      }
+
+      for (size_t i = 0; i < wf.ncols(); ++i) {
+        Column coli = wf.retrieve_column(i);
+        bool is_grouped = ctx.has_group_column(
+                            wf.get_frame_id(i),
+                            wf.get_column_id(i)
+                          );
+        if (!is_grouped){
```
samukweku, line 102: @oleksiyskononenko is the null count check irrelevant? I thought it would skip if there were no nulls? or does the `is_grouped` boolean somehow cover that?

samukweku:
@oleksiyskononenko thanks for the review; I have a question regarding the null count check which was removed. asides that, it is good to merge. thanks !

src/core/expr/fexpr_fillna.cc
oleksiyskononenko: It would be nice to have this check if it comes for free. However, to actually calculate the number of nulls in the column one needs to loop over the column data in a loop that is kind of similar to the loop which we implement in `fillna()`. Since most of the columns have some missing data, it means that in most of the cases we will have to go through the data twice. My feeling that in this case it is better to remove the check.

oleksiyskononenko: One thing we can actually do here is to check if the stats is already computed and the number of NA's is already known, something similar to https://github.com/h2oai/datatable/blob/main/src/core/column/column_impl.cc#L175

samukweku: I assumed it was already precomputed and available?

oleksiyskononenko: Nope, until touched it is not there and needs to be computed. Let me push an update, so that we reuse it if available.

PR #3313, opened 2022-07-15
[ENH] Column renaming

### LINKED ISSUE #2504: [FR] Option to add a name to grouping in ``by``, especially for boolean expressions
Instead of a default ``C0``, it would be nice to have some relevant name

## Example:
```

             'Gender': ['F', 'M', 'F', 'M', 'F', 'M', 'F', 'M', 'M', 'M'],
                       'algebra', 'stats', 'stats', 'algebra', 'bio', 'bio'],
             'Grade': grades}

df = dt.Frame(data)
df[:, dt.mean(f.Grade), by(f.ExamYear < 2009)]

   | C0 | Grade
---+----+---------
 0 |  0 | 60.6667
 1 |  1 | 70.8571
```
## Suggested form:
```
df[:, dt.mean(f.Grade), by(name = f.ExamYear < 2009)]
```

samukweku:
- Closes #2684
- closes #2504 

Implementation for column renaming

```py

             'Gender': ['F', 'M', 'F', 'M', 'F', 'M', 'F', 'M', 'M', 'M'],
                       'algebra', 'stats', 'stats', 'algebra', 'bio', 'bio'],
             'Grade': grades}

df = dt.Frame(data)

# proposal via this PR
df[:, dt.mean(f.Grade), by((f.ExamYear < 2009).alias('grp'))]

   |   grp    Grade
   | bool8  float64
-- + -----  -------
 0 |     0  60.6667
 1 |     1  70.8571
[2 rows x 2 columns]
```


samukweku:
@oleksiyskononenko early stages on this PR; the idea is formalised; however, your review will be helpful and guide me in the right direction. 

I have also noticed some issues regarding the methods, which I'll bring up in this PR after your review

oleksiyskononenko:
Do you see any benefits renaming `as()` to `alias()`? It seems to me that `alias` is an an _alternative_ name, while here we create a _new_ name, so `as()` could be more appropriate.

src/core/expr/fexpr_alias.cc
```diff
@@ -0,0 +1,122 @@
+//------------------------------------------------------------------------------
+// Copyright 2019-2022 H2O.ai
```
oleksiyskononenko, line 2:
```suggestion
// Copyright 2022 H2O.ai
```

tests/dt/test-alias.py
```diff
@@ -0,0 +1,116 @@
...
+    DT = dt.Frame()
+    msg = "The number of arguments\\(1\\) passed to names does not match "
+    msg = msg + "the number of specified columns\\(0\\)."
+    with pytest.raises(ValueError, match = msg):
+         DT[:, alias(DT, 'C0')]
+
+
+# #-------------------------------------------------------------------------------
+# # Normal
+# #-------------------------------------------------------------------------------
```
oleksiyskononenko, line 79:
```suggestion
#-------------------------------------------------------------------------------
# Normal
#-------------------------------------------------------------------------------
```

src/core/expr/fexpr_alias.cc
```diff
@@ -0,0 +1,122 @@
...
+      }
+      auto gmode = wf.get_grouping_mode();
+      Workframe outputs(ctx);
+
+      for (size_t i = 0; i < wf.ncols(); ++i) {
+        Workframe base_frame(ctx);
+        base_frame.add_column(wf.retrieve_column(i),
+                              std::string(),
+                              gmode);
+        base_frame.rename(names_[i]);
```
oleksiyskononenko, line 76: My feeling is that `.rename()` should be able to rename a set of columns to different names. Right now this method only accepts one name for all columns and in some cases it actually adds a prefix...

samukweku:
> Do you see any benefits renaming `as()` to `alias()`? It seems to me that `alias` is an an _alternative_ name, while here we create a _new_ name, so `as()` could be more appropriate.

@oleksiyskononenko `as` is a python keyword, hence my issue. Alternatively, if we could just make `as` a method, we can possibly avoid this issue. How to do that though is beyond me (I made too many errors; so I took the easy way out). 

As always, I am open to learning how to do things, so pls feel free to guide me

oleksiyskononenko:
Yeah, you're right. If we introduce a function `dt.as()` then we could potentially have issues when doing `from datatable import as`. So here we can only support `.as()` as an f-method.

At the same time we already have a couple of functions with the same names as the python's built-ins: `dt.min()`, `dt.max()` and `dt.sum()`. However, to make them work properly we had to add wrappers that would decide which function (dt or python) to call:
https://github.com/h2oai/datatable/blob/main/src/datatable/expr/reduce.py#L113-L142

Unfortunately, I don't think it is possible to have such a wrapper for a python's keyword.

samukweku:
If we can implement it as a method only on `f`, similar to extend and remove, I feel we can avoid the keyword issue. Unfortunately I am not knowledgable enough to implement that

samukweku:
By the way, @oleksiyskononenko is there a difference between FExpr and Expr. I always thought they were the same, but it seems there is an Expr object and FExpr

samukweku:
Also, the `alias` idea I borrowed from pyspark and Polaris libraries

oleksiyskononenko:
@samukweku No worries, we will figure out how to move forward with this PR. Let's first finalize #3310 and #3311, then we will come back and implement a proper column renaming.

samukweku:
bungled this, closing this and creating a new one : #3333

PR #3328, opened 2022-08-04
Improve test coverage on AppVeyor

oleksiyskononenko:
In this PR we adjust AppVeyor builds to 
- enable `pyarrow` tests;
- on Windows, enable C++ tests by testing debug wheels for Python 3.9;
- on Windows, fix builds to properly report failures;
- for consistency, rename `DTTEST` to `DT_TEST` and namespace `dttest` to `dt::tests`.

Note, that, when enabled, the C++ tests [redefine](https://github.com/h2oai/datatable/blob/main/src/core/utils/tests.h#L91-L100) `protected` and `private` keywords to `public`. This is a pretty dangerous approach, that we might need to reconsider, because this redefinition only happens in the files which include `utils/tests.h`. On Windows, for instance, this caused a pile of linking errors due to the fact that some methods were expected to be `public`, but were declared as `protected` or `private`.

### DIFF
--- a/ci/appveyor.yml
+++ b/ci/appveyor.yml
@@ -216,8 +216,6 @@ build_script:
 
     $DT_WHEEL = ls dist/*$env:DT_BUILD_SUFFIX.debug-cp38-*.whl
 
-    echo "DT_WHEEL = $DT_WHEEL"
-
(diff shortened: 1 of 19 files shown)


oleksiyskononenko:
Merging as failures are not related to this PR.

PR #3334, opened 2022-08-11
Add shortcuts for grouped columns in `cummin()` and `cummax()`

### LINKED ISSUE #3330: skip computation for grouped column
- related to the cumulative functions #3279 , with the idea from @oleksiyskononenko https://github.com/h2oai/datatable/pull/3311#discussion_r939595404
- grouped columns should be skipped since there is really nothing to do there
- cummax()
- cummin()

oleksiyskononenko:
Closes #3330

### DIFF
--- a/src/core/expr/fexpr_cumminmax.cc
+++ b/src/core/expr/fexpr_cumminmax.cc
@@ -53,15 +53,29 @@ class FExpr_CumMinMax : public FExpr_Func {
 
     Workframe evaluate_n(EvalContext& ctx) const override {
       Workframe wf = arg_->evaluate_n(ctx);
-      Groupby gby = Groupby::single_group(wf.nrows());
+      Groupby gby = ctx.get_groupby();
 
-      if (ctx.has_groupby()) {
+      if (!gby) {
+        gby = Groupby::single_group(wf.nrows());
+      } else {
         wf.increase_grouping_mode(Grouping::GtoALL);
-        gby = ctx.get_groupby();
       }
 
       for (size_t i = 0; i < wf.ncols(); ++i) {
-        Column coli = evaluate1(wf.retrieve_column(i), gby);
+        Column coli = wf.retrieve_column(i);
+        auto typei = coli.type();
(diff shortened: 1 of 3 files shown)


src/core/expr/fexpr_cumminmax.cc
```diff
@@ -53,15 +53,29 @@ class FExpr_CumMinMax : public FExpr_Func {
...
+      } else {
         wf.increase_grouping_mode(Grouping::GtoALL);
-        gby = ctx.get_groupby();
       }
 
       for (size_t i = 0; i < wf.ncols(); ++i) {
-        Column coli = evaluate1(wf.retrieve_column(i), gby);
+        Column coli = wf.retrieve_column(i);
+        auto typei = coli.type();
+        if (!(typei.is_numeric_or_void() || typei.is_boolean() ||
```
samukweku, line 67: Since we can compare strings, can't we allow strings here?

oleksiyskononenko: Actually, it would be nice to have some consistency here. For instance,
```
>>> DT = dt.Frame(["a", "b"])
>>> DT.min()
   | C0   
   | str32
-- + -----
 0 | NA   
[1 row x 1 column]
>>> DT[:, f.C0.min()]
TypeError: Unable to apply reduce function min() to a column of type str32
```

So currently we don't support min/max for string columns in any way. We even have some inconsistency (by design), between the Frame methods and the f-methods.

In theory, we could support `cummin()` and `cummax()` for string column, since we can sort those columns. However, I would not change the behavior until we change the regular `min()` and `max()`. It also depends on the community needs, if this support is really something people would like to have.

From what I see now pandas, for instance, supports min/max for strings, but numpy doesn't. So it is kind of arguable if this is something we want to support.

samukweku (approved):

src/core/expr/fexpr_cumminmax.cc
oleksiyskononenko: Would be nice to convert this conversation to discussions, but it seems there is no way.

Issue #3309, opened 2022-07-06
`na.rm` or something similar for aggregations


samukweku:
Not sure if there is an existing ticket, so creating this. Adding a `na.rm`(similar to R) or `skipna` (pandas flavour) parameter in aggregations might be useful

I observed this in functions like `first`:

```py
from datatable import dt, f, by, sort
from numpy import nan

DT = dt.Frame([{'A': 1, 'B': 'a', 'C': 5.0},
 {'A': 1, 'B': 'a', 'C': nan},
 {'A': 2, 'B': None, 'C': 4.0},
 {'A': 2, 'B': 'b', 'C': 4.0},
 {'A': 2, 'B': 'b', 'C': nan},
 {'A': 3, 'B': 'c', 'C': 9.0},
 {'A': 3, 'B': None, 'C': nan},
 {'A': 3, 'B': 'd', 'C': nan},
 {'A': 3, 'B': None, 'C': 9.0},
 {'A': 4, 'B': 'e', 'C': 8.0},
 {'A': 4, 'B': 'e', 'C': nan},
 {'A': 5, 'B': None, 'C': 2.0},
 {'A': 6, 'B': 'h', 'C': nan},
 {'A': 6, 'B': 'h', 'C': nan}])

DT
   |     A  B            C
   | int64  str32  float64
-- + -----  -----  -------
 0 |     1  a            5
 1 |     1  a           NA
 2 |     2  NA           4
 3 |     2  b            4
 4 |     2  b           NA
 5 |     3  c            9
 6 |     3  NA          NA
 7 |     3  d           NA
 8 |     3  NA           9
 9 |     4  e            8
10 |     4  e           NA
11 |     5  NA           2
12 |     6  h           NA
13 |     6  h           NA
[14 rows x 3 columns]

# returns the exact first value, even if it is a null
DT[:, f.B.first(), 'A']

   |     A  B
   | int64  str32
-- + -----  -----
 0 |     1  a
 1 |     2  NA
 2 |     3  c
 3 |     4  e
 4 |     5  NA
 5 |     6  h
[6 rows x 2 columns]

# a workaround is with sort
# this might be limiting in scenarios where you do not need the sort
# or you have other columns and the sorting affects your output/analysis
DT[:, f.B.first(), 'A', sort(f.B, na_position='last')]
   |     A  B
   | int64  str32
-- + -----  -----
 0 |     1  a
 1 |     2  b
 2 |     3  c
 3 |     4  e
[...]

PR #3346, opened 2022-09-03
[ENH] `nth` function

### LINKED ISSUE #3128: nth function/method for groupby

```

        'y': [1, 2, 3, 1, 2, 1],
        'n': [3, 2, 1, 1, 2, 1]}

DT = dt.Frame(data)

DT

   |     x      y      n
   | int32  int32  int32
-- + -----  -----  -----
 0 |     1      1      3
 1 |     1      2      2
 2 |     1      3      1
 3 |     2      1      1
 4 |     2      2      2
 5 |     3      1      1
[6 rows x 3 columns]
```

```

    |   x    sum      nth
    | int32  int64  int32
 -- + -----  -----  -----
  0 |     1      6      2
  1 |     2      3      2
  2 |     3      1     NA
 [3 rows x 3 columns]

```

                force=True)``

samukweku:
Implement `dt.nth(cols, n=0)` function to return the nth row (also per group) for the specified columns. If `n` goes out of bounds, NA-row is returned. 

Closes #3128


samukweku:
@oleksiyskononenko MVP implementation of the `nth` function, similar to pandas' nth function/dplyr's nth function. Feedback appreciated before adding docs

oleksiyskononenko:
I wonder if you want to replace `first()`/`last()` with `nth()`? Like `nth(0)` would mean `first()` and `nth(-1)` would mean `last()`?

samukweku:
@oleksiyskononenko not a bad idea, as it is more generic.

oleksiyskononenko:
So if you look at how the `first()` / `last()` functions work, you will see that it is a purely virtual column and not even a rowindex on the source column: https://github.com/h2oai/datatable/blob/main/src/core/expr/head_reduce_unary.cc#L112-L165

`FirstLast_ColumnImpl::get_element()` just gets the first or last element from a group. What you need to do in the case of the `nth(n=...)` function, is to see how your `n` compares to the group size and return `NA` if it is out of bounds.


In the case the column is grouped, `first()` / `last()` immediately return `NA` column for zero-rows frame, or the source column (note, the first or last elements from each group from the grouped column is actually the grouped column itself).

So for the `nth(n=...)` function: `n` could only be `0` or `-1`, otherwise, `NA` column should be returned.

samukweku:
@oleksiyskononenko what's the disadvantage of using a rowindex for this function? Performance? Or something else?

oleksiyskononenko:
Yeah, current implementation uses as much memory as needed for `Buffer buf = Buffer::mem(gby.size() * sizeof(int32_t));`. When you do it virtual as for `first()` / `last()`, you won't really need additional memory.

samukweku:
Thanks @oleksiyskononenko

samukweku:
@oleksiyskononenko made changes to the `nth` function, based on feedback. I also added a `skipna` parameter, to get non null values. 

When you have some time, kindly have a look at the PR; your feedback is always appreciated.

samukweku:
I made a blunder on this and should have just done a git pull 🤦

src/core/expr/fexpr.cc
```diff
@@ -442,6 +442,20 @@ DECLARE_METHOD(&PyFExpr::min)
     ->docs(dt::doc_FExpr_min);
 
 
+oobj PyFExpr::nth(const XArgs& args) {
+  auto nthFn = oobj::import("datatable", "nth", "skipna");
```
oleksiyskononenko, line 446: I'm not sure what means imporing `skipna`. You only need to import `nth` from `datatable`.

src/core/expr/fexpr_nth.cc
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+
+    template <typename T>
+    Column make(Column&& col, int32_t n, const Groupby& gby) const {
+      if (SKIPNA){
+        return Column(new Latent_ColumnImpl(new NthSkipNA_ColumnImpl<T>(std::move(col), n, gby)));
+      } else {
+        return Column(new NTH_ColumnImpl<T>(std::move(col), n, gby));
+      }
```
oleksiyskononenko, line 79: Why you need a separate implementation for the skippna case? You can have a `SKIPNA` template parameter and based on its value you could call either a normal `nth()` implementation, or the one that skips missings.

I don't think we save any resources if `NthSkipNA_ColumnImpl` creates a new data column on the first access. Yes, it needs to iterate through the data, but for each group you have to go through the group independently until you find the nth valid element.

samukweku: @oleksiyskononenko so I understand better, are you saying the iteration can happen within the Virtualcolumn? within the `get_element` function? I thought that would be comparatively slow, because of the iteration

oleksiyskononenko: Yeah, the iteration will take some time, but it takes time in the current implementation also.

The only reason we implemented cumsum, cumprod, etc in this way was the benefit of calculating all the elements at once as, for instance, `cumsum(i + 1) = x(i + 1) + cumsum(i)`. 

For the `nth()` function there is no benefit to know nth for the previous group. The only benefit could be to save all the nth while iterating, but then we will use a lot of memory and there is no guarantee the user will ever need other nth's.

samukweku: thanks for the explanation @oleksiyskononenko

src/core/expr/fexpr.cc
samukweku: how do I pull in the `skipna` parameter into the `f` method?

src/core/expr/fexpr_nth.cc
samukweku: updated the code; looking forward to your review and feedback before adding the docs. Also, I need your help on how to implement the `get_element` for strings. Any reason why strings don't work with the current implementation?

src/core/expr/fexpr_nth.cc
```diff
@@ -64,19 +63,15 @@ class FExpr_Nth : public FExpr_Func {
         case SType::INT64:   return make<int64_t>(std::move(col), n, gby);
         case SType::FLOAT32: return make<float>(std::move(col), n, gby);
         case SType::FLOAT64: return make<double>(std::move(col), n, gby);
-        //case SType::STR32:   return make<CString>(std::move(col), n, gby);
+        //case SType::STR32:   return make<CString>(std::move(col), n, gby); -- any ideas on how to make this work @oleksiyskononenko
```
oleksiyskononenko, line 66: What exactly doesn't work here?

oleksiyskononenko:
Thanks @samukweku I will take a look. 

Btw, if you see PR to be red, go to AppVeyor to check why it fails and try to fix. AppVeyor is publicly accessible, so that you can see the exact cause of the failure. For instance here something is still wrong with the `skipna` parameter: https://ci.appveyor.com/project/h2oops/datatable/builds/44837594/job/emhlqqeofy5lwsch#L5037

It is always nice to have PR green or know why it fails, otherwise I'm not sure what part of your PR you're having issues with.

src/core/expr/fexpr_nth.cc
samukweku: the `make` function throws an error for string functions, and I can't figure out the error for that

src/core/expr/fexpr_nth.cc
oleksiyskononenko: It seems that the strings do work now?

samukweku: Yes @oleksiyskononenko it works now 🤷

src/core/column/nth.h
```diff
@@ -0,0 +1,82 @@
...
+#define dt_NTH_h
+#include "column/virtual.h"
+#include "stype.h"
+
+namespace dt {
+template<typename T, bool SKIPNA>
+class NTH_ColumnImpl : public Virtual_ColumnImpl {
+  private:
+    Column col_;
+    int32_t nth_;
```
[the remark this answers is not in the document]
samukweku: @oleksiyskononenko any particular reason to use `n`, instead of `nth_`?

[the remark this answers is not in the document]
oleksiyskononenko: Well, we say “give me the n-th element when n equals to something”, right? So the function name should be “nth()” and the parameter name should be “n” everywhere. This is what pandas does, for instance: https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.GroupBy.nth.html

src/core/column/nth.h
```diff
@@ -0,0 +1,82 @@
...
+      size_t i0, i1;
+      gby_.get_group(i, &i0, &i1);
+      xassert(i0 < i1);
+      pos = n_<0 ? static_cast<size_t>(n_)+i1
+                 : static_cast<size_t>(n_)+i0;
+      bool isvalid = false;
+      if (pos >= i0 & pos< i1){
+        for (size_t ii = pos; ii < i1; ++ii) {
+          isvalid = col_.get_element(ii, out);
+          if (isvalid || !SKIPNA) {break;}
```
oleksiyskononenko, line 67: In this case the `SKIPNA` value will be checked for every iteration of the loop until the valid element is found. Compiler will probably optimize the code and remove `|| 0` for `SKIPNA == 0`. 

Anyways, It is more efficient and clear to have two separate paths: one for `SKIPNA == 0` and another for `SKIPNA == 1`. So once you test your `pos` to be inside the group bounds, then you do
```
if (SKIPNA) {

// Your loop to find the first non-missing element

} else {

// Simply return the nth element no matter if it is valid or not

}
```

src/core/column/nth.h
```diff
@@ -0,0 +1,82 @@
...
+    size_t n_children() const noexcept override {
+      return 1;
+    }
+
+    const Column& child(size_t i) const override {
+      xassert(i == 0);  (void)i;
+      return col_;
+    }
+
+    bool get_element(size_t i, T* out) const override{
```
oleksiyskononenko, line 56: Having a space before the opening brackets (here and in many other places on this PR) will help to make the code more readable.

src/core/column/nth.h
```diff
@@ -0,0 +1,82 @@
...
+      return col_;
+    }
+
+    bool get_element(size_t i, T* out) const override{ 
+      size_t pos;          
+      size_t i0, i1;
+      gby_.get_group(i, &i0, &i1);
+      xassert(i0 < i1);
+      pos = n_<0 ? static_cast<size_t>(n_)+i1
+                 : static_cast<size_t>(n_)+i0;
```
oleksiyskononenko, line 62: Spaces around arithmetic operations like `+`, will also help.

src/core/column/nth.h
```diff
@@ -0,0 +1,82 @@
...
+      if (pos >= i0 & pos< i1){
+        for (size_t ii = pos; ii < i1; ++ii) {
+          isvalid = col_.get_element(ii, out);
+          if (isvalid || !SKIPNA) {break;}
+        }
+      }
+      if (isvalid){
+        return true;
+      }
+      else {*out = GETNA<T>();}
```
oleksiyskononenko, line 73: When you `return false` later, it doesn't matter what the content of `out` is. The element returned is not valid anyways. Why not just remove this whole `if-else` and simply `return isvalid`?

oleksiyskononenko:
Btw, we also need to add padding otherwise the code generates too many warnings: https://ci.appveyor.com/project/h2oops/datatable/builds/44859288/job/4llvi83iew79r0f5?fullLog=true#L405

samukweku:
@oleksiyskononenko how do I go about the padding?

src/core/expr/fexpr_nth.cc
```diff
@@ -44,7 +44,7 @@ class FExpr_Nth : public FExpr_Func {
         for (size_t i = 0; i < wf.ncols(); ++i) {
           Column coli = wf.retrieve_column(i);
           coli = evaluate1(std::move(coli), n_, gby);
-          outputs.add_column(std::move(coli), std::string(), Grouping::GtoONE);
+          outputs.add_column(std::move(coli), std::string(wf.retrieve_name(i)), Grouping::GtoONE);
```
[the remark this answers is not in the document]
samukweku: @oleksiyskononenko the grouping is GtoONE, unlike the others that are GtoALL, so I expect a size mismatch error ?

oleksiyskononenko:
@samukweku You need to pad the `gby_` property with 4 bytes. Just add `size_t : 32;` after that.

src/core/expr/fexpr_nth.cc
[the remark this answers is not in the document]
oleksiyskononenko: >  // GtoONE
  //   Each group is mapped to exactly 1 row. This grouping mode is
  //   common as a result of "reduce" operations such as `sum`, `sd`,
  //   `min`, etc. A column with this level may need to be expanded
  //   in order to become conformable with any full-sized column.

So it's fine to have this grouping mode and it may be expanded if needed.

src/core/column/nth.h
```diff
@@ -33,7 +33,7 @@ class NTH_ColumnImpl : public Virtual_ColumnImpl {
     Groupby gby_;
 
   public:
-    NTH_ColumnImpl(Column&& col, int32_t n,  const Groupby& gby)
+    NTH_ColumnImpl (Column&& col, int32_t n,  const Groupby& gby)
```
oleksiyskononenko, line 36: Space between the method name and a bracket is not really needed, because it is natural to write `f(x)` with no spaces. Just take a look at the rest of dt code.

src/core/expr/fexpr.cc
```diff
@@ -489,7 +489,7 @@ DECLARE_METHOD(&PyFExpr::min)
 
 oobj PyFExpr::nth(const XArgs& args) {
   auto nthFn = oobj::import("datatable", "nth");
-  oobj n = args[0].to_oobj();
+  oobj n = args[0].to_oobj() ? args[0].to_oobj() : py::oint(0);
```
oleksiyskononenko, line 492: Also, space between the condition and `?` for ternary operators is not really needed. Just like in the normal sentence you wouldn't put the space between a word and `?`.

src/core/expr/fexpr_nth.cc
```diff
@@ -16,7 +16,7 @@ class FExpr_Nth : public FExpr_Func {
     int32_t n_;
 
   public:
-    FExpr_Nth(ptrExpr&& arg, py::oobj n)
+    FExpr_Nth (ptrExpr&& arg, py::oobj n)
```
oleksiyskononenko, line 19: Only curly brackets should be space padded, see the rest of dt code for a reference.

src/core/expr/fexpr_nth.cc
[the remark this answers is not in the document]
samukweku: ok, I'm not sure how to implement that. do you mind showing me how?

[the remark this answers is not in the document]
oleksiyskononenko: Just implement it in exactly the same way as you did for `cumsum`, `cummin`, and all the other cumulative functions.

oleksiyskononenko:
I guess we should also add tests to check that in the case when `skipna == False`

- for `n >= 0`: `dt.nth(f[:], n)` is equivalent to `dt.first(dt.shift(f[:], -n))`;
- for `n < 0`: `dt.nth(f[:], n)` is equivalent to `dt.last(dt.shift(f[:], n + 1))`.

Obviously, performance of `dt.nth()` should be better than the one of the equivalent chained functions.

oleksiyskononenko:
I did some refactoring (keeping the way the output `Workframe` is created by adding new columns). What is left here is documentation and also a couple of tests to compare `nth()` vs `first()`/`last()`/`shift()`.

src/core/expr/fexpr_nth.cc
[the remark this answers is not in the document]
oleksiyskononenko: I've been looking at this and let's keep it in the way it works now.

src/core/column/nth.h
```diff
@@ -23,54 +23,65 @@
...
+      size_t ni = (n_ >= 0)? static_cast<size_t>(n_) + i0
+                           : static_cast<size_t>(n_) + i1;
+
       bool isvalid = false;
-      if (pos >= i0 && pos< i1){
+      if (ni >= i0 && ni < i1){
         if (SKIPNA){
-          for (size_t ii = pos; ii < i1; ++ii) {
+          size_t ii = ni;
+          while (ii < i1 && !isvalid) {
```
samukweku, line 79: Part of my learning, what is the reason for choosing while instead of for here?

src/core/column/nth.h
```diff
@@ -23,54 +23,65 @@
...
+
+
 template<typename T, bool SKIPNA>
-class NTH_ColumnImpl : public Virtual_ColumnImpl {
+class Nth_ColumnImpl : public Virtual_ColumnImpl {
   private:
     Column col_;
-    int32_t n_;
     Groupby gby_;
+    int32_t n_;
```
samukweku, line 34: Any particular reason for the positional argument change between gby and n? Again, part of me learning

src/core/column/nth.h
oleksiyskononenko: Because when you need to break the loop based on some condition it is more natural to use the while loop, rather than to write `if (condition) break;` inside the for loop.

Btw, my feeling is that the logic of `skipna` is wrong in this implementation. For pandas it is 
>Apply the specified dropna operation before counting which row is the nth row. Only supported if n is an int.

In this implementation what we do is return the first non-missing after the n-th row. However, we don't remove missings until the n-th row.

```python
>>> DT = dt.Frame([None, 1, None])
>>> DT[:, dt.nth(f[:], n=0, skipna=True)]
   |   C0
   | int8
-- + ----
 0 |    1
[1 row x 1 column]
>>> DT[:, dt.nth(f[:], n=1, skipna=True)]
   |   C0
   | int8
-- + ----
 0 |    1
[1 row x 1 column]
>>> DT[:, dt.nth(f[:], n=2, skipna=True)]
```

We need to change the logic.

src/core/column/nth.h
oleksiyskononenko: Just refactoring. Since groupby is related to the column, it is better to keep these two together. Also, in this case the padding (that is related to the usage ot `int32_t`) goes immediately after the `n_` declaration. If, in future, we add any other parameters at the end, we will easily see how much padding we should add.

oleksiyskononenko:
Regarding the `skipna` parameter. I guess the logic on this PR is wrong: https://github.com/h2oai/datatable/pull/3346#discussion_r980954267 It could be fixed, but the thing is that currently each column is treated separately — so when we detect missings we also do it separately. As a result, the `n` calculated could be a different value for each column.

src/core/column/nth.h
samukweku: Which logic sounds better? I do not have any preferences

oleksiyskononenko:
Basically, with the current implementation it is not possible to implement `skipna` as `all` or `any`. Because each column doesn't know what happens with the another. I propose we remove `skipna` parameter for the moment, because the same functionality could be achieved by doing `rowall()` / `rowany()`, filtering only the needed rows, then using `nth()`.

oleksiyskononenko:
There is also another issue with the current implementation

```python
DT = dt.Frame([0, 1, 0])
DT[:, dt.nth(f.C0, 0), by(f.C0)]
```

Results in
```python
AssertionError: Assertion 'i < nrows()' failed in src/core/column.cc, line 222
```

samukweku:
My thought process for skipna is not related to pandas nth completely. If the nth position is a null, skip till the next non null value. Dropping all nulls or any nulls doesn't really make sense to me. I'm not sure we can replicate with rowall/rowany. `first`/'last` has a similar [issue](#3309), which I feel skipna takes care of

samukweku:
Thanks for pointing the missing test; I'll look into it. By the way, should we extend nth to include lists/slices, or make that as a separate issue with a more explicit function name?

oleksiyskononenko:
@samukweku When we say `skipna` it feels like we want to skip na's by counting the nth row. Skipping na's starting only from the nth row doesn't feel intuitive to me. Imagine the following example:

```python
>>> DT = dt.Frame([[None, 1, 2], [None, None, "two"]])
>>> DT
   |    C0  C1   
   | int32  str32
-- + -----  -----
 0 |    NA  NA   
 1 |     1  NA   
 2 |     2  two  
[3 rows x 2 columns]
>>> DT[:, nth(f[:], skipna=True)]
   |    C0  C1   
   | int32  str32
-- + -----  -----
 0 |     1  two  
[1 row x 2 columns]
```

In this case it feels like `nth(f[:])` should return the nth row not counting the missing values, however, the result is coming from different rows and values go out of sync.

samukweku:
@oleksiyskononenko that looks fine as it is the nth row per column. If the user wants a uniform output on a row basis, then the `i` section should be used instead, in my opinion

oleksiyskononenko:
A row of data normally means a record or an observation, but I'm not sure what is the meaning of the mix like in the example above. What pandas does with `dropna` looks more reasonable, because then `nth()` of one column is logically consistent with `nth()` for multiple columns.

P.S. Please check your messages on gitter.

samukweku:
lets say the user wants the first non null value per column, this function does that, pandas covers that with `first_valid_index`. I feel if the user wants consistency row wise, then use the `I` section, although that can be limiting too. pandas nth is only for groups


have a look at [this](https://stackoverflow.com/questions/73282239/retrieve-first-non-na-value-based-on-another-time-variable from SO, which skipna as currently implemented would solve easily.  our reduction functions are generally per column, which makes sense.

At any rate, I agree that we remove skipna until we agree on the logic(feedback from users would be more helpful)

oleksiyskononenko:
Well, if we implement it the pandas way and you need the nth non-null per a column you can always do something like this: `DT[:, [nth(f.col1), nth(f.col2), …]]`.

But if we implement it as it is now, there will be no way to get the nth all/any non-missing row.

samukweku:
@oleksiyskononenko if you don't mind, can you show me how to apply row_all or row_any within evaluate_n?

src/core/expr/fexpr_nth.cc
```diff
@@ -106,20 +107,21 @@ class FExpr_Nth : public FExpr_Func {
...
-
-  if (skipna) {
-    return PyFExpr::make(new FExpr_Nth<true>(as_fexpr(cols), n));
-  } else {
-    return PyFExpr::make(new FExpr_Nth<false>(as_fexpr(cols), n));
+  auto dropna = args[2].to<std::string>("None");
+  bool dropna_check = dropna == "any" || dropna == "all" || dropna == "None";
+  if (!dropna_check) {
+    throw ValueError() << "Parameter `dropna` in nth() should be "
+        "either `None`, `any`, or `all`";
```
oleksiyskononenko, line 121: For `None` we don't need it to be a string. `None` could be a Python's `None`. We can also keep the name as `skipna`. Not sure why pandas has both `skipna` and `dropna` for the same purpose.

oleksiyskononenko:
@samukweku Yeah, if we want to apply `rowall()`/`rowany()` we first need to convert columns to booleans based on their validity, then either pass the result to `Nth_ColumnImpl` to use a the validity mask or convert it to a rowindex applying to workframe.

samukweku:
@oleksiyskononenko any ideas on how to fix the rowall implementation?

oleksiyskononenko:
So do you want to implement (1) or (2) from https://github.com/h2oai/datatable/pull/3346#issuecomment-1262885185 ?

samukweku:
@oleksiyskononenko At any rate, I'm open to either - whichever you feel will be more efficicent

samukweku:
@oleksiyskononenko no pressure here; still hoping for your guidance code wise on this

oleksiyskononenko:
@samukweku I've been thinking about this PR and guess we should first do the following:

1. see how the current implementation of `nth()` stands vs `first()`/`last()`/`shift()`, see https://github.com/h2oai/datatable/pull/3346#issuecomment-1259007371 My feeling is that both memory and performance-wise it is going to be almost identical;
2. check memory/performance of `first()`/`last()`/`shift()` combined with `rowall()`/`rowany()` for emulating the `skipna` parameter. My feeling is that it could still be ok. 

If my feeling is correct, then there will be no need to a separate `nth()` implementation and in the case (1) it could just be an alias for the chained `first()`/`last()`/`shift()`, while in the case (2) it could just add step of `rowall()`/`rowany()`.

samukweku:
all good @oleksiyskononenko i'll have a look

samukweku:
@oleksiyskononenko the shift combined with first/last works great for skipna=False. For a groupby however, since the dropna is applied per column (referencing Pandas), the rowany/rowall doesnt fit. Probably have to iterate through each group to remove all nulls, or any

oleksiyskononenko:
>dropna is applied per column

Why is it applied per column? Look at the example below
```python
>>> import pandas as pd
>>> PD = pd.DataFrame([[1,2,None,1.1],[3,None,4,3.3],[5,6,7,7.7],[8,None,None,8.8]])
>>> PD
   0    1    2    3
0  1  2.0  NaN  1.1
1  3  NaN  4.0  3.3
2  5  6.0  7.0  7.7
3  8  NaN  NaN  8.8
>>> PD.groupby(0).nth(0, dropna='all')
     1    2    3
0               
1  2.0  NaN  1.1
3  NaN  4.0  3.3
5  6.0  7.0  7.7
8  NaN  NaN  8.8
>>> PD.groupby(0).nth(0, dropna='any')
     1    2    3
0               
1  NaN  NaN  NaN
3  NaN  NaN  NaN
5  6.0  7.0  7.7
8  NaN  NaN  NaN
```

As you see rows returned by `nth()` are either rows from the original frame, or fully missing rows. If `dropna` is applied per column, the column values would go out of sync.

samukweku:
Thanks @oleksiyskononenko totally misunderstood the implementation. I'll look at it some more and see if I can update the PR

oleksiyskononenko:
What I try to understand here is the advantage of implementing the `nth()` function. My feeling is that the same functionality could be achieved by combining first/last/shift/row* functions. Obviously, it is nice to have the `nth()`, but since we implement it in C++, we need to make sure it matches first/last/shift/row* combination. Same is true for `fillna()`.

oleksiyskononenko:
@samukweku I don't think `nth()` takes into account the grouped column. Even with no `dropna` parameter, `nth()` gives you an empty frame on the grouped columns:
```
>>> PD.iloc[:, :1].groupby(0).nth(0)
Empty DataFrame
Columns: []
Index: [1, 3, 5, 8]
```

samukweku:
at the moment, we can close this PR and stick with the current implementations (first/last/shift) ... yes, nth would offer a nice abstraction, but the current implementation is fine. for fillna,  adding fill with a value is just a nice to have, and makes null filling in a single location. 

yea, i'll close this ticket, probably add documentation/examples on how to replicate nth, and look at the other issues instead

samukweku:
by the way @oleksiyskononenko what would be the easy way to convert the reducer functions (and others possibly) to FExpr objects, without rewriting them?

oleksiyskononenko:
Well, I think this implementation is almost finished, we just need to add `dropna` support. I'm not against having a separate `nth()` function, because the combination of first/last/shift/row* could be not obvious to users. The same regarding `fillna()`, it is a pretty useful shortcut.

oleksiyskononenko:
@samukweku Unfortunately, I guess you do need to rewrite them. They're not using the new `FExpr` class.

samukweku:
@oleksiyskononenko I have no idea how to use the `FExpr_RowAll` function. I need your help with this

samukweku:
@oleksiyskononenko the skipna part as you said is the one that's left. I can add tests for `first/last/shift` results to match current implementation. The rowall/rowany is what I can't code unfortunately, which is why I need your help on it

samukweku:
@oleksiyskononenko thanks for the reviews so far. Still looking forward to your input on this one - thanks

oleksiyskononenko:
Sure, I will take a look at this PR too.

src/core/expr/fexpr_nth.cc
```diff
@@ -0,0 +1,167 @@
...
+        case SType::INT32:   return Column(new Isna_ColumnImpl<int32_t>(std::move(col)));
+        case SType::TIME64:
+        case SType::INT64:   return Column(new Isna_ColumnImpl<int64_t>(std::move(col)));
+        case SType::FLOAT32: return Column(new Isna_ColumnImpl<float>(std::move(col)));
+        case SType::FLOAT64: return Column(new Isna_ColumnImpl<double>(std::move(col)));
+        case SType::STR32:
+        case SType::STR64:  return Column(new Isna_ColumnImpl<CString>(std::move(col)));
+        default: throw RuntimeError();
+      }
+    }
```
oleksiyskononenko, line 126: So what is the purpose of the `evaluate2` method and why `evaluate1` is commented out?

samukweku: kindly ignore that - I was experimenting with the `FExpr_RowAll` function, to see how it works

samukweku:
just could not resolve the conflicts - moved to #3403

PR #3348, opened 2022-09-09
Fix cumulative functions when invoked on the latent columns

### LINKED ISSUE #3345: Cumulative functions crash when applied to some reducers
- Did you find a bug in datatable, or maybe the bug found you?
Method chaining fails in some cases - I have noticed it for reducing vs non-reducing aggregations.

- How to reproduce the bug?
```py
from datatable import dt, f
DT = dt.Frame({"A": [1, 1, 2, 1, 2],
               "B": [None, 2, 3, 4, None],
               "C":[True, False, False, True, True]})
DT
   |     A      B      C
   | int32  int32  bool8
-- + -----  -----  -----
 0 |     1     NA      1
 1 |     1      2      0
 2 |     2      3      0
 3 |     1      4      1
 4 |     2     NA      1
[5 rows x 3 columns]


# this works
DT[:, f[:].cumsum().sum(), 'A']
Out[159]:
   |     A      B      C
   | int32  int64  int64
-- + -----  -----  -----
 0 |     1      8      4
 1 |     2      6      1
[2 rows x 3 columns]

# this fails
In [161]: DT[:, f[:].sum().cumsum(), 'A']
AttributeError                            Traceback (most recent call last)
Input In [161], in <cell line: 1>()
----> 1 DT[:, f[:].sum().cumsum(), 'A']

In [160]: DT[:, f[:].sum(), 'A']
Out[160]:
   |     A      B      C
   | int32  int64  int64
-- + -----  -----  -----
 0 |     1      6      2
 1 |     2      3      1
[2 rows x 3 columns]

In [163]: DT[:, dt.cumsum(f[:].sum()), 'A']
[3]    50175 segmentation fault  ipython
```

  In case it is not obvious, please tell us what result should your code
  produce.

- Your environment?
  datatable version - 1.1.0
python version - 3.9.12
operating system - mac monterrey

oleksiyskononenko:
Cumulative functions acces the column's data in parallel. If the input column appears to be a `Latent_ColumnImpl`, materialization of such a column happens in parallel. However, the `Latent_ColumnImpl::materialize()` is not a thread-safe method and invoking it from multiple threads results in a segfault.

In this PR we `vivify()` the source column, i.e. materialization, if necessary, happens in one thread only, before processing it in parallel.

Closes #3345

### DIFF
--- a/src/core/column/cumminmax.h
+++ b/src/core/column/cumminmax.h
@@ -21,6 +21,7 @@
 //------------------------------------------------------------------------------
 #ifndef dt_COLUMN_CUMMINMAX_h
 #define dt_COLUMN_CUMMINMAX_h
+#include "column/latent.h"
 #include "column/virtual.h"
 #include "parallel/api.h"
 #include "stype.h"
@@ -45,10 +46,12 @@ class CumMinMax_ColumnImpl : public Virtual_ColumnImpl {
 
 
     void materialize(Column& col_out, bool) override {
+      Latent_ColumnImpl::vivify<T>(col_);
+
       Column col = Column::new_data_column(col_.nrows(), col_.stype());
       auto data = static_cast<T*>(col.get_data_editable());
-
       auto offsets = gby_.offsets_r();
+
       dt::parallel_for_dynamic(
         gby_.size(),
         [&](size_t gi) {
--- a/src/core/column/cumsumprod.h
+++ b/src/core/column/cumsumprod.h
@@ -21,6 +21,7 @@
 //------------------------------------------------------------------------------
 #ifndef dt_COLUMN_CUMSUM_h
 #define dt_COLUMN_CUMSUM_h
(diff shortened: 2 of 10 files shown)


samukweku (approved):

samukweku:
@oleksiyskononenko What does vivify do? How does it help to manage the thread safe issue?

oleksiyskononenko:
`vivify()` just accesses the zeroth element in a column, and that, in its turn, triggers materialization for latent columns (they are designed to be materialized on the first access). Since we put `vivify()` in the single-threaded part of the code, materialization will be triggered from one thread only.

The issue we had was related to the fact that we tried to access latent column's data from the parallel region and that actually triggered materialization from multiple threads. However, the latent column's materialization routine is not thread-safe.

samukweku:
Thanks @oleksiyskononenko

PR #3370, opened 2022-10-14
Add basic support for `Grouping::GtoFEW`

oleksiyskononenko:
It seems that `dt.categories()` is the first datatable function, that needs to produce `Grouping::GtoFEW` columns. That's because the number of categories in a categorical column could be anything between `0` and `nrows - 1`. Currently, datatable doesn't really support `Grouping::GtoFEW`, but may need it for the cases when `dt.categories()` is combined with other f-expressions, or when `dt.categories()` is applied to columns that have different number of underlying categories.

In this PR we
- add some basic support for `Grouping::GtoFEW` grouping mode;
- adjust `dt.categories()` to produce `Grouping::GtoFEW` columns, that in the case of uneven number of rows are  promoted to `Grouping::GtoALL`;
- do minor refactoring in `dt.alias()` function.

WIP for #1691

### DIFF
--- a/src/core/expr/declarations.h
+++ b/src/core/expr/declarations.h
@@ -75,7 +75,7 @@ namespace expr {
   //   Each group is mapped to exactly `groupsize` rows. This is the
   //   most common grouping mode. Any simple column, or a function of
   //   a simple column will be using this mode. Few groupby functions
-  //   may use this mode too
+  //   may use this mode too.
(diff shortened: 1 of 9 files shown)


samukweku (approved):

oleksiyskononenko:
Merging, error is not relevant to this PR.

PR #3371, opened 2022-10-17
Implement `dt.codes()`

oleksiyskononenko:
Implement `dt.codes()` to get integer codes for categorical columns. Since we currently don't support unsigned column types, the internal representation of codes has been also changed to signed integers.

WIP for #1691


samukweku:
@oleksiyskononenko what's the benefit of having access to the codes? It may be worthwhile to add that somewhere

samukweku (approved):

oleksiyskononenko:
@samukweku Well, many of the machine learning methods do not deal with the categorical data directly, but use the corresponding integer features (codes) instead. Here, since we are implementing categorical columns, we should provide full access to the underlying data, i.e. an ability to retrieve the categories and the codes.

PR #3151, opened 2021-08-24
Include gendoc file into the source distribution

st-pasha:
File `gendoc.py` and the accompanying doc files were missing from the source distribution.

Closes #3150

PR #3375, opened 2022-10-28
ci/ext.py: drop explicit `-m64` to fix build on exotic archs

Mrmaxmeier:
nixpkgs builds with this change [here](https://github.com/NixOS/nixpkgs/blob/38164d1660dcc24b41a5a22f1e9ef075a8e26714/pkgs/development/python-modules/datatable/default.nix#L33), but it seems like this has not made it to upstream yet.

### DIFF
--- a/ci/ext.py
+++ b/ci/ext.py
@@ -294,7 +294,6 @@ def build_extension(cmd, verbosity=3):
         # Common link flags
         ext.compiler.add_linker_flag("-shared")
         ext.compiler.add_linker_flag("-g")
-        ext.compiler.add_linker_flag("-m64")
         if macos:
             ext.compiler.add_linker_flag("-undefined", "dynamic_lookup")
         if linux:


oleksiyskononenko:
Thanks for the PR, @Mrmaxmeier. The reason for `-m64` flag is that datatable only supports 64 bit systems. I see that this PR builds with no issues, but not sure if there are cases when a linker could fall back to `-m32`. Do you have any thoughts on this?

Btw, you can upgrade the package in your repo to datatable `1.0.0` and python `3.7+` here: https://github.com/NixOS/nixpkgs/blob/38164d1660dcc24b41a5a22f1e9ef075a8e26714/pkgs/development/python-modules/datatable/default.nix#L14-L15

oleksiyskononenko:
Btw, if you upgrade datatable to `1.0.0` you won't need these dependencies https://github.com/NixOS/nixpkgs/blob/38164d1660dcc24b41a5a22f1e9ef075a8e26714/pkgs/development/python-modules/datatable/default.nix#L41

Mrmaxmeier:
> Thanks for the PR, @Mrmaxmeier. The reason for `-m64` flag is that datatable only supports 64 bit systems. I see that this PR builds with no issues, but not sure if there are cases when some linkers could fall back to `-m32`. Do you have any thoughts on this?

Oh right, seems reasonable given the 64-bit requirement.

> Btw, you can upgrade the package in your repo to datatable `1.0.0` and python `3.7+` here

Updating to 1.0.0 is currently blocked on https://github.com/h2oai/datatable/pull/3151 as the fix has not made it to PyPI yet and nixpkgs tries to build the package wheel from PyPI source tarballs.
I'll close this PR and keep an eye out for potential 1.0.1 or 1.1.0 releases. :+1:

oleksiyskononenko:
I wonder how removing `-m64` helps your project? We have quite a lot of asserts to check for a 64-bit system: https://github.com/h2oai/datatable/blob/main/src/core/datatablemodule.cc#L68-L95 

Btw, you don't need to upgrade to the new datatable version to remove `propagatedBuildInputs = [ typesentry blessed ];`, these dependencies were dropped as of version `0.11.0`.

oleksiyskononenko:
@Mrmaxmeier just to let you know, the `-m64` flag will be dropped on `arm64` only once  https://github.com/h2oai/datatable/pull/3469 is merged.

Issue #3081, opened 2021-07-17
Improve reducers parallelization logic


oleksiyskononenko:
Right now datatable reducers are only parallelized in terms of the groups. When there is only one group, even if it is huge, datatable does all the work in a single-thread. For instance,

```python
from datatable import dt, f
DT = dt.Frame(range(10**9))
print(DT[:, dt.sum(f[0])])
```

will utilize one thread only taking a significant amount of time to complete. We should improve the parallelization logic by taking into account not just the number of groups, but also their sizes.

PR #3388, opened 2022-11-25
[ENH] FExpr for sum/prod

samukweku:
- WIP #2562


samukweku:
@oleksiyskononenko I dont know how to adjust the code for groupby on self; the groupby column(s) are excluded before the aggregation, hence the flawed results.

src/core/expr/fexpr_sumprod.cc
```diff
@@ -0,0 +1,131 @@
...
+#include "expr/workframe.h"
+#include "python/xargs.h"
+#include "stype.h"
+
+namespace dt {
+namespace expr {
+
+
+template <bool SUM>
+class FExpr_SumProd : public FExpr_Func {
```
oleksiyskononenko, line 36: Since `sum()` and `prod()` are both unary functions, I guess they should be better inherited from `FExpr_FuncUnary`.

oleksiyskononenko:
The test that is failing is `test_reduce_sum` and not `test_reduce_sum_same_column`, so I guess the issue is not aggregation on the grouped column, but something else.

I also see another issue, I need to configure AppVeyor to report the failing test more verbosely. Now it just stops with
```
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Windows fatal exception: access violation
```
that is not helpful.

src/core/expr/fexpr_sumprod.cc
samukweku: We use `FExpr_Func` for cumsum and cumprod, that's why I used it here. So single functions should inherit from func_unary? Any suggestions on usage?

oleksiyskononenko: Right. It seems that `FExpr_FuncUnary` is currently not useful for reducers or cumulative functions, because it doesn't take into account the groupby information. We probably need to implement `FExpr_FuncUnaryReducer` (or something similar (not as a part of this PR).

src/core/column/sumprod.h
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+
+    bool get_element(size_t i, T* out) const override {
+      T result;
+      if (SUM){
+        result = 0;
+      } else {
+        result = 1;
+      }
```
oleksiyskononenko, line 52:
```suggestion
      T result = !SUM;      
```

src/core/column/sumprod.h
```diff
@@ -0,0 +1,91 @@
...
+        if (isvalid){
+          if (SUM) {
+            result += value;
+          } else {
+            result *= value;
+          }          
+        }
+      }
+      *out = result;
+      return true;
```
oleksiyskononenko, line 67:
```suggestion
      return true; // the result is never a missing value
```

src/core/expr/fexpr_sumprod.cc
```diff
@@ -0,0 +1,131 @@
...
+            << "Invalid column of type `" << stype << "` in " << repr();
+      }
+    }
+
+
+    template <typename T>
+    Column make(Column &&col, SType stype, const Groupby &gby) const {
+       col.cast_inplace(stype);
+       return Column(new SumProd_ColumnImpl<T, SUM>(std::move(col), gby)
+       );
```
oleksiyskononenko, line 99:
```suggestion
       return Column(new SumProd_ColumnImpl<T, SUM>(std::move(col), gby));
```

src/core/expr/fexpr_sumprod.cc
```diff
@@ -0,0 +1,131 @@
...
+      Groupby gby = ctx.get_groupby();
+      if (!gby) {
+        gby = Groupby::single_group(wf.nrows());
+      }
+      
+      Workframe outputs(ctx);
+      for (size_t i = 0; i < wf.ncols(); ++i) {
+         Column coli = evaluate1(wf.retrieve_column(i), gby);
+         outputs.add_column(std::move(coli), wf.retrieve_name(i), Grouping::GtoONE);
+      }
```
oleksiyskononenko, line 64: So we don't have any shortcuts for the grouped columns?

samukweku: is there a benefit to that? it will be reduced, so the same steps should be involved (`evaluate1`)

src/core/expr/head_reduce_unary.cc
```diff
@@ -183,248 +183,6 @@ static Column compute_gfirstlast(Column&& arg, const Groupby&) {
...
-
-    bool get_element(size_t i, U* out) const override {
-      T value;
-      size_t i0, i1;
-      groupby.get_group(i, &i0, &i1);
-      bool isvalid = arg.get_element(i, &value);
-      *out = isvalid? static_cast<U>(i1 - i0) * static_cast<U>(value)
-                    : U(0);
-      return true; // *out is never an NA
-    }
```
oleksiyskononenko, line 393: Look at this class and this method. This is how the sum could be calculated for a grouped column. You don't need to go through all the elements, because they're all equal, so you just multiply the size of group by only one value.

src/core/expr/head_reduce_unary.cc
```diff
@@ -183,248 +183,6 @@ static Column compute_gfirstlast(Column&& arg, const Groupby&) {
...
```
oleksiyskononenko, line 272: Btw, for `prod()` it seems that we had a bug in the grouped case (because it is just a copy-paste from the grouped sum implementation):
```
>>> DT
   |    C0
   | int32
-- + -----
 0 |     2
 1 |     2
 2 |     2
[3 rows x 1 column]
>>> DT[:, dt.prod(f.C0)]
   |    C0
   | int64
-- + -----
 0 |     8
[1 row x 1 column]
>>> DT[:, dt.prod(f.C0), by(f.C0)]
   |    C0     C1
   | int32  int64
-- + -----  -----
 0 |     2      6
[1 row x 2 columns]
```

src/core/expr/fexpr_sumprod.cc
oleksiyskononenko: Yes, please take a look at my comments below.

src/core/expr/head_reduce_unary.cc
oleksiyskononenko: It should be `8` in the both cases, but in the last case it calculated the sum instead of the product.

oleksiyskononenko: So here we actually supposed to have a shortcut: the value ^ number of elements in the group.

samukweku:
I noticed  a stark difference when calling sum on the Frame, as against an fexpr; is there something I am missing in the implementation. Looked at the code for the stats.cc, and the only thing i could attribute it to is the parallel operation. is it portable to the FExpr implementation? related to #3081 maybe

Screenshot_3

oleksiyskononenko:
Yes, because reducers are parallelized in terms of groups, while the frame method is parallelized in terms of rows (I believe). So when there is only one group, the difference is huge.

Have you tried to compare the existing `dt.sum()` function vs your new `FExpr`? I guess perfomance should be more or less the same.

samukweku:
The example I shared has no groupby involved, so if I follow your line of thought, then the parallelization rowwise should be applicable to the FExpr. Cos the parallelization of the frame method is awesome and I feel we can adopt it into the FExpr implementation. Unless I am misunderstanding the implementation of get_element

src/core/column/sumprod.h
```diff
@@ -0,0 +1,101 @@
...
+      gby_.get_group(i, &i0, &i1);
+      if (is_grouped_){
+        bool isvalid = col_.get_element(i, &value);
+        if (isvalid){
+          if (SUM) {            
+            result = static_cast<T>(i1-i0) * value;
+          } else {
+            for (size_t i = i0; i < i1; ++i) {
+              result *= value;
+            }
```
oleksiyskononenko, line 60: For a grouped column you can avoid the loop too, because the `result` is `value` to the power of `i1 - i0`.

src/core/column/sumprod.h
```diff
@@ -0,0 +1,101 @@
...
+        for (size_t i = i0; i < i1; ++i) {
+          bool isvalid = col_.get_element(i, &value);
+          if (isvalid){
+            if (SUM) {
+              result += value;
+            } else {
+              result *= value;
+            }          
+          }
+        }
```
oleksiyskononenko, line 73: Ok, so you see that this loop is not parallelized at all. So for the case when there is only one group (ungrouped frame), we calculate sum and product in a single thread. While the frame method `.sum()` is fully parallel.

samukweku: would it be ok then to implement the parallelised version for non groups? or it is not worth copying? do we need to point users to the Frame method, if they are looking for fast aggregations on non-grouped operations?

src/core/column/sumprod.h
samukweku: yea, it was easier for me to loop here, I am not sure of how to handle `pow(val, power)`, especially when usually it is set to return`double` type

src/core/column/sumprod.h
oleksiyskononenko: Yes, we need to do something about it as a part of #3081. Here I suggest we only implement what is required for this PR and check that performance of `dt.sum()` is not affected (even if it is not good comparing to `Frame.sum()`.

samukweku: first image is the current implementation in `main`:
Screenshot_7

the second image is the `PR`:

Screenshot_8

src/core/column/sumprod.h
oleksiyskononenko: Yes, unfortunately there is `std::pow()` function for integers, but what we could do here is to implement our own `pow()` function by using [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring). This way we don't need to worry about the parameter type (just make it a template).

src/core/column/sumprod.h
oleksiyskononenko: Yeah, so performance of `dt.sum()` for this PR is comparable to what we currently have in `main`. At the same time, for groupby we are slightly better than pandas, but with no groupby we are worse. Am I correct with interpreting your numbers?

samukweku: Yes you are right. Hopefully we can improve the performance for both grouped and ungrouped. Pypolars is much faster on both counts.

oleksiyskononenko: We should be good in the case when the number of groups is `>=` the number of cores, and those groups are comparable in their sizes. Otherwise, we are very far from being optimal. 

What I wonder though, is why pandas is so good on ungrouped frames... I thought pandas is single-threaded, just like what we do in this case.

samukweku: I think somehow numpy makes the sum computation really fast... How I do not know

samukweku: for prod, datatable groupby is slower than pandas: 
Screenshot_11

src/core/column/sumprod.h
```diff
@@ -55,8 +55,14 @@ namespace dt {
...
-              result *= value;
+            size_t power = i1-i0;
+            while (power) {
+              if (power % 2) {
+                result *= value;
+              }
+              if (!result) break;
+              value *= value;
+              power /= 2;
             }
```
oleksiyskononenko, line 66: I guess this could be extracted as a separate function and placed into `models/utils.cc`. You can make it `inline` though.

Also, I'm not sure if `if (!result) break;` make it faster in average. I understand this is a shortcut for the case when `value` is `0`?

oleksiyskononenko:
Currently, this PR gives quite a lot of warnings that we need to address.

oleksiyskononenko:
Also, I guess you're missing `Latent_ColumnImpl` as a wrapper for sum/prod reducers. Otherwise, the reduced columns stay purely virtual and accessing its elements will lead to costly re-calculations.

src/core/expr/head_reduce_unary.cc
```diff
@@ -183,248 +183,6 @@ static Column compute_gfirstlast(Column&& arg, const Groupby&) {
...
-  *out = sum;
-  return true;  // *out is not NA
-}
-
-
-
-template <typename T, typename U>
-static Column _sum(Column&& arg, const Groupby& gby) {
-  return Column(
-          new Latent_ColumnImpl(
```
oleksiyskononenko, line 330: `Latent_ColumnImpl` is missing in the new implementation.

samukweku: correct me if I'm wrong; I had the notion that Latent_ColumnImpl is called when running parallelisation, and as such data would only be calculated when called. The virtual however didnt need it

src/core/column/sumprod.h
samukweku: was not aware of `models/utils.cc` . I'll check it out. 

You are right that the  `if(!result) break` is a shortuct for when value is 0; i'll remove it

src/core/expr/head_reduce_unary.cc
oleksiyskononenko: `Latent_ColumnImpl` is materialized on the first data access, right not your implementation is virtual, i.e. the data are never materialized and calculated every time from scratch

src/core/column/sumprod.h
oleksiyskononenko: `model/utils` is just a good place to put a new function to, then we could re-use it if needed. 

The issue with `if (!result) break` is that it will be calculated every loop iteration for non-zero elements also. If you want to have a shortcut: you can check `value` outside of the loop. But I don't really know how much you will save.

oleksiyskononenko:
I fixed all the warnings and did some re-factoring. In particular, you had a mismatch in `ipow()` signature and the call: different order of the `base` and `exp` parameters that made it wrong. Since it was not caught by the tests, I guess we need to add test for `prod()` on the grouped columns.

src/datatable/expr/reduce.py
```diff
@@ -110,13 +106,6 @@ def corr(col1, col2):
     return Expr(OpCodes.CORR, (col1, col2))
 
 
-# noinspection PyShadowingBuiltins
-def sum(iterable, start=0):
-    if isinstance(iterable, (Expr, core.FExpr)):
-        return Expr(OpCodes.SUM, (iterable,))
-    else:
-        return _builtin_sum(iterable, start)
-
```
oleksiyskononenko, line 119: This function allowed to use the python `sum()` along with the datatable one. Now, as you removed it, the python sum (in some cases) could be shadowed by the datatable one. For instance, if now someone does `from datatable import *`, the built in `sum()` will be overwritten by datatable version.

src/core/column/sumprod.h
```diff
@@ -31,55 +31,51 @@ namespace dt {
...
+    bool is_grouped_;
+    size_t : 56;
 
   public:
-    SumProd_ColumnImpl(Column &&col, bool is_grouped, const Groupby &gby)
-      :Virtual_ColumnImpl(gby.size(), col.stype()),
-       col_(std::move(col)),
-       is_grouped_(is_grouped),
-       gby_(gby)
+    SumProd_ColumnImpl(Column &&col, Groupby &&gby, bool is_grouped)
```
samukweku: @oleksiyskononenko what's the idea behind const vs && in this case?

src/core/column/sumprod.h
```diff
@@ -31,55 +31,51 @@ namespace dt {
...
       } 
-      *out = result;     
+      *out = result;
       return true; // the result is never a missing value
     }
 
 
     ColumnImpl *clone() const override {
-      return new SumProd_ColumnImpl(Column(col_), is_grouped_, gby_);
+      return new SumProd_ColumnImpl(Column(col_), Groupby(gby_), is_grouped_);
```
samukweku, line 78: does order matter? or is it just for ease of reading? I am referring to `is_grouped` comeing after `Groupby` in the function definition

src/core/column/sumprod.h
oleksiyskononenko: If you look at the way we are getting the groupby object from `ctx`, i.e.
```python
Groupby gby = ctx.get_groupby();
```
you will notice, that we essentially create a copy of that object. Unfortunately, we can't do
```python
const Groupby& gby = ctx.get_groupby();
```
to avoid copying, because later we may need to assign a single group to `gby`.

Actually, creating a copy is ok in this case, because we need to store that groupby object safely in `SumProd_ColumnImpl`. However, we are passing this copy around as `const Groupby& gby`, so when it comes to `SumProd_ColumnImpl` we have to create its own copy of the object. To prevent this additional copy from being created, it would be nice to just `std::move()` the groupby object around. 

I found that `fexpr_sumprod.cc` is not the only place where we may need to make this change, so it would be nice to have a separate PR with the fix.

oleksiyskononenko: Also, when we do reduction for several columns at the same time, we do need as many groupby copies as the number of the columns, so we can't simply `std::move()` the initial groupby. Anyways, I've reverted these changes as this issue needs more understanding.

src/datatable/expr/reduce.py
samukweku: @oleksiyskononenko any suggestions on this?

oleksiyskononenko: Just do not remove this shadowing prevention and use `core.sum()` instead of the old reducer function name. Take a look at the changes I just pushed.

oleksiyskononenko (approved):

src/datatable/expr/reduce.py
samukweku: thanks @oleksiyskononenko

PR #3396, opened 2022-12-12
Add `dt.prod()` test for a grouped column, `by()` docs adjustment

### LINKED ISSUE #3390: `dt.prod()` calculates sum instead of a product for the grouped columns
```python
>>> DT = dt.Frame([1, 1])
>>> DT[:, dt.prod(f.C0)]
   |    C0
   | int64
-- + -----
 0 |     1
[1 row x 1 column]
>>> DT[:, dt.prod(f.C0), by(f.C0)]
   |   C0     C1
   | int8  int64
-- + ----  -----
 0 |    1      2
[1 row x 2 columns]
```

Note, the product on the grouped column should still be `1`, i.e. the correct result for the last example is
```python
   |   C0     C1
   | int8  int64
-- + ----  -----
 0 |    1      1
[1 row x 2 columns]
```

samukweku:
- the fix for #3390 has been already pushed as a part of #3388, here we merely add a corresponding test;
- as of #2472, `f[:]` excludes the groupby columns, in this PR we make the corresponding adjustments to the docs. 

Closes #3390

### DIFF
--- a/tests/test-groups.py
+++ b/tests/test-groups.py
@@ -300,6 +300,14 @@ def test_reduce_prod():
     assert f1.to_list() == [["blue", "green", "red"],
                             [2, -7, 65]]
 
+def test_reduce_prod_same_column():
+    # See issue #3390
+    f0 = dt.Frame({"ints" : [0, 1, 0, 0, 1, 2]})
+    f1 = f0[:, {"prod" : prod(f.ints)}, f.ints]
+    frame_integrity_check(f1)
+    assert_equals(f1, dt.Frame({"ints" : [0, 1, 2], "prod" : [0, 1, 2]/dt.int64}))
+                        
+
 
 #-------------------------------------------------------------------------------
 # Groupby on large datasets
--- a/tests/test-groups.py
+++ b/tests/test-groups.py
@@ -302,14 +302,6 @@ def test_reduce_prod():
 
 def test_reduce_prod_same_column():
     # See issue #3390
-    f0 = dt.Frame({"ints" : [0, 1, 0, 0, 1, 2]})
-    f1 = f0[:, {"prod" : prod(f.ints)}, f.ints]
-    frame_integrity_check(f1)
-    assert_equals(f1, dt.Frame({"ints" : [0, 1, 2], "prod" : [0, 1, 2]/dt.int64}))
-
-
-def test_reduce_prod_same_column1():
-    # See issue #3390
     f0 = dt.Frame({"ints" : [0, 1, 2, 2, 1, 2]})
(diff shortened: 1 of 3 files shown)


oleksiyskononenko:
So it looks that we already had a test in `test_reduce.py` that should caught the bug
```
def test_prod_grouped():
    DT = dt.Frame(A=[True, False, True, True], B=[None, None, None, 10], C=[2,3,5,0.1])
    RES = DT[:, prod(f[:]), by(f.A)]
    REF = dt.Frame(A=[False, True], B=[1, 10]/dt.int64, C=[3,1.0]/dt.float64)
    frame_integrity_check(RES)
    assert_equals(RES, REF)
    assert str(RES)
```

However, `prod(f[:])` actually calculated products for `B` and `C` columns only...

tests/test-groups.py
```diff
@@ -300,6 +300,22 @@ def test_reduce_prod():
...
+    frame_integrity_check(f1)
+    assert_equals(f1, dt.Frame({"ints" : [0, 1, 2], "prod" : [0, 1, 2]/dt.int64}))
+
+
+def test_reduce_prod_same_column1():
+    # See issue #3390
+    f0 = dt.Frame({"ints" : [0, 1, 2, 2, 1, 2]})
+    f1 = f0[:, {"prod" : prod(f.ints)}, f.ints]
+    frame_integrity_check(f1)
+    assert_equals(f1, dt.Frame({"ints" : [0, 1, 2], "prod" : [0, 1, 8]/dt.int64}))
```
oleksiyskononenko, line 316: I guess this test is enough, as the one above only tests a very simple case when `prod(ints) == ints`.

samukweku:
For groupby , `f[:]` excludes the grouping column - the user would have to explicitly add the grouping column to the `j` section

oleksiyskononenko:
Yes, so I am thinking if we ever documented this behavior.

samukweku:
Was handled here : #2472 ... Not sure if it was documented in the groupby docs

samukweku:
Yea, I will update the docs to reflect that

samukweku:
``
If
      ``j`` is a "select-all" slice (i.e. ``:``), then those columns will
      also be excluded from the list of all columns so that they will be
      present in the output only once.
``

The above seems to cover this and can be found here : https://datatable.readthedocs.io/en/latest/api/dt/by.html

oleksiyskononenko:
Anyways, don't worry about that, let me just push a minor commit to this PR and we're all set.

oleksiyskononenko (approved):

oleksiyskononenko:
I made some minor adjustments to this PR, see if it looks good to you.

PR #3402, opened 2023-01-02
Refactor `mean()` reducer to use `FExpr`

samukweku:
- refactor `mean()` reducer to use `FExpr`;
- fix `median()` to use `float64` for void columns.

WIP for #2562

### DIFF
--- a/src/core/column/mean.h
+++ b/src/core/column/mean.h
@@ -46,7 +46,6 @@ class Mean_ColumnImpl : public Virtual_ColumnImpl {
 
 
     bool get_element(size_t i, T* out) const override {
-      
       T value;
       size_t i0, i1;
       gby_.get_group(i, &i0, &i1);
--- a/src/core/column/mean.h
+++ b/src/core/column/mean.h
@@ -1,0 +1,98 @@
+//------------------------------------------------------------------------------
+// Copyright 2022 H2O.ai
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#ifndef dt_COLUMN_MEAN_h
+#define dt_COLUMN_MEAN_h
+#include "column/virtual.h"
+#include "stype.h"
+namespace dt {
+
+
+template <typename T>
+class Mean_ColumnImpl : public Virtual_ColumnImpl {
+  private:
+    Column col_;
+    Groupby gby_;
+    bool is_grouped_;
+    size_t : 56;
+
+  public:
+    Mean_ColumnImpl(Column &&col, const Groupby& gby, bool is_grouped)
+      : Virtual_ColumnImpl(gby.size(), col.stype()),
+        col_(std::move(col)),
+        gby_(gby),
+        is_grouped_(is_grouped)
+    {
+      xassert(col_.can_be_read_as<T>());
+    }
+
+
+    bool get_element(size_t i, T* out) const override {
+      
+      T value;
+      size_t i0, i1;
+      gby_.get_group(i, &i0, &i1);
+
+      if (is_grouped_){
+        bool is_valid = col_.get_element(i, &value);
+        if (!is_valid) return false;
+        *out = static_cast<T>(value);
+        return true;
+      } else {
+          double sum = 0;
+          int64_t count = 0;
+          for (size_t gi = i0; gi < i1; ++gi) {
+            bool is_valid = col_.get_element(gi, &value);
+            if (is_valid) {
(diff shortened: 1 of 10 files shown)


samukweku:
@oleksiyskononenko I do not understand the rules for `void` when computing the mean. If you have some time, I'd love your feedback on it

src/core/expr/fexpr_mean.cc
```diff
@@ -0,0 +1,126 @@
...
+        gby = Groupby::single_group(wf.nrows());
+      } 
+
+      for (size_t i = 0; i < wf.ncols(); ++i) {
+      bool is_grouped = ctx.has_group_column(
+                        wf.get_frame_id(i),
+                        wf.get_column_id(i)
+                      );
+      Column coli = evaluate1(wf.retrieve_column(i), gby, is_grouped);        
+      outputs.add_column(std::move(coli), wf.retrieve_name(i), Grouping::GtoONE);
```
oleksiyskononenko, line 68:
```suggestion
        bool is_grouped = ctx.has_group_column(
                            wf.get_frame_id(i),
                            wf.get_column_id(i)
                          );
        Column coli = evaluate1(wf.retrieve_column(i), gby, is_grouped);        
        outputs.add_column(std::move(coli), wf.retrieve_name(i), Grouping::GtoONE);         
```

src/core/column/mean.h
```diff
@@ -0,0 +1,98 @@
...
+        col_(std::move(col)),
+        gby_(gby),
+        is_grouped_(is_grouped)
+    {
+      xassert(col_.can_be_read_as<T>());
+    }
+
+
+    bool get_element(size_t i, T* out) const override {
+
```
oleksiyskononenko, line 49:
```suggestion
```

src/core/expr/fexpr_mean.cc
```diff
@@ -0,0 +1,126 @@
...
+      switch (stype) {
+        case SType::VOID: {
+          return Column(new ConstNa_ColumnImpl(gby.size(), stype));
+        }
+          
+        case SType::BOOL:
+        case SType::INT8:
+        case SType::INT16:
+        case SType::INT32:        
+        case SType::INT64:
```
oleksiyskononenko, line 87: I don't think you can just fall through here. You can't `get_element()` from the int columns to `double`. See how a similar `switch` was implemented before.

src/core/expr/fexpr_mean.cc
```diff
@@ -0,0 +1,126 @@
...
+        case SType::FLOAT64:
+          return make<double>(std::move(col), SType::FLOAT64, gby, is_grouped);
+        case SType::TIME64: {
+          Column coli = make<double>(std::move(col), SType::FLOAT64, gby, is_grouped);
+          coli.cast_inplace(SType::TIME64);
+          return coli;
+        }
```
oleksiyskononenko, line 94: We can probably do a similar trick to `DATE32`. Also, if we support datetimes, we need to adjust documentation — it now says we only support numerics for `mean()`.

oleksiyskononenko:
@samukweku So what is your question on voids?

src/core/expr/fexpr_mean.cc
```diff
@@ -0,0 +1,126 @@
...
+      return outputs;
+    }
+
+
+    Column evaluate1(Column &&col, const Groupby& gby, bool is_grouped) const {
+      SType stype = col.stype();
+
+      switch (stype) {
+        case SType::VOID: {
+          return Column(new ConstNa_ColumnImpl(gby.size(), stype));
```
oleksiyskononenko, line 80:
```suggestion
          return Column(new ConstNa_ColumnImpl(gby.size(), SType::FLOAT64));
```

oleksiyskononenko: According to documentation
>The column stypes are float32 for float32 columns, and float64 for all the other numeric types.

That's the reason you need to return `SType::FLOAT64` column, and not `SType::VOID`.

src/core/expr/fexpr_mean.cc
samukweku:
```
using U = typename std::conditional<std::is_same<T, float>::value,
                                      float, double>::type;
```

this is the switch condition used. not familiar with the construct - my interpretation is that if the type is float32, then the output is float32, else the output should be float64. that's why i implemented it as such. open to learning

oleksiyskononenko: Yes, you're correct with respect to the return type. I didn't notice you're casting in-place in the `make()` function. It's all ok, I guess.

src/core/expr/fexpr_mean.cc
```diff
@@ -64,20 +66,23 @@ class FExpr_Mean : public FExpr_Func {
...
+    Column evaluate1(Column &&col, const Groupby& gby, bool has_by, bool is_grouped) const {
       SType stype = col.stype();
 
       switch (stype) {
         case SType::VOID: {
-          return Column(new ConstNa_ColumnImpl(gby.size(), SType::FLOAT64));
+          std::cout<<has_by<<std::endl;
+          Column coli = has_by? Column(new ConstNa_ColumnImpl(gby.size(), SType::FLOAT64))
+                              : Column(new ConstNa_ColumnImpl(gby.size()));
+          return coli;
```
oleksiyskononenko, line 85: This seems to be some debug changes?

src/core/expr/fexpr_mean.cc
```diff
@@ -64,20 +66,23 @@ class FExpr_Mean : public FExpr_Func {
                             wf.get_frame_id(i),
                             wf.get_column_id(i)
                           );
-        Column coli = evaluate1(wf.retrieve_column(i), gby, is_grouped);        
+        Column coli = evaluate1(wf.retrieve_column(i), gby, has_by, is_grouped);
```
oleksiyskononenko, line 68: So why do we need to have `has_by`? The code was just fine before this commit, it was only the stype that was wrong in several tests.

samukweku: I guess I misunderstood then. So all void columns should return float64, whether or not a groupby is present?

oleksiyskononenko: Yeah, there is no reason to distinguish.

oleksiyskononenko:
Is there anything else you're planning to do on this PR? Not sure why the build is failing.

samukweku:
@oleksiyskononenko Just ur review and merge is what's left

oleksiyskononenko:
The build is read, so we first need to make it green before merging. Can you re-trigger the build to see if it was a temp glitch on AppVeyor?

samukweku:
Sure, will do

oleksiyskononenko:
@samukweku once I merge https://github.com/h2oai/datatable/pull/3412, please remove your changes to `test_median*` test and merge main onto your PR. `test_median*` tests have typos and testing `mean()` instead of `median()`.

oleksiyskononenko:
Nevermind, I've already made the required fixes. See if you have anything else to be done for this PR, otherwise, I guess it is all good to be merged.

samukweku:
Thanks @oleksiyskononenko all good from me. Ok to merge

oleksiyskononenko (approved):

PR #3403, opened 2023-01-03
[ENH] `nth` function

### LINKED ISSUE #3128: nth function/method for groupby

samukweku:
Implement `dt.nth(cols, n)` function to return the nth row (also per group) for the specified columns. If `n` goes out of bounds, NA-row is returned. 

Closes #3128


samukweku:
not sure how `median` is related to `nth` - triggering an `Address boundary error`

oleksiyskononenko:
@samukweku let's separate `first()` and `last()` re-factoring from the new `nth()` feature. Also, could you please restore `skipna` parameter, so that I help you to make it work (got time for that)?

PR #3419, opened 2023-02-01
[ENH] Implement comparison methods

samukweku:
- WIP #2946 
- Create methods for comparisions (>, <, ==, ...)

```py
DT[:, f.A.eq(B) & f.C.ne(f.D)]
```
equivalent to 

```py
DT[:, (f.A==f.B) & (f.C!=f.D)]
```


oleksiyskononenko:
Do we have any idea as to how useful these could be? I don't see any mention of the comparison methods on https://github.com/h2oai/datatable/issues/2946.

samukweku:
I'll add it to #2946 ... It's more of convenience/aesthetics especially when chaining multiple calls

oleksiyskononenko:
If you feel they're needed we can add them of course. I'm jus wondering if anyone is going to use those.

PR #3425, opened 2023-02-14
Fix rendering documentation for functions with optional parameters

### LINKED ISSUE #3424: Rendering documentation for functions with optional parameters
Optional parameters in a function signature are normally specified inside the square brackets. For instance, 
as
```python
:signature: __getitem__(self, i, j[, by][, sort][, join])
```
in the [rst file](https://github.com/h2oai/datatable/blob/main/docs/api/frame/__getitem__.rst).

However, when the rst file is parsed into html by our documentation system, the signature [becomes](https://datatable.readthedocs.io/en/latest/api/frame/__getitem__.html)
```python
self[i, j, [by], [sort], [join]]
```

and commas are formatted as though they're not optional, that is kind of confusing. We need to fix the way we're generating html for functions with optional parameters and make it consistent with the rst signatures.

oleksiyskononenko:
Closes #3424

### DIFF
--- a/docs/_ext/xfunction.py
+++ b/docs/_ext/xfunction.py
@@ -1015,21 +1015,32 @@ class XobjectDirective(SphinxDirective):
                                      classes=["sig-open-paren"])
             params = xnodes.div(classes=["sig-parameters"])
             last_i = len(self.parsed_params) - 1
+
+            # `comma` will be rendered in the beginning on the corresponding
+            # parameter, because for an optional `param` it has to go inside
+            # the square brackets.
+            comma = None
+
             for i, param in enumerate(self.parsed_params):
                 classes = ["param"]
-                comma = None
+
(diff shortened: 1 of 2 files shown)


oleksiyskononenko:
@samukweku If you have time, after docs has been built for this PR, please check if function signatures are rendered in the same way as on readthedocs. Except, for the ones with optional parameters that have been fixed. Thanks!

samukweku (approved):

samukweku:
nws @oleksiyskononenko

samukweku:
@oleksiyskononenko  previewed locally and it looks same as on readthedocs

image

oleksiyskononenko:
Yeah, I was more worried about the rest of the signatures, but they look fine. If you spot any issues in future, please let me know.

PR #3433, opened 2023-03-01
Replace spin-lock with conditional variable for `SleepTask`

sh1ng:
Using spin-lock when threads are waiting could sometimes result in hangs on systems with heavy load. In this PR we replace spin-lock based `SleepTask` with the one based on the standard conditional variable. The performance could be slightly better or worse comparing to what we currently have, however, this change assures no hangs will happen.


src/core/parallel/job_idle.cc
```diff
@@ -26,6 +26,9 @@
 #include "parallel/api.h"
 #include "parallel/job_idle.h"
 #include "parallel/thread_pool.h"
+#include <condition_variable>
+#include <mutex>
+#include <iostream>
```
oleksiyskononenko, line 31:
```suggestion
```

st-pasha (approved):
Looks legit!

src/core/parallel/job_idle.cc
```diff
@@ -149,15 +152,19 @@ SleepTask::SleepTask(Job_Idle* idle_job)
...
+  std::unique_lock<std::mutex> lk(cv_m_);
+  cv_.wait(lk,[&]{return job_ != nullptr;});
   xassert(job_);
   thpool->assign_job_to_current_thread(job_);
 }
 
 
 void SleepTask::wake_up(int nth, ThreadJob* next_job) {
-  job_ = next_job;
-  semaphore_.signal(nth);
```
oleksiyskononenko, line 160: `nth` can now be removed from `wake_up()` parameters, since it is not used anymore

sh1ng:
LGTM, let's merge it.

oleksiyskononenko:
I added a small test locally and found a glitch, looking now as to how to fix it.

oleksiyskononenko:
I did some additional benchmarks and it appears that in many cases conditional variable based approach outperforms the spin-lock. Don't want to overcomplicate it, so reverted it back to what you had. It is better to have out-of-the-box datatable to have no unwanted hangs.

Issue #3417, opened 2023-01-31
Incorrect output for nested reducers


samukweku:
Nesting aggregation calls sometimes produces incorrect results

```py
>>> from datatable import dt, f
>>> DT = dt.Frame([1, 2])
>>> DT[:, dt.mean(dt.sum(f.C0))]
   |         C0
   |    float64
-- + ----------
 0 | 2.8823e+18
[1 row x 1 column]
>>> DT[:, dt.sum(dt.mean(f.C0))]
   |          C0
   |     float64
-- + -----------
 0 | 2.31584e+77
[1 row x 1 column]
```

The expected behavior is
```py
>>> DT[:, dt.mean(dt.sum(f.C0))]
   |    C0
   | float64
-- + -----
 0 |     3
[1 row x 1 column]
>>> DT[:, dt.sum(dt.mean(f.C0))]
   |      C0
   | float64
-- + -------
 0 |     1.5
[1 row x 1 column]
```

datatable version: 1.1
python version: 3.9
operating system: linux

samukweku:
Another example - this time for `countna`: 

```py
DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])
DT
   |     G      V
   | int32  int32
-- + -----  -----
 0 |     1     NA
 1 |     1     NA
 2 |     1     NA
 3 |     2     NA
 4 |     2      3
 5 |     2      5
[6 rows x 2 columns]

DT[:, dt.mean(f.V), dt.by(f.G)] # wrong output
   |     G        V
   | int32  float64
-- + -----  -------
 0 |     1       NA
 1 |     2        4
[2 rows x 2 columns]

# expected output
   |     G    V1
   | int32  int8
-- + -----  ----
 0 |     1     1
 1 |     2     0
[2 rows x 2 columns]


#intermediate frame:
DT[:, dt.mean(f.V), dt.by(f.G)]
   |     G        V
   | int32  float64
-- + -----  -------
 0 |     1       NA
 1 |     2        4
[2 rows x 2 columns]


```

samukweku:
the `countna` issue is solved with #3440

oleksiyskononenko:
I guess your example is missing `countna()` function.

oleksiyskononenko:
@samukweku You say `countna()` is wrong, but in your example you use `count()`. The expected result is, at the same time, provided for `countna()`. Could you please check it one more time?

oleksiyskononenko:
[...]

PR #3440, opened 2023-03-10
[ENH] Refactor count/countna to use FExpr

### LINKED ISSUE #3441: `dt.countna()` returns wrong results for grouped columns
```python
>>> DT = dt.Frame([None])
>>> DT[:, dt.countna(f.C0), dt.by(f.C0)]
   |   C0     C1
   | void  int64
-- + ----  -----
 0 |   NA      0
[1 row x 2 columns]
```

The result should actually be
```python
   |   C0     C1
   | void  int64
-- + ----  -----
 0 |   NA      1
[1 row x 2 columns]
```

samukweku:
- convert `count()` and `countna()` to `FExpr`;
- fix `countna()` when applied to grouped columns;
- change`ReduceUnary_ColumnImpl` for more type flexibility.

WIP for #2562 
Closes #3441


src/core/column/countna.h
```diff
@@ -0,0 +1,64 @@
+//------------------------------------------------------------------------------
+// Copyright 2022 H2O.ai
```
oleksiyskononenko, line 2:
```suggestion
// Copyright 2023 H2O.ai
```

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,152 @@
+//------------------------------------------------------------------------------
+// Copyright 2022 H2O.ai
```
oleksiyskononenko, line 2:
```suggestion
// Copyright 2023 H2O.ai
```

src/core/expr/head_reduce_unary.cc
```diff
@@ -308,170 +308,170 @@ static Column compute_gsd(Column&& arg, const Groupby& gby) {
...
-    case SType::TIME64:
-    case SType::INT64:   return _gcount<int64_t,NA>(std::move(arg), gby);
-    case SType::FLOAT32: return _gcount<float,NA>(std::move(arg), gby);
-    case SType::FLOAT64: return _gcount<double,NA>(std::move(arg), gby);
-    case SType::STR32:
-    case SType::STR64:   return _gcount<CString,NA>(std::move(arg), gby);
-    default: throw _error("count", arg.stype());
-  }
-}
+// //------------------------------------------------------------------------------
```
oleksiyskononenko, line 311: The commented part of the code could safely be removed.

src/core/column/countna.h
```diff
@@ -0,0 +1,64 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#ifndef dt_COLUMN_COUNTNA_h
+#define dt_COLUMN_COUNTNA_h
+#include "column/reduce_unary.h"
+namespace dt {
+
+
+template <typename T, typename U, bool COUNT, bool COUNTT, bool IS_GROUPED>
```
oleksiyskononenko, line 28: Not sure I understand the name `COUNTT`.

samukweku:
@oleksiyskononenko Also need your help in diagnosing why the tests fail in appveyor but work locally?

src/core/column/countna.h
```diff
@@ -0,0 +1,64 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#ifndef dt_COLUMN_COUNTNA_h
+#define dt_COLUMN_COUNTNA_h
+#include "column/reduce_unary.h"
+namespace dt {
+
+
+template <typename T, typename U, bool COUNT_NOT_NULL, bool COUNT_ALL_ROWS, bool IS_GROUPED>
```
oleksiyskononenko, line 28: We only have two functions that are implemented here, i.e. `count()` or `countna()`, why do we need two booleans to handle that? If we introduce one boolean `COUNTNA`, then
- for `COUNTNA == 0` we implement `count()` logic;
- for `COUNTNA == 1` we implement `countna()` logic.

src/core/column/countna.h
```diff
@@ -0,0 +1,64 @@
...
+class CountNA_ColumnImpl : public ReduceUnary_ColumnImpl<T, U, IS_GROUPED> {
```
oleksiyskononenko, line 29: And the class could be just named as `Count_ColumnImpl`, then `COUNTNA` will indicate we want to calculate missing values only.

src/core/column/countna.h
samukweku: One Boolean is to get count of rows for valid elements, the second Boolean gets count of all rows and does not consider element validity. I renamed the booleans for more clarity. Open to suggestions on renaming and better implementation

oleksiyskononenko: >the second Boolean gets count of all rows and does not consider element validity

Does it mean you want to implement a new functionality here? Because currently `count()` only calculates number of non-missing elements, while `countna()` returns the number of missings.

oleksiyskononenko: I'm actually wondering what is the best way to return a number of rows per group when using `by()`. It seems that one has to do `count() + countna()` that is not really good. Is it something you wanted to address?

oleksiyskononenko: We probably need to update docs https://datatable.readthedocs.io/en/latest/api/dt/count.html and make parameter `cols` to be optional.

samukweku: The implementation takes care of the internal `count0()` - this section of the code in `fexpr_count_countna.cc` takes care of that: 

```py
    Workframe evaluate_n(EvalContext &ctx) const override {
      Workframe outputs(ctx);
      Workframe wf = arg_->evaluate_n(ctx);
      Groupby gby = ctx.get_groupby();
      // this covers a scenario where
      // we dont care about the presence or absence of NAs
      // we just want the total number of rows
      bool count_all_rows = arg_->get_expr_kind() == Kind::None;

      if (count_all_rows && !gby) {
        auto value = static_cast<int64_t>(ctx.nrows());
        Column coli = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
        outputs.add_column(std::move(coli), "count", Grouping::GtoONE);
        return outputs;
      }

      if (count_all_rows && gby) {
        Column coli = Column(new Latent_ColumnImpl(
                new CountAllRows_ColumnImpl(gby)
              ));
        outputs.add_column(std::move(coli), "count", Grouping::GtoONE);
        return outputs;
      }
```

In the code above, there is no addition of count and countna, it just grabs the total number of rows. Within a groupby it just does an i1 - 10 ( i basically reused the internal count0() implementation)

I have updated the docs to reflect that count of all rows is possible, irrespective of the presence of nulls. An example already existed to show that as well.

oleksiyskononenko: Yep, exactly.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,165 @@
...
+        case SType::STR64:
+          return make<CString, int64_t>(std::move(col), gby, is_grouped);                
+        default:
+          throw TypeError()
+            << "Invalid column of type `" << stype << "` in " << repr();
+      }
+    }
+
+
+    template <typename T, typename U>
```
oleksiyskononenko, line 125: I'm not sure why we need this additional `U` parameter since it is always `int64_t`. I can imagine we may have cases when the return type depends on the column type, but it is not the case for `count()` and `countna()`.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -99,37 +99,37 @@ class FExpr_CountNA : public FExpr_Func {
...
+    template <typename T>
     Column make(Column &&col, const Groupby& gby, bool is_grouped) const {
       if (is_grouped) {
-        return Column(new Latent_ColumnImpl(new Count_ColumnImpl<T, U, COUNTNA, true>(
+        return Column(new Latent_ColumnImpl(new Count_ColumnImpl<T, int64_t, COUNTNA, true>(
           std::move(col), gby
         )));
       } else {
-        return Column(new Latent_ColumnImpl(new Count_ColumnImpl<T, U, COUNTNA, false>(
+        return Column(new Latent_ColumnImpl(new Count_ColumnImpl<T, int64_t, COUNTNA, false>(
```
oleksiyskononenko: I don't think we need to pass it to `Count_ColumnImpl` either, because it is always `int64_t` and you know this inside the class already.

samukweku: @oleksiyskononenko it is used to generate the column stype in reduce_unary: 
```py
template <typename T, typename U, bool IS_GROUPED>
class ReduceUnary_ColumnImpl : public Virtual_ColumnImpl {
  protected:
    Column col_;
    Groupby gby_;
    

  public:
    ReduceUnary_ColumnImpl(Column &&col, const Groupby& gby)
      : Virtual_ColumnImpl(gby.size(), stype_from<U>),
        col_(std::move(col)),
        gby_(gby)
    {
      xassert(col_.can_be_read_as<T>());
    }


    ColumnImpl *clone() const override {
      return new ReduceUnary_ColumnImpl(Column(col_), Groupby(gby_));
    }


    size_t n_children() const noexcept override {
      return 1;
    }


    const Column &child(size_t i) const override {
      xassert(i == 0);
      (void)i;
      return col_;
    }
};
```
open to a better way of doing this

oleksiyskononenko: Yes, but for that you don't need to pass `int64_t` to `Count_ColumnImpl`, because this class already knows it is going to have `int64` stype and `int64_t` element type.

src/core/column/countna.h
```diff
@@ -0,0 +1,59 @@
...
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#ifndef dt_COLUMN_COUNTNA_h
+#define dt_COLUMN_COUNTNA_h
+#include "column/reduce_unary.h"
+namespace dt {
+
+
+template <typename T, typename U, bool COUNTNA, bool IS_GROUPED>
+class Count_ColumnImpl : public ReduceUnary_ColumnImpl<T, U, IS_GROUPED> {
```
oleksiyskononenko, line 29: Like here, you don't need any `U`, you can immediately pass `int64_t` to `ReduceUnary_ColumnImpl` constructor.

src/core/column/countna.h
```diff
@@ -0,0 +1,59 @@
...
+  public:
+    using ReduceUnary_ColumnImpl<T, U, IS_GROUPED>::ReduceUnary_ColumnImpl;
+
+    bool get_element(size_t i, U* out) const override {
```
oleksiyskononenko, line 33: Here `U` can also be safely replaced with `int64_t`, because the result of both `count()` and `countna()` is always `int64`.

src/core/column/countna.h
```diff
@@ -0,0 +1,59 @@
...
+      T value;
+      size_t i0, i1;
+      this->gby_.get_group(i, &i0, &i1);
+      int64_t count = 0;
```
oleksiyskononenko, line 37: Even here you can see you're using `int64_t` and not `U` for the final result.

docs/api/dt/count.rst
```diff
@@ -1,16 +1,17 @@
...
-    :src: src/core/expr/head_reduce_unary.cc count_reducer
+    :src: src/core/expr/fexpr_count_countna.cc pyfn_count
     :cvar: doc_dt_count
     :tests: tests/test-reduce.py
-    :signature: count(cols)
+    :signature: count(cols=None)
 
-    Calculate the number of non-missing values for each column from `cols`.
+    Calculate the number of non-missing values for each column from `cols`, if `cols` is provided, 
+    or the total number of rows if `cols` is not provided.
```
oleksiyskononenko, line 9:
```suggestion
    Calculate the number of non-missing values for each column from `cols`. When `cols` is not provided, 
    calculate the total number of rows.
```

docs/api/dt/count.rst
```diff
@@ -1,16 +1,17 @@
...
 
     Parameters
     ----------
     cols: FExpr
-        Input columns.
+        Input columns. If no `cols` is passed, then the count of all rows is returned.
```
oleksiyskononenko, line 14:
```suggestion
        Input columns if any.
```

oleksiyskononenko: We don't need to mention here what is returned, as this is not a `return:` section.

src/core/column/count_all_rows.h
```diff
@@ -0,0 +1,71 @@
...
+// IN THE SOFTWARE.
+//------------------------------------------------------------------------------
+#ifndef dt_COLUMN_COUNTALLROWS_h
+#define dt_COLUMN_COUNTALLROWS_h
+#include "column/virtual.h"
+#include "parallel/api.h"
+#include "stype.h"
+namespace dt {
+
+class CountAllRows_ColumnImpl : public Virtual_ColumnImpl {
```
oleksiyskononenko, line 29: Why not to inherit this class from `ReduceUnary_ColumnImpl`?

src/core/column/count_all_rows.h
```diff
@@ -0,0 +1,71 @@
...
+    ColumnImpl* clone() const override {
+      return new CountAllRows_ColumnImpl(gby_);
+    }
+
+
+    size_t n_children() const noexcept override {
+      return 0;
+    }
+
+    void materialize(Column &col_out, bool) override {
```
oleksiyskononenko, line 49: Why do we want to materialize it? This could be a virtual column that will return a number of elements for each requested group.

src/core/column/countna.h
```diff
@@ -0,0 +1,59 @@
...
+
+    bool get_element(size_t i, int64_t* out) const override {
+      T value;
+      size_t i0, i1;
+      this->gby_.get_group(i, &i0, &i1);
+      int64_t count = 0;
+      if (IS_GROUPED){
+        bool isvalid = this->col_.get_element(i, &value);
+        if (COUNTNA){
+          count = isvalid? 0: static_cast<int64_t>(i1 - i0);
```
oleksiyskononenko, line 41:
```suggestion
          count = isvalid? 0 : static_cast<int64_t>(i1 - i0);          
```

src/core/column/countna.h
```diff
@@ -0,0 +1,59 @@
...
+      size_t i0, i1;
+      this->gby_.get_group(i, &i0, &i1);
+      int64_t count = 0;
+      if (IS_GROUPED){
+        bool isvalid = this->col_.get_element(i, &value);
+        if (COUNTNA){
+          count = isvalid? 0: static_cast<int64_t>(i1 - i0);          
+        } else {
+            count = isvalid? static_cast<int64_t>(i1 - i0) : 0;            
+          }
```
oleksiyskononenko, line 44:
```suggestion
        }        
```

oleksiyskononenko: What is the editor you're using? Not sure why it sometimes messes up indentation.

src/core/column/countna.h
```diff
@@ -0,0 +1,59 @@
...
+        *out = count;
+        return true;  // *out is not NA
+      } else {
+          for (size_t gi = i0; gi < i1; ++gi) {        
+            bool isvalid = this->col_.get_element(gi, &value);
+            count += COUNTNA? !isvalid : isvalid;
+          }
+          *out = count;
+          return true;  // *out is not NA
+        }
```
oleksiyskononenko, line 54:
```suggestion
        for (size_t gi = i0; gi < i1; ++gi) {        
          bool isvalid = this->col_.get_element(gi, &value);
          count += COUNTNA? !isvalid : isvalid;
        }
        *out = count;
        return true;  // *out is not NA
      }
```

src/core/column/reduce_unary.h
```diff
@@ -26,15 +26,16 @@
 namespace dt {
 
 
-template <typename T, bool IS_GROUPED>
+template <typename T, typename U, bool IS_GROUPED>
```
oleksiyskononenko, line 29: I guess since we now have two types here, it would be nice to have either some intuitive names like `T_IN` and `T_OUT`, or add some comments as to what `T` and `U` mean. The former is probably more preferable.

src/core/column/mean.h
```diff
@@ -25,20 +25,20 @@
 namespace dt {
 
 
-template <typename T, bool IS_GROUPED>
-class Mean_ColumnImpl : public ReduceUnary_ColumnImpl<T, IS_GROUPED> {
+template <typename T, typename U, bool IS_GROUPED>
```
oleksiyskononenko, line 28: We don't really need to pass `U` for mean, minmax and sumprod calculations, because it is identical to `T`.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,165 @@
...
+      // we dont care about the presence or absence of NAs
+      // we just want the total number of rows
+      bool count_all_rows = arg_->get_expr_kind() == Kind::None;
+
+      if (count_all_rows && !COUNTNA) {
+        Column coli;
+        if (gby){
+          coli = Column(new Latent_ColumnImpl(
+                new CountAllRows_ColumnImpl(gby)
+              ));
```
oleksiyskononenko, line 67: Indentation issue.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,165 @@
...
+        } else {
+            auto value = static_cast<int64_t>(ctx.nrows());
+            coli = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
+          }
```
oleksiyskononenko, line 71: Same here.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,165 @@
...
+    Workframe evaluate_n(EvalContext &ctx) const override {
+      Workframe outputs(ctx);
+      Workframe wf = arg_->evaluate_n(ctx);
+      Groupby gby = ctx.get_groupby();
+      // this covers scenarios where
+      // we dont care about the presence or absence of NAs
+      // we just want the total number of rows
+      bool count_all_rows = arg_->get_expr_kind() == Kind::None;
+
+      if (count_all_rows && !COUNTNA) {
```
oleksiyskononenko, line 62: `countna()` could also be called with no `cols`, in such a case we are currently returning `0` for each group. Is this case covered on this PR?

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,165 @@
...
+      for (size_t i = 0; i < wf.ncols(); ++i) {
+        bool is_grouped = ctx.has_group_column(
+                            wf.get_frame_id(i),
+                            wf.get_column_id(i)
+                          );
+        Column coli = wf.retrieve_column(i);   
+        if (COUNTNA && !ctx.has_groupby() && (coli.stype() == SType::VOID)) {
+          int64_t nrows = static_cast<int64_t>(ctx.nrows());
+          coli = Const_ColumnImpl::make_int_column(1, nrows, SType::INT64);
+        } else {coli = evaluate1(std::move(coli), gby, is_grouped);}
```
oleksiyskononenko, line 89:
```suggestion
        } else {
          coli = evaluate1(std::move(coli), gby, is_grouped);
        }
```

tests/dt/test-countna.py
```diff
@@ -71,6 +71,12 @@ def test_dt_count_na1(src):
...
-    EXP = dt.Frame(G=[1,2], V1=[3,1], V2=[3,0])
+    EXP = dt.Frame(G=[1,2], V1=[3,1], V2=[1,0])
     RES = DT[:, [dt.countna(f.V), dt.countna(dt.mean(f.V))], dt.by(f.G)]
     assert EXP.to_list() == RES.to_list()
+
+
+def test_dt_countna_void():
+    DT = dt.Frame([None])
+    RES = DT[:, dt.countna(f.C0), dt.by(f.C0)]
+    EXP = dt.Frame({"C0":[None], "C1":[1]/dt.int64})
```
oleksiyskononenko, line 82: New line is needed.

tests/dt/test-countna.py
```diff
@@ -71,6 +71,12 @@ def test_dt_count_na1(src):
 
 def test_dt_count_na2():
     DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])
-    EXP = dt.Frame(G=[1,2], V1=[3,1], V2=[3,0])
+    EXP = dt.Frame(G=[1,2], V1=[3,1], V2=[1,0])
```
oleksiyskononenko, line 74: Hm, I wonder if that was a bug?

src/core/expr/fexpr_count_countna.cc
samukweku: is it useful though?

tests/dt/test-countna.py
samukweku: yea, it was. I didnt think of it until i compared results. The other surprising part for me is the fact that nested groupby works here with no drama, unlike sum/prod/min/max ... any ideas?

src/core/column/countna.h
samukweku: i use vscode ... any better one to suggest for use with c++?

src/core/expr/fexpr_count_countna.cc
oleksiyskononenko: I don't know, but may be good for consistency.

tests/dt/test-countna.py
oleksiyskononenko: Then may be open an issue illustrating the bug, so that we knew it is fixed by this PR. Do you know what part of the C++ code had the bug?

As for the aggregations errors, I actually found the reason why it errors, but had no time to fix that. I will work on that soon.

samukweku: no idea on what part of the c++ code caused the bug. Do i link it to the existing [issue](#3417 )?

src/core/expr/fexpr_count_countna.cc
samukweku: what am I missing with the indetation here? the `else` curly brackets line up

tests/dt/test-countna.py
oleksiyskononenko: Yes, you can. It is still a bug in `countna()`, just add an example.

src/core/expr/fexpr_count_countna.cc
oleksiyskononenko: The code for both if and else should have the same indentation. In this case the code that goes after else should be moved two spaces to the left, including the closing bracket.

src/core/expr/fexpr_count_countna.cc
samukweku: yea, I dont think it fits in. we are replicating SQL here, where count(*) returns all rows, whereas count(column) counts the non null rows for that column.

oleksiyskononenko: Well, but if someone already using it, it makes no sense to remove this functionality.

oleksiyskononenko: I don’t think anyone is using countna(), the use-case could be countna(cols), where cols could become None at some point. My feeling is that returning 0 or None is better in this case, than throwing an error.

docs/api/dt/count.rst
```diff
@@ -16,6 +16,8 @@
     return: Expr
         f-expression having one row, and the same names and number of columns
         as in `cols`. All the returned column stypes are `int64`.
+        If `cols` is not provided, the total number of rows 
+        (a combination of the count of missing and non-missing values) is returned.
```
oleksiyskononenko, line 20: Oh, by the way, could you please update the "See Also" sections for `count()` and `countna()` to point to each other? Should be pretty useful.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,171 @@
...
+      // we dont care about the presence or absence of NAs
+      // we just want the total number of rows
+      bool count_all_rows = arg_->get_expr_kind() == Kind::None;
+
+      if (count_all_rows && !COUNTNA) {
+        Column coli;
+        auto value = static_cast<int64_t>(ctx.nrows());
+        if (gby){
+          coli = Const_ColumnImpl::make_int_column(value, 1, SType::INT64);
+          coli = Column(new Latent_ColumnImpl(new CountAllRows_ColumnImpl(std::move(coli), gby)));
```
oleksiyskononenko, line 67: We need to streamline this and inherit `CountAllRows_ColumnImpl` from
`ReduceNullary_ColumnImpl`. Then you won't need to create a dummy `coli` here. `ReduceNullary_ColumnImpl` could look something like this

```cpp
class ReduceNullary_ColumnImpl : public Virtual_ColumnImpl {
  protected:
    Groupby gby_;

  public:
    ReduceUnary_ColumnImpl(const Groupby& gby, SType stype)
      : Virtual_ColumnImpl(gby.size(), stype),
        gby_(gby)
    {}


    ColumnImpl *clone() const override {
      return new ReduceNullary_ColumnImpl(Groupby(gby_), this->stype());
    }


    size_t n_children() const noexcept override {
      return 0;
    }
};
```

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,171 @@
...
+        } else{            
+            coli = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
+          }
```
oleksiyskononenko, line 70:
```suggestion
        }
```

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,171 @@
...
+                            wf.get_frame_id(i),
+                            wf.get_column_id(i)
+                          );
+        Column coli = wf.retrieve_column(i);   
+        if (COUNTNA && !ctx.has_groupby() && (coli.stype() == SType::VOID)) {
+          int64_t nrows = static_cast<int64_t>(ctx.nrows());
+          coli = Const_ColumnImpl::make_int_column(1, nrows, SType::INT64);
+        } else {
+            coli = evaluate1(std::move(coli), gby, is_grouped);
+          }
```
oleksiyskononenko, line 96:
```suggestion
        }
```

src/core/expr/fexpr_count_countna.cc
oleksiyskononenko, line 70:
```suggestion
          coli = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
        }
```

src/core/expr/fexpr_count_countna.cc
oleksiyskononenko, line 96:
```suggestion
          coli = evaluate1(std::move(coli), gby, is_grouped);
        }
```

src/core/expr/fexpr_count_countna.cc
samukweku: is it not simpler to just derive the `CountAllRows_ColumnImpl` from the `Virtual_ColumnImpl` directly? or are there plans to reuse the `ReduceNullary_ColumnImpl` somewhere else?

oleksiyskononenko: Currently, there is only one function in head_reduce_nullary.cc, that is count0(). So if you want, you can inherit directly from Virtual_ColumnImpl. Inheriting from ReduceNullary_ColumnImpl will just add an indicator that count0() is not going to use any columns as an input.

src/core/expr/fexpr_sumprod.cc
```diff
@@ -83,27 +83,26 @@ class FExpr_SumProd : public FExpr_Func {
...
         case SType::INT32:
         case SType::INT64:
-          return make<int64_t>(std::move(col), SType::INT64, gby, is_grouped);
+          return make<int64_t>(std::move(col), gby, is_grouped);
         case SType::FLOAT32:
-          return make<float>(std::move(col), SType::FLOAT32, gby, is_grouped);
+          return make<float>(std::move(col), gby, is_grouped);
         case SType::FLOAT64:
-          return make<double>(std::move(col), SType::FLOAT64, gby, is_grouped);
+          return make<double>(std::move(col), gby, is_grouped);
```
oleksiyskononenko, line 90: Before we set the output stype in this `switch` and then casted the column to this type as 
```
col.cast_inplace(stype);
```

Now, you removed the stypes, so how the code will create `int64` result column for, let's say, `int8` input?

samukweku: The int64 is created in the sumprod.h column

samukweku: `template <typename T_IN, bool SUM, bool IS_GROUPED>
class SumProd_ColumnImpl : public ReduceUnary_ColumnImpl<T_IN, T_IN, IS_GROUPED> {
  public:
    using ReduceUnary_ColumnImpl<T_IN, T_IN, IS_GROUPED>::ReduceUnary_ColumnImpl;

    bool get_element(size_t i, T_IN* out) const override {
      T_IN result = !SUM; // 0 for `sum()` and 1 for `prod()`
      T_IN value;
      size_t i0, i1;
      this->gby_.get_group(i, &i0, &i1);
`

The int64 is passed here

oleksiyskononenko: I think there is an error here and that's why the build is failing on AppVeyor.

There are two different things involved: the column's stype and the column's element type. What we did before was
- based on the column's stype determine the resulting column's stype;
- cast the input column to the resulting column's stype;
- get elements based on the resulting column's element type.

What you're doing now
- eliminated the resulting column stype;
- not casting the initial column to the resulting column stype;
- trying to get elements based on the resulting column element type.

The latter will not work, because the input column's `get_element()` could only be safely used for the input column element type.

What was the reason you removed passing the stype and casting?

samukweku: ahh thanks for the explanation. if that's the case, then i can just use `<int8_t, int64_t>` then ... that way `get_element` gets the value based on `int8`, which is the dtype of the input column, while the output column is created as `int64`.

I didnt think the column casting was necessary; however, with your explanation I have a better understanding - i'll update the code for sumprod, minmax, mean

oleksiyskononenko: Yes, you can just revert your changes introduced to that switch, because we indeed need all the details like column value type, column stype, etc.

src/core/expr/fexpr_sumprod.cc
```diff
@@ -83,6 +83,7 @@ class FExpr_SumProd : public FExpr_Func {
         case SType::INT16:
         case SType::INT32:
         case SType::INT64:
+          col.cast_inplace(SType::INT64);
```
oleksiyskononenko, line 86: You can do this, but then you need to also do the same for floats, etc, so better to revert the changes to what it was before this PR.

oleksiyskononenko: This failure is not related to your PR:

```
    def test_dt_count_na2():
        DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])
        EXP = dt.Frame(G=[1,2], V1=[3,1], V2=[1,0])
>       RES = DT[:, [dt.countna(f.V), dt.countna(dt.mean(f.V))], dt.by(f.G)]
E       AssertionError: Assertion 'i < nrows()' failed in src\core\column.cc, line 251
DT         = <Frame#229a5d3da80 6x2>

```

You need to disable this test until https://github.com/h2oai/datatable/issues/3417 is resolved.

src/core/column/reduce_unary.h
```diff
@@ -26,19 +26,20 @@
...
 class ReduceUnary_ColumnImpl : public Virtual_ColumnImpl {
   protected:
     Column col_;
     Groupby gby_;
+    
 
   public:
     ReduceUnary_ColumnImpl(Column &&col, const Groupby& gby)
-      : Virtual_ColumnImpl(gby.size(), col.stype()),
+      : Virtual_ColumnImpl(gby.size(), stype_from<T_OUT>),
```
oleksiyskononenko, line 38: You need to pass `stype` here just like it was before, otherwise it is not possible to properly get it from `T_OUT`. I'm not sure what was the reason to change it?

src/core/expr/fexpr_mean.cc
```diff
@@ -80,42 +80,42 @@ class FExpr_Mean : public FExpr_Func {
...
         case SType::FLOAT64:
-          return make<double>(std::move(col), SType::FLOAT64, gby, is_grouped);
+          return make<double, double>(std::move(col), gby, is_grouped);
         case SType::FLOAT32:
-          return make<float>(std::move(col), SType::FLOAT32, gby, is_grouped);
-
+          return make<float, float>(std::move(col), gby, is_grouped);
         case SType::DATE32: {
-          Column coli = make<double>(std::move(col), SType::FLOAT64, gby, is_grouped);
+          Column coli = make<double, double>(std::move(col), gby, is_grouped);
```
oleksiyskononenko, line 95: So why it is still not passing stype here and why do we need to pass additional type to `make`? I propose you revert the changes related to this switch, as we already discussed it will not work this way.

src/core/expr/fexpr_sumprod.cc
```diff
@@ -97,7 +96,8 @@ class FExpr_SumProd : public FExpr_Func {
 
 
     template <typename T_IN>
-    Column make(Column &&col, const Groupby& gby, bool is_grouped) const {
+    Column make(Column &&col, SType stype, const Groupby& gby, bool is_grouped) const {
+      col.cast_inplace(stype);
```
oleksiyskononenko, line 100: You need to restore it for all the reducers, as this PR is only related to `count()` function it is better to only change what is necessary.

oleksiyskononenko:
@samukweku I guess we need to go through this PR and revert all the changes related to reducers other than `count()`. Only minor changes should be left that make other reducers compatible with the new `ReduceUnary_ColumnImpl`, i.e. calling `ReduceUnary_ColumnImpl<T, T>` instead of just `ReduceUnary_ColumnImpl<T>`.

src/core/column/reduce_unary.h
samukweku: the change was made here to allow passing int64 for count, instead of casting say a float to int64, as we did for the other reducers so far(sum/prod/mean). For min/max, we did not need to cast since it is within the same data type. Another option would be to pass an explicit stype here, instead of deriving the stype from T_OUT

oleksiyskononenko: Well, but the resulting column for `count()` has `int64` stype, so `col.stype()` will work in this case too. Anyways, `stype_from<T_OUT>` can only give you a guess.

samukweku: why does `stype_from<T_OUT>` not work properly? maybe something I am missing from my interpretation of the function? or is `Type::from` d right way to go?

src/core/expr/fexpr_mean.cc
samukweku: take for instance `<int8_t, double>` - `int8_t` is the dtype of the input column, while `double` is the dtype of the output column. this is what makes up the template form `<T_IN, T_OUT>`, and in the reduce_unary.h file, in line 42, there is a `col_.can_be_read_as<T_IN>` and since int8_t is the dtype of the input column, there should be no issue. This allows us to pass dtype for scenarios where deriving the dtype from the input column might not be desirable. Say for example we are doing a count and we have a `FLOAT64` column, we shouldnt cast it to int64 to lose precision. instead we can do `<double, int64_t>` and reduce_unary.h will take care of the rest. As usual, I am open to feedback - helps with my understanding of the codebase. thanks @oleksiyskononenko

src/core/column/reduce_unary.h
oleksiyskononenko: The way to go is to use `col.stype()` as it was before..

It will not work because there is no strict way to get stype from the underlying type. We may have different column to have the same C++ type: date32/int32, time64/int64, boo8/int8, etc.

src/core/expr/fexpr_mean.cc
oleksiyskononenko: This piece of code is not related to `count()`, so here we don't need to change anything.

src/core/column/mean.h
```diff
@@ -25,20 +25,20 @@
 namespace dt {
 
 
-template <typename T, bool IS_GROUPED>
-class Mean_ColumnImpl : public ReduceUnary_ColumnImpl<T, IS_GROUPED> {
+template <typename T_IN, typename T_OUT, bool IS_GROUPED>
+class Mean_ColumnImpl : public ReduceUnary_ColumnImpl<T_IN, T_OUT, IS_GROUPED> {
```
oleksiyskononenko, line 29: For mean, for instance, you only need to pass one `T` and then call `ReduceUnary_ColumnImpl<T, T, IS_GROUPED>` — because you already casted the input column to the corresponding stype. May be I don't really understand what you're trying to accomplish here...

src/core/column/reduce_unary.h
samukweku: so it is ok to cast say a float64 to int64 for the count() function? no need to worry about precision?

oleksiyskononenko: But why do we need to cast it for count?

oleksiyskononenko: All reducers are separate classes, casting was needed for some of them outside of this PR. What I say is that you need to restore the original implementation for those reducers, otherwise it will not work with only passing the C++ type.

Note, tests are failing not for count…

src/core/column/mean.h
```diff
@@ -25,35 +25,35 @@
 namespace dt {
 
 
-template <typename T_IN, typename T_OUT, bool IS_GROUPED>
-class Mean_ColumnImpl : public ReduceUnary_ColumnImpl<T_IN, T_OUT, IS_GROUPED> {
+template <typename T_IN, bool IS_GROUPED>
```
oleksiyskononenko, line 28: For mean you can just call it T, as T_IN is the same as T_OUT.

oleksiyskononenko:
I've merged #3446, so now you need to merge `main` onto this PR and
- make use of `FExpr_ReduceUnary` for the counting reducers, will make life easier with less code;
- adopt other changes that should fix issues with `count()` and `countna()` used as nested reducers.

oleksiyskononenko:
You just needed to do `git merge main` on your branch.

samukweku:
@oleksiyskononenko i'm at a loss on how to adapt the current count function to the reduce_unary format? 

```py
std::string repr() const override {
      std::string out = COUNTNA? "countna" : "count";
      out += '(';
      if (arg_->get_expr_kind() != Kind::None) out += arg_->repr();      
      out += ')';
      return out;
    }
```

with the condition, if it is to count all rows then `count()`, else `count(f.A)`

also, there are a couple of conditions baked in that I do not know how to convert to the current form: 

```py
Workframe evaluate_n(EvalContext &ctx) const override {
      Workframe outputs(ctx);
      Workframe wf = arg_->evaluate_n(ctx);
      Groupby gby = ctx.get_groupby();
      // this covers scenarios where
      // we dont care about the presence or absence of NAs
      // we just want the total number of rows
      bool count_all_rows = arg_->get_expr_kind() == Kind::None;

      if (count_all_rows && !COUNTNA) {
        Column coli;
        auto value = static_cast<int64_t>(ctx.nrows());
        if (gby){
          coli = Column(new Latent_ColumnImpl(new CountAllRows_ColumnImpl(gby, SType::INT64)));
        } else{            
          coli = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
        }
        outputs.add_column(std::move(coli), "count", Grouping::GtoONE);
        return outputs;
      }
```

another part where I am stumped is the stype. in `reduce_unary.cc` the stype is derived directly from the column, but count has just an int64  as a single output, regardless of the input stype. It feels unsafe to cast the original column to an int64 - imagine a string column - i do not think it is safe to convert to int64.

need your guidance on how to proceed

oleksiyskononenko:
@samukweku 

— instead of repr() just add the same logic to name(), see how it is done for sum/prod;

— for count() with no arguments we should not create an unary FExpr, but rather a nullary, so the condition should be moved outside;

— for reduce_unary.cc just keep your updates with respect to T_IN/T_OUT.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+        case SType::TIME64: 
+        case SType::INT64:
+          return make<int64_t>(std::move(col), SType::INT64, gby, is_grouped);
+        case SType::FLOAT32:
+          return make<float>(std::move(col), SType::INT64, gby, is_grouped);
+        case SType::FLOAT64:
+          return make<double>(std::move(col), SType::INT64, gby, is_grouped);
+        case SType::STR32:
+        case SType::STR64:
+          return make<CString>(std::move(col), SType::INT64, gby, is_grouped);
```
oleksiyskononenko, line 67: We don't need to pass `stype` here, because it is always `SType::INT64`.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+    template <typename T>
+    Column make(Column&& col, SType stype, const Groupby& gby, bool is_grouped) const {
+      if (is_grouped) {
+        return Column(new Latent_ColumnImpl(new Count_ColumnImpl<T, COUNTNA, true>(
+          std::move(col), stype, gby
+        )));
+      } else {
+        return Column(new Latent_ColumnImpl(new Count_ColumnImpl<T, COUNTNA, false>(
+          std::move(col), stype, gby
+        )));
```
oleksiyskononenko, line 83: Don't need `stype` here either, because `Count_ColumnImpl` knows it is always `SType::INT64`.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+#include "expr/fexpr_reduce_unary.h"
+#include "expr/eval_context.h"
+#include "expr/workframe.h"
+#include "python/xargs.h"
+#include "stype.h"
+namespace dt {
+namespace expr {
+
+template<bool COUNTNA>
+class FExpr_CountNA : public FExpr_ReduceUnary {
```
oleksiyskononenko, line 36: Better to call it `FExpr_Count`, as it is more general just like this class is.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+          std::move(col), stype, gby
+        )));
+      }
+    }
+};
+
+
+// gets the count of all rows - nulls are not checked
+template<bool COUNTNA>
+class FExpr_CountNA_AllRows : public FExpr_Func {
```
oleksiyskononenko, line 91: Also, better to call this class `FExpr_Count_Rows`, as it is not just counting all rows — for group by it is the number of rows per group.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+  public:
+    FExpr_CountNA_AllRows(){}
+
+    std::string repr() const override {
+      std::string out = COUNTNA ? "countna(None)" 
+                                : "count()";
```
oleksiyskononenko, line 97:
```suggestion
      std::string out = COUNTNA? "countna()" 
                               : "count()";
```

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+  Workframe evaluate_n(EvalContext &ctx) const override {
+    Workframe wf(ctx);
+    Groupby gby = ctx.get_groupby();
+    Column col;
+
+    if (!gby) {
+      gby = Groupby::single_group(wf.nrows());
+    }
```
oleksiyskononenko, line 108: This block can be removed, because `gby` is never empty in the context. If there is no `by()` clause, `gby` will be filled with a single group already.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+      col = Const_ColumnImpl::make_int_column(gby.size(), 0, SType::INT64);
+      wf.add_column(std::move(col), std::string(), Grouping::GtoONE);
+      return wf;
+    }
+    
+    if (ctx.has_groupby()) {
+      col = Column(new Latent_ColumnImpl(new CountAllRows_ColumnImpl(gby)));
+    } else {
+        auto value = static_cast<int64_t>(ctx.nrows());
+        col = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
```
oleksiyskononenko, line 120:
```suggestion
      auto value = static_cast<int64_t>(ctx.nrows());
      col = Const_ColumnImpl::make_int_column(1, value, SType::INT64);
```

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+    Groupby gby = ctx.get_groupby();
+    Column col;
+
+    if (!gby) {
+      gby = Groupby::single_group(wf.nrows());
+    } 
+
+    if (COUNTNA) {
+      col = Const_ColumnImpl::make_int_column(gby.size(), 0, SType::INT64);
+      wf.add_column(std::move(col), std::string(), Grouping::GtoONE);
```
oleksiyskononenko, line 112:
```suggestion
      wf.add_column(std::move(col), "countna", Grouping::GtoONE);
```

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+    }
+    wf.add_column(std::move(col), "count", Grouping::GtoONE);
+    return wf;
+  }
+
+};
+
+
+static py::oobj pyfn_count(const py::XArgs &args) {
+  auto count = args[0].to_oobj_or_none();
```
oleksiyskononenko, line 130: Better to call this variable `arg` or something, because `count` is the function name.

src/core/expr/fexpr_count_countna.cc
```diff
@@ -0,0 +1,160 @@
...
+  if (count.is_none()) {
+    return PyFExpr::make(new FExpr_CountNA_AllRows<false>());
+  }
+  return PyFExpr::make(new FExpr_CountNA<false>(as_fexpr(count)));
+}
+
+static py::oobj pyfn_countna(const py::XArgs &args) {
+  auto countna = args[0].to_oobj_or_none();
```
oleksiyskononenko, line 138: Same here.

src/core/expr/fexpr_mean.cc
```diff
@@ -100,4 +100,4 @@ DECLARE_PYFN(&pyfn_mean)
     ->n_required_args(1);
 
 
-}}  // dt::expr
+}}  // dt::expr
```
oleksiyskononenko, line 103: Newline is missing.

src/core/expr/fexpr_mean.cc
```diff
@@ -81,7 +81,7 @@ class FExpr_Mean : public FExpr_ReduceUnary {
 
       return is_grouped? std::move(col)
                        : Column(new Latent_ColumnImpl(new Mean_ColumnImpl<T>(
-                           std::move(col), gby
+                           std::move(col), stype, gby
```
oleksiyskononenko, line 84: We don't need to pass `stype` to `Mean_ColumnImpl`. It is the `Mean_ColumnImpl` that needs to pass `stype` to `ReduceUnary_ColumnImpl` constructor and it already knows that the `stype` is simply `col.stype()`.

src/core/expr/fexpr_minmax.cc
```diff
@@ -52,19 +52,21 @@ class FExpr_MinMax : public FExpr_ReduceUnary {
...
+        return make<int64_t>(std::move(col), SType::INT64, gby, is_grouped);
         case SType::TIME64:
-          return make<int64_t>(std::move(col), gby, is_grouped);
+          return make<int64_t>(std::move(col), SType::TIME64, gby, is_grouped);
         case SType::FLOAT32:
-          return make<float>(std::move(col), gby, is_grouped);
+          return make<float>(std::move(col), SType::FLOAT32, gby, is_grouped);
         case SType::FLOAT64:
-          return make<double>(std::move(col), gby, is_grouped);
+          return make<double>(std::move(col), SType::FLOAT64, gby, is_grouped);
```
oleksiyskononenko, line 69: Passing `stype` is not needed here either.

src/core/expr/fexpr_minmax.cc
```diff
@@ -73,10 +75,10 @@ class FExpr_MinMax : public FExpr_ReduceUnary {
 
 
     template <typename T>
-    Column make(Column&& col, const Groupby& gby, bool is_grouped) const {
+    Column make(Column&& col, SType stype, const Groupby& gby, bool is_grouped) const {
       return is_grouped? std::move(col)
                        : Column(new Latent_ColumnImpl(new MinMax_ColumnImpl<T, MIN>(
-                           std::move(col), gby
+                           std::move(col), stype, gby
```
oleksiyskononenko, line 81: Not needed here either.

src/core/expr/fexpr_sumprod.cc
```diff
@@ -110,4 +110,4 @@ DECLARE_PYFN(&pyfn_prod)
     ->n_required_args(1);
 
 
-}}  // dt::expr
+}}  // dt::expr
```
oleksiyskononenko, line 113: Newline is missing.

src/core/expr/fexpr_sumprod.cc
```diff
@@ -72,11 +72,11 @@ class FExpr_SumProd : public FExpr_ReduceUnary {
       col.cast_inplace(stype);
       if (is_grouped) {
         return Column(new Latent_ColumnImpl(new SumProd_ColumnImpl<T, SUM, true>(
-          std::move(col), gby
+          std::move(col), stype, gby
         )));
       } else {
         return Column(new Latent_ColumnImpl(new SumProd_ColumnImpl<T, SUM, false>(
-          std::move(col), gby
+          std::move(col), stype, gby
```
oleksiyskononenko, line 79: Not needed here too.

tests/dt/test-countna.py
```diff
@@ -68,9 +68,27 @@ def test_dt_count_na1(src):
     RES = df[:, dt.countna(f[:])]
     assert_equals(EXP, RES)
 
-
+@pytest.mark.xfail(reason="commented out till #3417 is resolved.")
```
oleksiyskononenko, line 71: You can enable this test, as #3417 is resolved for all `FExpr` reducers.

tests/dt/test-countna.py
```diff
@@ -68,9 +68,27 @@ def test_dt_count_na1(src):
...
+    DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])
+    EXP = dt.Frame(G=[1,2], C0=[0,0])
+    RES = DT[:, dt.countna(), f.G]
+    assert EXP.to_list() == RES.to_list()
+
+def test_dt_countna_None():
+    DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])
+    EXP = dt.Frame(C0=[0])
+    RES = DT[:, dt.countna()]
+    assert EXP.to_list() == RES.to_list()
```
oleksiyskononenko, line 94: Newline is missing.

src/core/expr/fexpr_mean.cc
samukweku: i dont understand - `reduce_unary` requries an SType:

```py
template <typename T_IN, typename T_OUT>
class ReduceUnary_ColumnImpl : public Virtual_ColumnImpl {
  protected:
    Column col_;
    Groupby gby_;
    

  public:
    ReduceUnary_ColumnImpl(Column &&col, SType stype, const Groupby& gby)
      : Virtual_ColumnImpl(gby.size(), stype),
        col_(std::move(col)),
        gby_(gby)
    {
      xassert(col_.can_be_read_as<T_IN>());
    }


    ColumnImpl *clone() const override {
      return new ReduceUnary_ColumnImpl(Column(col_), this->stype(), Groupby(gby_));
    }
```
using the stype allows passing either the col's stype, or in the case of count, pass SType::INT64

src/core/expr/fexpr_count_countna.cc
samukweku: I'm confused here ... usually we check for an empty gby here right - is there a condition here that negates this need?

src/core/expr/fexpr_minmax.cc
samukweku: stype passed here is used to cast in place

src/core/expr/fexpr_count_countna.cc
samukweku: The original `reduce_unary` derives its stype from the col: 

```py
template <typename T>
class ReduceUnary_ColumnImpl : public Virtual_ColumnImpl {
  protected:
    Column col_;
    Groupby gby_;

  public:
    ReduceUnary_ColumnImpl(Column &&col, const Groupby& gby)
      : Virtual_ColumnImpl(gby.size(), col.stype()),
        col_(std::move(col)),
        gby_(gby)
    {
      xassert(col_.can_be_read_as<T>());
    }


    ColumnImpl *clone() const override {
      return new ReduceUnary_ColumnImpl(Column(col_), Groupby(gby_));
    }


    size_t n_children() const noexcept override {
      return 1;
    }


    const Column &child(size_t i) const override {
      xassert(i == 0);
      (void)i;
      return col_;
    }
};
```

I tweaked it such that you can explicitly pass an SType instead - this is handy for the count function

```py
template <typename T_IN, typename T_OUT>
class ReduceUnary_ColumnImpl : public Virtual_ColumnImpl {
  protected:
    Column col_;
    Groupby gby_;
    

  public:
    ReduceUnary_ColumnImpl(Column &&col, SType stype, const Groupby& gby)
      : Virtual_ColumnImpl(gby.size(), stype),
        col_(std::move(col)),
        gby_(gby)
    {
      xassert(col_.can_be_read_as<T_IN>());
    }


    ColumnImpl *clone() const override {
      return new ReduceUnary_ColumnImpl(Column(col_), this->stype(), Groupby(gby_));
    }


    size_t n_children() const noexcept override {
      return 1;
    }


    const Column &child(size_t i) const override {
      xassert(i == 0);
      (void)i;
      return col_;
    }
};
```

Are you suggesting I revert to your initial setup, and just use a different class altogether for the `count` function? maybe adapt the `FExpr_Count_Rows` function instead?

docs/api/dt/countna.rst
```diff
@@ -17,6 +17,7 @@
     return: Expr
```
oleksiyskononenko, line 17: Should be `return: FExpr`.

docs/api/dt/countna.rst
```diff
@@ -64,6 +65,16 @@
          0 |     2
         [1 row x 1 column]
 
+    Get the count if no col is passed:
```
oleksiyskononenko, line 68:
```suggestion
    When no `cols` is passed, the number of missing values returned is zero:
```

src/core/expr/fexpr_mean.cc
oleksiyskononenko: >i dont understand - reduce_unary requries an SType

Yes, but you can get it from `col.stype()` inside of `Mean_ColumnImpl`.

src/datatable/expr/reduce.py
```diff
@@ -104,10 +109,10 @@ def corr(col1, col2):
 
 
 # noinspection PyShadowingBuiltins
-def sum(iterable, start=0):
+def sum(iterable=None, start=0):
```
oleksiyskononenko, line 112: I'm not sure why we need this change. Also, this PR is for the count function, so I propose we don't change others to make it clear.

src/core/expr/fexpr_count_countna.cc
oleksiyskononenko: Yes, but that check was wrong. Look here: https://github.com/h2oai/datatable/blob/main/src/core/expr/eval_context.h#L56-L58

That's why I had to push these changes: https://github.com/h2oai/datatable/blob/main/src/core/expr/fexpr_reduce_unary.cc#L35-L40

samukweku:
`Yes, but you can get it from col.stype() inside of Mean_ColumnImpl.` Do you mind showing me how, an example will make things much clearer for me. thanks @oleksiyskononenko

oleksiyskononenko:
@samukweku Sure, just let me know when you're finished with changes to this PR and I will push the other needed changes, then we merge it.

oleksiyskononenko:
@samukweku I'm not sure why we need dozens of the old commits been pushed to all your PRs?

oleksiyskononenko:
May I ask you not to push anything onto this branch? As you asked me to show you how to handle several review comments, I was going to push some changes here and suddenly got +60 old commits to deal with.

src/core/expr/fexpr_minmax.cc
```diff
@@ -56,11 +56,9 @@ class FExpr_MinMax : public FExpr_ReduceUnary {
         case SType::INT16:
           return make<int16_t>(std::move(col), gby, is_grouped);
         case SType::INT32:
-        return make<int32_t>(std::move(col), gby, is_grouped);
         case SType::DATE32:
           return make<int32_t>(std::move(col), gby, is_grouped);
         case SType::INT64:
-        return make<int64_t>(std::move(col), gby, is_grouped);
```
oleksiyskononenko: Not sure why this change was introduced, we don't need additional treatment of date32/time64 as internally there are the same as int32/int64.

oleksiyskononenko (approved):

oleksiyskononenko:
@samukweku Please take a look at the changes, and let me know if you have any questions.

samukweku:
@oleksiyskononenko my apologies; must have been when I rebased the branch. apologies on that

oleksiyskononenko:
@samukweku To merge `main`, simply do `git merge main` being on the destination branch. It will then create just one commit and make things simple. Anyways, I guess I've resolved all the conflicts and PR is ready for your review.

samukweku:
looks good. thanks @oleksiyskononenko . need an explanation of the code in reduceunary: 

```
    ReduceUnary_ColumnImpl(Column &&col, const Groupby& gby)
      : ReduceUnary_ColumnImpl(std::move(col), gby, col.stype())
    {}


    ColumnImpl *clone() const override {
      return new ReduceUnary_ColumnImpl(Column(col_), Groupby(gby_), this->stype());
    }
```

the first part of the code gets its type directly from the column, while the other part seems to get it from the SType provided? maybe an explanation of clone and the preceding code would be helpful.

oleksiyskononenko:
@samukweku The first part of the code you're referring to is a constructor, that is needed for the case when `stype_out` is the same as `col.stype()`. It is calling the main constructor, that accepts three parameters. This is exactly what you've been asking about, the way we need to deal with mean, min/max, etc.

samukweku:
Thanks for the explanation @oleksiyskononenko LGTM to merge

PR #3446, opened 2023-03-21
Fix chained `FExpr` reducers

oleksiyskononenko:
Chained reducers is a pretty rare use case, because the second reducer gets a scalar (per group) as an input. Some SQL engines don't even allow that. However, we never seems to forbid chaining explicitly, at the same time, we had no tests and didn't think such a use case through. As a result, we either produced wrong results (see #3417) or even segfaulted when chaining was involved. In this PR we fix all the known related issues and add some tests. We also

- refactor `FExpr` reducers to inherit from `FExpr_ReduceUnary`;
- improve reducers logic with respect to grouped/non-grouped frames.

WIP for #3417

### DIFF
--- a/docs/releases/v1.1.0.rst
+++ b/docs/releases/v1.1.0.rst
@@ -151,6 +151,8 @@
     -[fix] Fixed :func:`dt.median()` when used in a groupby context with
        :attr:`void <dt.Type.void>` columns. [#3411]
 
+    -[fix] Allow chained reducers to be used for :class:`dt.FExpr`s. [#3417]
+
 
     fread
     -----
--- a/src/core/column/mean.h
+++ b/src/core/column/mean.h
@@ -25,36 +25,29 @@
 namespace dt {
 
 
-template <typename T, bool IS_GROUPED>
-class Mean_ColumnImpl : public ReduceUnary_ColumnImpl<T, IS_GROUPED> {
+template <typename T>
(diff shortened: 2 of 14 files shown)


oleksiyskononenko:
@samukweku once this PR is merged to `main`, you will need to merge `main` onto your branches and make required changes. From now on, we should use `FExpr_ReduceUnary` for all the new `FExpr` unary reducers to prevent issues and unnecessary code duplication.

oleksiyskononenko:
@oleksiyskononenko Thanks, let me know how it goes. Also, if you have time, please build this PR locally and try to run the examples you found have issues with the nested reducers. You can also add some test that you think are appropriate.

samukweku:
Sure, I'll test it locally

samukweku:
This test passes but the output is wrong: 

```py
def test_dt_count_na2():
    DT = dt.Frame(G=[1,1,1,2,2,2], V=[None, None, None, None, 3, 5])
    EXP = dt.Frame(G=[1,2], V1=[3,1], V2=[3,0])
    RES = DT[:, [dt.countna(f.V), dt.countna(dt.mean(f.V))], dt.by(f.G)]
    assert EXP.to_list() == RES.to_list()
```

The output should be `dt.Frame(G=[1,2], V1=[3,1], V2=[1,0])`

oleksiyskononenko:
Try using sum, prod, min, max or mean (refactored to FExpr reducers) instead of countna in your example and see if the result is correct.

PR #3467, opened 2023-05-26
Fix target column type detection in `LinearModel`

### LINKED ISSUE #3466: `LinearModel` error when detecting target column types for early stopping
The perfectly valid `LinearModel` fit call
```python
from datatable import dt
from datatable.models import LinearModel

DT_X = dt.Frame([1, 2, 3])
DT_y = dt.Frame([True, False, True])

lm = LinearModel()
lm.fit(DT_X, DT_y, DT_X, DT_y)
```
errors with
```python
TypeError: Training and validation target columns must have the same ltype, got: int and bool
```

The issue here is that the model didn't properly read the training target column type.

oleksiyskononenko:
In this PR we
- fix target column type detection in `LinearModel`;
- remove disabled tests, that the model never supported (were inherited from `Ftrl`).

Closes #3466

### DIFF
--- a/docs/releases/v1.1.0.rst
+++ b/docs/releases/v1.1.0.rst
@@ -196,6 +196,9 @@
     -[fix] Fixed undefined behavior when :class:`LinearModel <dt.models.LinearModel>`
         predicted on frames with missing values. [#3260]
 
+    -[fix] Fixed target column type detection in
+        :class:`LinearModel <dt.models.LinearModel>`. [#3466]
+
 
     General
     -------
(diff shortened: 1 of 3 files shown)


st-pasha:
why remove tests?

oleksiyskononenko:
The tests removed were inherited from `Ftrl` that supports categorical features, the `LinearModel` doesn't, so they were disabled from the very beginning only adding meaningless log messages when doing `make test`. If at any point we need to re-use some code from these tests, it is available in `tests/models/test-ftrl.py`.

Issue #3506, opened 2025-02-25
Add support for python > 3.11


martynward:
Are there any plans to support python versions greater than 3.11? Datatable is a great library and incredibly useful, but is beginning to fall behind on python releases and soon I won't be able to use it as other packages force a python version update.

PR #3511, opened 2025-09-11
General version updates (#3506)

ivomac:
General updates and fixes that should help with the support of python up to 3.13 (as requested in #3506).


metrizable:
Hi, I was wondering if there was interest in landing this?

